PR #13, opened 2016-08-08
Changes related to api integration

parthshah:
@vikasrohit - I've made quite a few changes in this PR.

- introduced [redux-promise-middleware](https://github.com/pburtchaell/redux-promise-middleware):
 This replaces our custom middlewares. I feel like this is much cleaner and alllows us to support chaining of promises etc. This is along the same principles outlined in react-redux reducing boilerplate.
- Using [axios](https://github.com/mzabriskie/axios) instead of isomorphic-fetch: Axios allows us to intercept API calls where we can handle user token auth. I don't think fetch supports that right now. Let me know if you think we should not be using axios for some reason. If we can figure out how to correctly handle auth token with fetch we could also use this middleware https://github.com/agraboso/redux-api-middleware
- [recompose](https://github.com/acdlite/recompose/blob/master/docs/API.md): This has a lot of useful little utilities, i am using branch() and renderComponent() to simplify showing loading indicator. see https://github.com/acdlite/recompose/blob/master/docs/API.md#branch
- loadMembers() : I've wired up fetching project member data after loadProjects() and getProject(). this maintains the state with:
```
state: 
  | --- projectSearch
  |             | --- isLoading
  |             | --- error
  |             | --- projects: []
  |             | --- totalCount
  | --- members
  |             | --- members: { '132': { userId: 132, handle: 'pshah'.... }
  | --- projectState
  |             | --- project
  |             | --- isLoading
  |             | --- error
```

Managing files is getting a little chaotic with actions/projects.js , reducers/projects.js etc . Let me know if you think there's a way to simplify it.


parthshah:
fyi @talesforce @mahestro

PR #20, opened 2016-08-09
initial commit for project list view

parthshah:
@vikasrohit please review


vikasrohit:
@parthshah 

Here are few observations:

1. UserAvatar and GridView can reside in react-components repo
2. Views for reach column of the project list, can be extracted as separate components to enable reuse.
3. I am bit concerned about the css class names chosen (I know that it is output from community) in the css. There are few css classes .txt-italic, .drop-down etc which too generic and can cause css conflicts later on with other modules. Further, some css classes can be specific e.g. item-status-date can be project-status-date.

PR #24, opened 2016-08-12
AS#148343855905798, Top bar - Logged in state

vikasrohit:
-- fixed logout link to use accounts app logout url
-- Removed dashboard link
-- used absolute link to help help

@parthshah fyi

### DIFF
--- a/src/components/TopBar/TopBar.jsx
+++ b/src/components/TopBar/TopBar.jsx
@@ -27,7 +27,7 @@ class TopBar extends Component {
     const userImage = this.props.userImage
     const domain = this.props.domain
     const mobileMenuUrl = this.props.mobileMenuUrl
-    const homePageUrl = 'https://connect.' + domain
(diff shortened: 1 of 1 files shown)


src/components/TopBar/TopBar.jsx
```diff
@@ -27,7 +27,8 @@ class TopBar extends Component {
     const userImage = this.props.userImage
     const domain = this.props.domain
     const mobileMenuUrl = this.props.mobileMenuUrl
-    const homePageUrl = '//www.' + domain
+    const homePageUrl = 'https://connect.' + domain
```
parthshah: Does this need to be an absolute link? Why not just use '/' ?

vikasrohit: Thanks @parthshah for noticing it. Yes, we can definitely get it from window.location.hostname

PR #26, opened 2016-08-24
minor content tweaks to header and footer

ajefts:
@parthshah Just doing some minor tweaks to get familiar with working with the code base and see how we work with the env...

### DIFF
--- a/src/components/TopBar/TopBar.jsx
+++ b/src/components/TopBar/TopBar.jsx
@@ -24,7 +24,7 @@ class TopBar extends Component {
 
   render() {
     const username = this.props.username
-    const userImage = this.props.userImage
+    const userImage = this.props.photoUrl
     const domain = this.props.domain
     const mobileMenuUrl = this.props.mobileMenuUrl
     const homePageUrl = window.location.protocol + '//' +window.location.hostname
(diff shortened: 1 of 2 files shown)


src/components/TopBar/TopBar.jsx
```diff
@@ -24,7 +24,7 @@ class TopBar extends Component {
 
   render() {
     const username = this.props.username
-    const userImage = this.props.userImage
+    const userImage = this.props.photoUrl
```
parthshah: Are you sure this works? Layout.jsx is providing props - username, userImage & domain. Where do you get photoUrl from?

ajefts: @parthshah I'm not actually.  userImage didn't seem to be working.  I noticed photoUrl in a couple other places for the user image.  When I changed it to photoUrl, it actually did have the effect of using the silhouette, but looks like it's not pulling in the image that's in my profile.

So, you're thinking props.userImage should be populated correctly?

ajefts: fyi @parthshah I switched it back to userImage. Looks like that is returning the user_id, as  opposed to the image...

parthshah:
@ajefts i'm closing this PR... the branches have diverted quite a bit. If we still need these changes it would be best to create a new PR against latest Dev.

ajefts:
no problem at all @parthshah

vic-appirio:
Deleted it - it is quite old at this point

PR #27, opened 2016-08-27
Fixed error in loading user's profile image in topbar

vikasrohit:
@parthshah fyi, please let me know if we can improve things here


vikasrohit:
Merging. But I am open to any changes that makes the code good.

PR #28, opened 2016-08-29
AS#166070325696601, Fix styles of create project screen

vikasrohit:
-- Rewritten(after getting changes lost in merge from dev) the changes for fixing the create project screen


vikasrohit:
Merging current set of changes...will create another PR with next set of changes for the same.

PR #123, opened 2016-09-21
Design styles

vic-appirio:
Code freeze design fixes.


vic-appirio:
The automation tests fail, but the .js files failing are the same in both branches. @parthshah → please review and merge if those tests can be fixed. We haven't modified the failing files anyway.

PR #126, opened 2016-09-21
Feature/feature picker fixes

vikasrohit:
AS#184327383400572, Project Details/ Feature Picker: Adding / Clicking on a custom feature leaves the feature picker modal unresponsive
AS#184327383400574, Feature picker || Custom feature is shown as form even after adding the feature

### DIFF
--- a/src/projects/FeatureSelector/CustomFeatureForm.jsx
+++ b/src/projects/FeatureSelector/CustomFeatureForm.jsx
@@ -66,7 +66,7 @@ class CustomFeatureForm extends Component {
   }
 
   render() {
-    const { isEdittable, onCancel } = this.props
+    const { isEdittable, onCancel, featureData } = this.props
     const { data, isActive, showDeleteModal } = this.state
     const submitButton = !isActive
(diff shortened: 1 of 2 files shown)


src/projects/FeatureSelector/CustomFeatureForm.jsx
```diff
@@ -88,22 +88,27 @@ class CustomFeatureForm extends Component {
...
+                name="title"
+                label="Feature name"
+                validations="minLength:1" required
+                validationError="Feature name is required"
+                wrapperClass="row"
+                // placeholder="My awesome feature"
+                value={ _.get(data, 'title', '') }
+              />
+            }
+            { !isActive ?
```
parthshah: Why do you need this code - block? You can set `disabled={isEdittable}`. TextInput & Textarea would always be visible.

vikasrohit: Agree. However, I had already made those changes before your comment about keeping the form even for added custom features. I was trying to match with the designs where we are not showing the form for the added features.

PR #451, opened 2016-10-10
Style top nav

vikasrohit:
@vic-appirio I am done with applying the padding dynamically for the wrapper div. Seem like it is working. Please check on your local, and if you find it good. Please merge the PR.


vic-appirio:
The top navigation is not hiding when "New Project" is overplayed. That's the only thing we need fixed.
screen shot 2016-10-10 at 9 10 32 am

src/components/TopBar/TopBar.scss
```diff
@@ -6,33 +6,42 @@ $tc-rgba-10: rgba(71,71,79,0.38);
...
-.tc-header{
+.tc-header {
   height: 60px;
-  background: $tc-gray-neutral-light;
+  // background: $tc-gray-neutral-light;
   line-height: 60px;
-  z-index: 100;
+  height: 60px;
+  background-color: $tc-black;
+  z-index: 20;
```
vikasrohit: @vic-appirio Do you know the reason of setting the z-index of the header to 100? Right now modals have z-index of 21 so we need to keep other element's z-index below that.

vic-appirio: Good to know, @vikasrohit!

Issue #267, opened 2016-09-28
throwing error on saving blank name of uploaded file 


swetaipec:
Steps:
1.) Open the application
2.) Login with username sweta.singh
3. create new project
4. go to SPecification
5. go to Upload file
6. Upload a file
7. edit file name
8. same as blank

Expected Result: should ask to input the valid name
Actual Result: throwing error

Environment: windows 7, chrome latest

![image](https://cloud.githubusercontent.com/assets/16131461/18932285/5860bebe-85ed-11e6-965f-1a205255a41c.png)

swetaipec:
throwing error on updating name

parthshah:
@vikasrohit - assigning this to you. Let me know if you want to launch challenges to fix a bunch of these issues.

vikasrohit:
@vic-appirio  @fnisen How to show the validation error?

@parthshah From the implementation point, we can not use Formsy form here because it would make nested forms(EditProjectForm and FileListItem). So, I guess, we need to handle the validation ourselves.

vikasrohit:
Should we put a character limit as well on file name? We have limit of 250 for the description and description is also a required field for updating the file information.

fnisen:
@vikasrohit can you please show a validation error right below the title, in our usual style? The copy can be: "The file name cannot be blank." 

Also: 
- Can we make description optional? 
- Let's go with 50 characters for the title limit. Good idea.

vikasrohit:
@fnisen description is made compulsory in API. @mdesiderio can we make it optional?

vikasrohit:
@fnisen I have changed the copy and put the max length on file name. However, I haven't remove required validation for description field because it requires changes in API. May be we can track that separately.

vikasrohit:
@parthshah did the change in his late night today. :) So, now all requirements are done for the issue.

fnisen:
@vikasrohit I get into an error state if I delete the entire name and then try to add a new one. I'm not able to save in that case: http://screencast.com/t/NVO5RSA2qq. 

Also, @parthshah, the description is still being required. I can't save without a description--there's an error message being thrown ("Whoops, there was a problem.")

fnisen:
OK, description portion is now working as expected (not required), so just seeing the name issue @vikasrohit

vikasrohit:
Fixed. Changes are in dev now.

PR #458, opened 2016-10-12
Github issue #267 throwing error on saving blank name of uploaded file

vikasrohit:
-- Fixed, @parthshah fyi

### DIFF
--- a/src/components/FileList/FileListItem.jsx
+++ b/src/components/FileList/FileListItem.jsx
@@ -59,14 +59,14 @@ export default class FileListItem extends React.Component {
   validateTitle(errors) {
     const title = this.refs.title.value
     if (!title || title.trim().length === 0) {
-      errors['title'] = 'Please enter a valid file name'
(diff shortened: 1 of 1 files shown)


src/components/FileList/FileListItem.jsx
```diff
@@ -33,23 +39,69 @@ export default class FileListItem extends React.Component {
...
+  }
+
+  validateTitle(errors) {
+    const title = this.refs.title.value
+    if (!title || title.trim().length === 0) {
+      errors['title'] = 'Please enter a valid file name'
+    }
+  }
+
+  validateDesc(errors) {
```
parthshah: I think it's okay if description in empty. It's not a required field. 
@fnisen ?

parthshah: It shouldn't be. I'll make the necessary changes to the API and update this branch. Thanks for finding this bug.

Issue #257, opened 2016-09-28
[Specification] - the same image for different types of files uploaded


onetliviunicolae:
Expected behavior

 - The image must be different.

Actual behavior

- The same image icon displayed (GZIP) for any types of file.

Steps to reproduce the problem

- Login into application as manager.

- Go to Dashboard.

- Edit any existing project.

- Go to Specification tab section.

- Go to Project Files (2) and upload a files: png, pdf, xls, ...

Environment

Operating System(s): Windows 7
Device(s): Laptop HT 635
Browser(s): Chrome 49.0.2623.112 m
            Mozilla Firefox 48.0.2  
            IE 11.0
User role: manager
Account used: pat_monahan

see attach!

![gzip](https://cloud.githubusercontent.com/assets/5114907/18931746/1fa3ad14-85d6-11e6-9357-1dfed246c32c.png)

parthshah:
@vikasrohit assigning this to you. Could we use some sort of react svg loader to load svgs through require('') ?

vikasrohit:
@parthshah I have committed the changes, please review the [PR](https://github.com/appirio-tech/connect-app/pull/460).

fnisen:
Fixed with https://github.com/appirio-tech/connect-app/pull/460

PR #460, opened 2016-10-12
Github issue #257 [Specification] - the same image for different types of files uploaded

vikasrohit:
-- Dynamically picked icon for the file type

fyi, @parthshah please let me know if you think it can be improved


parthshah:
Are there any comparisons on which method of loading svgs is more performant? 
This fix is fine for now IMO.

vikasrohit:
You mean comparison between loading svg directly as image and loading them as React component? If yes, I guess @vic-appirio has better idea over that.

vic-appirio:
I think that we can save ourselves time and effort if we can just dump the appropriate .svg files and dynamically convert them to react components. After some research of different approaches out there, [svg-react-loader](https://github.com/jhamlet/svg-react-loader/blob/version-0.4.0/README.md) seems like the best candidate for our tasks:
- design can dump/update/maintain the appropriate .svg files
- developers can request the icons in a consistent way, just need to know the correct filename
- icon properties could be handled by react-css-modules via `styleName` dynamically, which will limit the need to parameterize the .svg properties (consistency through reuse of global icon css classes).

Detailed [article for this approach](https://diessi.ca/blog/svg-images-as-react-components-with-webpack/);
Another [work-around for .SVG icon systems](http://blog.goguardian.com/nerds/an-odyssey-to-find-a-sustainable-icon-system-with-svgs-in-react)

As per performance - I have no idea which approach will be the fastest. My goal would be to load only 1 SVG and modify it properties instead of switching multiple static .svg or .png images.

fnisen:
cc @vikasrohit, @parthshah

vikasrohit:
Create separate issue to track this https://github.com/appirio-tech/connect-app/issues/481

PR #530, opened 2016-11-08
Updating copy on AddLink.jsx

fnisen:
Changed titles for "Add Link" fields


vikasrohit:
This might need some more changes in code as you have changed the field names which might be used in javascript code as well.

fnisen:
@vikasrohit so do you need me to do something or create an issue?

vikasrohit:
Sorry, @fnisen I need to validate the changes which I was not able to do. I would do that on Monday now. Basically, we can change the field label but can't change field name because doing that would require updating all existing projects' data to be updated as well to reflect the same, otherwise we would see blank link titles for existing links in projects.

vikasrohit:
@fnisen I have committed changes to avoid potential issues which could have cause one of new(added after merging this PR without any change) and old(before merging this PR) links to be unable to render properly. There might be some runtime error or the links would have been rendered as empty rows.

fnisen:
Thank you. 👍 Now live.

Issue #527, opened 2016-11-08
Discussions: Discussion thread not showing all posts


fnisen:
On our project discussion, https://connectv2.topcoder.com/projects/41/discussions, I was earlier seeing a state where not all the posts were showing. Here is what I saw (first message missing; notice no option to "Load Earlier Messages"):

<img width="761" alt="screen shot 2016-11-08 at 9 28 29 am" src="https://cloud.githubusercontent.com/assets/12552247/20118521/a2e9dfe2-a5b9-11e6-98f3-771032f6a18d.png">
 
Here is the full list of posts: 
<img width="734" alt="screen shot 2016-11-08 at 1 31 43 pm" src="https://cloud.githubusercontent.com/assets/12552247/20118541/b6ef72d6-a5b9-11e6-939e-5864df079c00.png">

Here is the correct, compressed view, which is now showing: 
<img width="737" alt="screen shot 2016-11-08 at 1 32 00 pm" src="https://cloud.githubusercontent.com/assets/12552247/20118551/c3d45fe8-a5b9-11e6-9cf7-bef56b09b8de.png">

vikasrohit:
@fnisen Sorry, but I didn't understand how to reproduce the issue. Do we have defined steps to reproduce it or its intermittent?

fnisen:
@vikasrohit it happened to me just that one time; not sure how to reproduce it. If you don't see any errors in the code, let's not do anything for now and see if it reoccurs.

vikasrohit:
@fnisen Actually, the code always doubtful because we have implemented the paging for posts with workarounds and hacks. And that is done because we don't have proper paging support in the API which in turn is limited by discourse's API limitations.
On the other hand while I was reviewing the submissions for a [challenge](http://www.topcoder.com/challenge-details/30055675/?type=develop&noncache=true), I saw another issue in dev env. Please check [this](https://connectv2.topcoder-dev.com/projects/273) project once and notice the posts of the first project status (title as "Testing message post in IE"). If you check the API response for the same, you will find only two posts for that topic while the UI is showing more than one comments.
<img width="1280" alt="screen shot 2016-11-14 at 5 58 11 pm" src="https://cloud.githubusercontent.com/assets/2417632/20264962/3c64cb80-aa94-11e6-96d4-1bc53a04aaa8.png">

fnisen:
@vikasrohit yikes. OK, what do you recommend? 

@parthshah FYI

vikasrohit:
[...]

PR #540, opened 2016-11-15
Github issue #527, Discussions: Discussion thread not showing all posts

vikasrohit:
-- Found critical issue when reviewing the code the mentioned issue. It was causing wrong posts to be shown under a topic.


parthshah:
Good catch. Thanks!

PR #554, opened 2016-11-23
Update index.html

fnisen:
Adding Heap Analytics tracking script

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,6 +6,10 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no">
     <base href="/">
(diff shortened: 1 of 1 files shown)


vikasrohit:
@fnisen This PR is against master branch. Is it intended?

vikasrohit:
Further, how can we verify if it is working fine? I mean is there any dashboard/ui by Heap Analytics where we can see, before merging into master, that it is configured correctly.

fnisen:
@vikasrohit - sorry, should have clarified in the PR. Yes, it was intentional because the data is separated per environment. And yes, you can check in the Heap Analytics app whether the data is flowing once the tracking code is added (I don't think there's a way to check beforehand). I've sent you an app invite by email.

vikasrohit:
Okay, but merging a PR in master, would build and deploy to production.

fnisen:
OK, what would you recommend?

PR #560, opened 2016-11-23
Update index.html to add Heap Analytics tracking

fnisen:
Will need to change the heap.load ID in prod

### DIFF
--- a/src/config/constants.js
+++ b/src/config/constants.js
@@ -213,3 +213,5 @@ export const CODER_BOT_USER_LNAME = 'Bot'
 export const PROJECT_MAX_COLORS = 5
 
 export const AUTOCOMPLETE_TRIGGER_LENGTH = 3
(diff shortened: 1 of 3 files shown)


vikasrohit:
@fnisen Updated the code to get the parameterized app id.

fnisen:
Thank you @vikasrohit! Double-checked everything and it's looking good. Data is being collected in Dev. Can you please deploy to prod so we can start tracking?

vikasrohit:
It would require all changes from the dev branch to be pushed to production. Are we good to go with production release with current set of changes in dev?

fnisen:
@vikasrohit don't worry about this. We ended up pushing out the changes today b/c we needed to rush out the new terms page. I also found some issues with Vic's style changes, which he was able to fix. Thanks.

Issue #579, opened 2016-12-09
General: allow "admin" role to have same rights as "manager" role


fnisen:
Currently there are 4 user roles in Connect: 
1. Customer
2. Copilot
3. Manager
4. Admin (actually a system role)
Admins can see all projects, but cannot participate in projects the same way that managers can. This has led to several admin users (such as Valerie) creating different accounts that allow them to serve as managers. To simplify their workflow, can we give admins the same access right as managers in Connect? Admins can maintain additional rights if needed (I'm not sure if they can do things that managers can't), but in general it would be good if Admin rights were a superset of Manager rights. From talking to Parth, this seems to be a frontend issue more than a backend one.

vikasrohit:
@fnisen I have made the changes to make admins behave just like managers. However, I am not able to test it out on my local as I am not aware of a user who only has the administrator role. I tried with tony's account but he has 'Connect Manager' roles as well, so I am not sure if it is working correctly or not. Do we have any such user in dev?
Further, I tried to add administrator role to my account to test the feature, however, even after looking at dozens of tables, I am not sure from where we are picking up the user's roles. @ajefts @parthshah  Can you guys please let me know how to update a user's role from the database?

parthshah:
Sure, see Authorizations DB (https://console.aws.amazon.com/rds/home?region=us-east-1#dbinstances:id=authorization1;sf=all) table role_assignment. roles table should have all the roles.

vikasrohit:
Thanks @parthshah Please review the PR #582. I have tested these changes on my local with vbailey user (I have added `administrator` role to the user apart from `copilot` role which she already had) and now I am able to see all projects and able to join them as Manager.

fnisen:
Merged by Parth and working for me in Dev, so closing this one.

PR #582, opened 2016-12-13
Github issue #579, General: allow "admin" role to have same rights as "manager" role

vikasrohit:
-- Aliased administrator roles to act like as connect manager

@parthshah fyi, please let me know if we need to make any other change to achieve the functionality. I have tested these changes on my local with vbailey user (I have added `administrator` role to the user apart from `copilot` role which she already had) and now I am able to see all projects and able to join them as Manager.

cc: @fnisen


parthshah:
👍

Issue #568, opened 2016-12-07
Dashboard: status update thread is too "sticky" 


fnisen:
Been seeing and have gotten reports of issues with our status update components not refreshing quickly enough. A user will be on one project, on the Dashboard, go to the listing and click on another project, and see the Dashboard for the previous project. It will refresh after a delay or sometimes not at all and require a manual refresh. 

@vikasrohit can you please see if you're able to reproduce? I've been able to on projects that are "In Review" and have some sort of manual updates. Could there be some issue with how we're caching the content in the React component?

vikasrohit:
@fnisen I can reproduce this on my local as well, though feeds from the previous projects appears for fraction of a second. I have committed some changes to fix this issue. However, I would like @parthshah to have a quick look at those changes just to avoid any silly side effect of my changes.

fnisen:
Thanks @vikasrohit! @parthshah please check it out.

fnisen:
Merged by Parth. Closing.

PR #584, opened 2016-12-13
Github issue #568, Dashboard: status update thread is too "sticky"

vikasrohit:
-- Cleared the redux state on successful projects load action

@parthshah Can you please have a quick look at the change and let me know if I have missed anything which can cause regression issue?


parthshah:
Looks good. Merging

PR #590, opened 2016-12-15
Github issue#587, Comments Sort order within a topic is incorrect

vikasrohit:
-- Fixed. We were missing sorting the posts when loading more posts for a feed

@parthshah fyi


vikasrohit:
Merging branch before checks are completed because it seems there is some issue with npm install in the builds.

PR #591, opened 2016-12-15
Github issue#585, Ignore 'user-joined' type posts in Topic comments

vikasrohit:
-- Ignored empty posts whose author are 'system'

@parthshah fyi


vikasrohit:
Merging branch before checks are completed because it seems there is some issue with npm install in the builds.

Issue #593, opened 2016-12-21
Increase timeout when posting topics / comments to message service


parthshah:
At times we need to first create a user in discourse and then create a topic. This at takes a long time to process. Unfortunately on FE the user sees a timeout and then when they refresh the page the post is visible. Please set the timeout to 1.5 mins to avoid such situations. 
fyi - @fnisen

vikasrohit:
@parthshah I have made the changes, not sure how to validate that axios is actually respecting the new timeout. There are few issues mentioned in web community about axios not respecting the timeout settings in some cases. Here is the PR https://github.com/appirio-tech/connect-app/pull/595

PR #595, opened 2016-12-22
Github issue #593, Increase timeout when posting topics / comments to message service

vikasrohit:
-- Done. But not tested yet. Not sure how to validate that axios is actually respecting the new timeout. There are few issues mentioned in web community about axios not respecting the timeout settings in some cases.

@parthshah fyi, I think these changes should be enough, but not able to test them.


parthshah:
Thanks 👍

PR #666, opened 2017-03-03
Define another screen and screens limit

### LINKED ISSUE #637: Define another screen and screens limit
### Expected behavior
If the "Define another screen" button is disabled due to the number of screens limit, it's confusing to let the user enter values into the fields of a screen which can't be saved.

### Actual behavior
Currently the new screen can be added but not saved. this is confusing.

Open the project specification page, select 3 as number of screens and add 3 new screens (4 total). You can't save the 4th screen and it shouldn't be possible to add it at all

### Screenshot/screencast
Attach or link a resource.
image

--

#### Environment
- OS: all
- Browser (w/version):all
- User role (client, copilot or manager):all
- Account used: all

ThomasKranitsas:
Closes #637


fnisen:
Please see my comments here: https://github.com/appirio-tech/connect-app/issues/637#issuecomment-283962767

ThomasKranitsas:
@gondzo I pushed a commit here.

You will notice that you can click the `Define another screen` button even if the previous screen is not filled.
This is because the validation applies on the `Save changes` button instead.
The `Define another screen` is **ONLY** used to add new screen.

In conclusion: You can add multiple screens but you can save them only when all of them are correctly filled.

FYI @fnisen

Issue #672, opened 2017-03-04
Saving product type into project.details


gondzo:
### Expected behavior
Product type (2nd step choice in the create project wizard) should be saved to details.products field. It should be an array field with only one element for now - the subtype for the current project.

Files that should be updated :
creating a project > https://github.com/appirio-tech/connect-app/blob/feature/productsv1/src/projects/create/components/ProjectWizard.jsx#L50
reading the product type > https://github.com/appirio-tech/connect-app/blob/feature/productsv1/src/projects/detail/containers/SpecificationContainer.jsx#L55 - should check if details.products[0] exists and load the specification template, otherwise fall back to topcoder.v1

### Actual behavior
Type is just saved to the project.type field.

### Steps to reproduce the problem
Create a new project
--

#### Environment
- OS:all
- Browser (w/version):all
- User role (client, copilot or manager):all
- Account used: all

fnisen:
@vikasrohit can you please weigh in on this one?

vikasrohit:
Sorry @fnisen missed to update the github, however, we (me and Samir) had discussion about this. This is valid and critical issue which is required before go for bug hunt.

fnisen:
@gondzo I'm not going to test this one b/c it's not a user issue. Can you please verify yourself or ask @vikasrohit is you need another opinion. Thank you.

gondzo:
Looks good.

PR #688, opened 2017-03-07
Save product type into project.details

ThomasKranitsas:
Related issue #672

### DIFF
--- a/src/projects/create/components/ProjectWizard.jsx
+++ b/src/projects/create/components/ProjectWizard.jsx
@@ -43,7 +43,8 @@ class ProjectWizard extends Component {
         devices: [],
         utm: {
           code: s.projectRef
-        }
+        },
+        products: [config[s.projectType].subtypes[s.projectSubType].id]
       },
       description: 'No description provided',
       name: s.projectName,
--- a/src/projects/create/components/ProjectWizard.jsx
+++ b/src/projects/create/components/ProjectWizard.jsx
@@ -48,7 +48,7 @@ class ProjectWizard extends Component {
       },
       description: 'No description provided',
       name: s.projectName,
-      type:  config[s.projectType].subtypes[s.projectSubType].id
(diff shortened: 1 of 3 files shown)


src/projects/create/components/ProjectWizard.jsx
```diff
@@ -43,7 +43,8 @@ class ProjectWizard extends Component {
         devices: [],
         utm: {
           code: s.projectRef
-        }
+        },
+        products: [config[s.projectType].subtypes[s.projectSubType].id]
```
vikasrohit: https://github.com/ThomasKranitsas/connect-app/blob/ef1b699e44f95b87d264392dd2376d8f48e93dd7/src/projects/create/components/ProjectWizard.jsx#L51 should save config[projectType].id (you need to add id field to the project type objects in config json) directly not the product (subType).

ThomasKranitsas: @vikasrohit sorry.. I'm not sure I understand what you mean here..
1) What ID(s) should I add in project types in config.json?

2) So the `project.type` should be the `Project type ID` and the `project.details.products[0]` should be the `Project subType ID`?

vikasrohit: 1. Id of project type would be values which we were storing in type field of the project before all these changes i.e. visual_design for Design, visual_prototype for Prototype, app_dev for Development and generic for work projects.
2. Yes, correct as said above project type id is which we were storing as project type earlier.

ThomasKranitsas: Thanks I'll push another commit in few minutes

ThomasKranitsas:
@vikasrohit can you check the new commit? thanks!

vikasrohit:
@gondzo although these changes should work, still please double check by running the application with these changes and merge after it.

PR #772, opened 2017-03-10
textarea min-height fixed for 1 row height.

vic-topcoder:
Fixing the min-height for text areas in the project specification


parthshah:
looks good.

PR #797, opened 2017-03-11
removing *

iit2009060:


gondzo:
already fixed in earlier commits

Issue #761, opened 2017-03-09
Design Guidelines question count missing 


karthikbecse:
### Expected behavior
should show question  counts correctly 

### Actual behavior
by default it is not shown 

### Steps to reproduce the problem
- Login As manager
- click New Project 
- Enter valid values 
- Click Design
- Click Wireframe
- Goto Project Specfication 
- See Design Guidelines question count 

### Screenshot/screencast

![screenshot-connect topcoder-dev com 2017-03-09 17-49-18](https://cloud.githubusercontent.com/assets/3225433/23750010/f99bd32c-04f0-11e7-8ae5-6ae15f46cfec.png)

--

#### Environment
- OS:Windows 8.1
- Browser (w/version):FF 52.0
- User role (client, copilot or manager):Manager
- Account used: pat_monahan

Colinh84:
Working on this. Can't assign myself yet though.

PR #810, opened 2017-03-11
Issue 761: Fix for question count

Colinh84:
https://github.com/appirio-tech/connect-app/issues/761


gondzo:
@Colinh84 this change makes the notes and files section display the question counter too, which should not be happening

PR #829, opened 2017-03-14
issue 761: fixed progress indicator to only show with questions

Colinh84:
https://github.com/appirio-tech/connect-app/issues/761


Colinh84:
@gondzo fixed to show only with questions

Issue #778, opened 2017-03-11
[$40] Getting unnecessary unsaved changes warning (regression)


fnisen:
### Expected behavior
User should not be warned against unsaved changes if project has been saved

### Actual behavior
User is warned there are unsaved changes
![screen shot 2017-03-10 at 4 30 42 pm](https://cloud.githubusercontent.com/assets/12552247/23818341/f353ff26-05ae-11e7-905f-b4a26d1362cf.png)

### Steps to reproduce the problem
- Access project https://connect.topcoder-dev.com/projects/1030/specification
- Use account pat_monahan/topcoder123!
- Make changes to project spec and click Save. Observe "Project updated" message. Click on Dashboard link.

### Screenshot/screencast
See above

--

#### Environment
- OS: Mac 10.5
- Browser (w/version): Chrome latest
- User role (client, copilot or manager): manager
- Account used: pat_monahan/topcoder123!

rrandom:
Can't repreduce.

gondzo:
visit the above link, add a new screen, save the project and click the dashboard link > the unsaved changes alert is displayed

gondzo:
@rrandom

fnisen:
@shubhendusaurabh how is it going with this issue? Will you be able to finish today?

shubhendusaurabh:
sorry won't be able to finish this.

fnisen:
@shubhendusaurabh how much more time do you need? You can hold on to this issue still--not need to drop it.

shubhendusaurabh:
@fnisen I gave it a try, but it looked tough in my limited React knowledge.

fnisen:
OK, thanks @shubhendusaurabh

ThomasKranitsas:
@fnisen I can imagine what the issues might be here but can you confirm that this is only for the `App screens` section?

fnisen:
@ThomasKranitsas it's not isolated from what I can tell. If I make changes in the "Definition" section (top section) and then try to navigate to a new tab within Connect (e.g., Dashboard), I still get the warning

gondzo:
Yes, app screens section only. App screens field isn't updated in the form
after its submitted. You can verify that by comparing
this.refs.form.getPristineValues () and this.refs.form.getCurrentValues ()
in EditProjectsForm.jsx

fnisen:
Thanks @gondzo!

gondzo:
[...]

PR #845, opened 2017-03-14
Fix unnecessary unsaved changes warning

ThomasKranitsas:
Related issue: #778

### DIFF
--- a/src/projects/detail/components/EditProjectForm.jsx
+++ b/src/projects/detail/components/EditProjectForm.jsx
@@ -52,7 +52,7 @@ class EditProjectForm extends Component {
       isSaving: false
     }, () => {
       // Reset form
-      if (this.refs.form) this.refs.form.reset(nextProps)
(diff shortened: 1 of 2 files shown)


gondzo:
@ThomasKranitsas this fixes the save issue, but introduces a bug in sections other than app screens. Editing any field in those requires saving the project twice before everything is actually saved. For example, edit project description, target device, or number of screens field, click save -> save gets enabled again. click again, the project is saved and save button is disabled.
I think 'this.refs.form.reset(nextProps)' causes that. Take a look at SpecQuestions.jsx (editing these form fields causes the above error), it's very similar to SpecScreenQuestions.jsx (editing these fields won't raise the above error)

ThomasKranitsas:
@gondzo can you please check with the latest commit? I tried by changing a few different sections and was working fine. Let me know if it's still not working.

Colinh84:
@gondzo This commit broke the form. If you make any changes in the appScreen section and click save, all saved information in the rest of the form is reset to empty.

Issue #751, opened 2017-03-09
[$50] 'Screen name' and 'Answer' block validation won't display to all the sections


sdgun:
Description
-----------------------------------------------------------------
'Screen name' and 'Answer' block validation won't display to all the sections

Steps to Reproduce
-----------------------------------------------------------------

- Open the site: https://connect.topcoder-dev.com
- Login as a valid user - Manager (Mentioned in the Environment section)
- In the Dashboard
- Open a Project (https://connect.topcoder-dev.com/projects/960/specification)
- Click on 'Specification' tab
- Select 9-15 from 'How many screens do you need designed?'
- Click Save Changes from that section
- Go to 'App Screens'
- Click 'Define another screen'
- Add 9 screens
- Add 'Screen name' and 'Answer' to the first 3 items
- Click Save Changes from that section
- Now scroll up and delete the first 2 screens
- Scroll down and check the below items
- Validation displayed only for two items

Expected Result(s)
-----------------------------------------------------------------
- Must display the validation for all the items

Actual Result
-----------------------------------------------------------------
- 'Screen name' and 'Answer' block validation won't display to all the sections

Environment
-----------------------------------------------------------------
- **Device(s):** Laptop
- **Browser(s):** IE11, Chrome 56.0.2924.87
- **Operating System:** Windows 7 Pro 64bit
- **User Accounts:** pat_monahan
- **User Role:** Manager

Image/Video/JS Log (If not attached here, Please check the comment section) 
-----------------------------------------------------------------

sdgun:
https://www.screencast.com/t/V1Is9bCG

fnisen:
@gondzo this looks pretty messy. What are the current rules for validation?

Colinh84:
@gondzo Need a bit of clarification here. When are these validation errors supposed to appear? Currently, they do not show up for any screen field that does not have content, and only appear when a field that has content and is saved has its content deleted.

gondzo:
@Colinh84 that's fine - we don't want all fields to have validation errors by default.

Colinh84:
@gondzo So then the correct thing to do here is to not show any validation errors when the screens are deleted, since the remaining screens never had content in the first place?

gondzo:
correct

gondzo:
@Colinh84 Can you handle #767 as well since it's related to this one and we'd get merge conflicts if we handle them separately ? Take a look at the comments too as there are a few items there too.

Colinh84:
[...]

PR #851, opened 2017-03-15
Issue 751: Fixed block validation bug

Colinh84:
https://github.com/appirio-tech/connect-app/issues/751

### DIFF
--- a/src/projects/detail/components/EditProjectForm.jsx
+++ b/src/projects/detail/components/EditProjectForm.jsx
@@ -52,7 +52,7 @@ class EditProjectForm extends Component {
       isSaving: false
     }, () => {
       // Reset form
-     // if (this.refs.form && this.refs.form.isChanged()) this.refs.form.reset(nextProps) <-- this is breaking form
+      if (this.refs.form && this.refs.form.isChanged()) this.refs.form.reset(nextProps)
     })
   }
 
--- a/src/projects/detail/components/EditProjectForm.jsx
+++ b/src/projects/detail/components/EditProjectForm.jsx
@@ -52,7 +52,7 @@ class EditProjectForm extends Component {
       isSaving: false
     }, () => {
       // Reset form
-      if (this.refs.form && this.refs.form.isChanged()) this.refs.form.reset(nextProps)
+     // if (this.refs.form && this.refs.form.isChanged()) this.refs.form.reset(nextProps) <-- this is breaking form
     })
   }
 
--- a/src/projects/detail/components/SpecScreens.jsx
+++ b/src/projects/detail/components/SpecScreens.jsx
@@ -9,6 +9,7 @@ class SpecScreens extends Component {
   constructor(props) {
     super(props)
     this.addScreen = this.addScreen.bind(this)
+    this.appScreen = []
   }
 
   componentWillMount() {
@@ -34,8 +35,12 @@ class SpecScreens extends Component {
   deleteScreen(index) {
     const screens = [...this.state.screens.slice(0, index),
       ...this.state.screens.slice(index + 1)]
-    this.setState({ screens })
     this.props.setValue(screens)
+    this.setState({ screens }, () =>{
+      screens.forEach((val, i) => {
+        this.appScreen[i].sForm.setFormPristine(true)
+      })
+    })
   }
 
   addScreen(screen, addEmptyScreen) {
@@ -65,6 +70,7 @@ class SpecScreens extends Component {
             <h5 className="screen-title">Screen {index + 1}</h5>
           </div>
           <EditAppScreenForm
+            ref={el => this.appScreen[index] = el}
             screen={screen}
             questions={this.props.questions}
             onUpdate={this.updateScreen.bind(this, index)}
--- a/src/projects/detail/components/EditAppScreenForm.jsx
+++ b/src/projects/detail/components/EditAppScreenForm.jsx
@@ -54,6 +54,7 @@ class EditAppScreenForm extends Component {
 
     return (
       <Formsy.Form
+        ref={el => this.sForm = el}
         onInvalid={() => this.onValidate(false)}
         onValid={() => this.onValidate(true)}
         onChange={this.update}
--- a/src/projects/detail/components/SpecScreenQuestions.jsx
+++ b/src/projects/detail/components/SpecScreenQuestions.jsx
@@ -25,7 +25,7 @@ const SpecScreenQuestions = ({questions, screen}) => {
     const elemProps = {
       name: q.fieldName,
       label: q.label,
-      value: _.get(screen, q.fieldName, undefined),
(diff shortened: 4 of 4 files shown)


src/projects/detail/components/EditProjectForm.jsx
```diff
@@ -52,7 +52,7 @@ class EditProjectForm extends Component {
       isSaving: false
     }, () => {
       // Reset form
-      if (this.refs.form && this.refs.form.isChanged()) this.refs.form.reset(nextProps)
+     // if (this.refs.form && this.refs.form.isChanged()) this.refs.form.reset(nextProps) <-- this is breaking form
```
vikasrohit: @gondzo I don't remember the exact issue, however, I guess this code block was added yesterday to fix one of the issues. Can you please check for regression here?

src/projects/detail/components/SpecScreens.jsx
```diff
@@ -34,8 +35,11 @@ class SpecScreens extends Component {
   deleteScreen(index) {
     const screens = [...this.state.screens.slice(0, index),
       ...this.state.screens.slice(index + 1)]
-    this.setState({ screens })
-    this.props.setValue(screens)
```
vikasrohit: @Colinh84 Shouldn't we set the value of the screens? I haven't tested it by running the code, however, I guess without this line it won't update the deleted screens in the parent form and if user clicks the Save Changes button immediately after deleting a screen, it won't delete the screen from the server. Have you tested it?

src/projects/detail/components/EditProjectForm.jsx
gondzo: it's added to prevent the unsaved changes dialog when leaving the page after saving the screens (and also to correctly disable the save button). Removing it gets us back to square one -> you can save multiple times without changing anything
@Colinh84

src/projects/detail/components/SpecScreens.jsx
gondzo: It won't enable the save button (if you edit/add a different screen and save, the change is persisted correctly)
@Colinh84 @vikasrohit

vikasrohit: Didn't get it @gondzo

gondzo: I'm just confirming that setValue is needed.
Without calling setState&setValue in the above snippet, the 'Save Changes' button won't be enabled so you can't even save the change when deleting a screen. Deleting a screen, then editing a different would enable save button and saving would correctly persist the changes (deleted screen would stay deleted).

vikasrohit: Okay 👍

src/projects/detail/components/EditProjectForm.jsx
vikasrohit: We can get rid of this code block (line number 53-56) if we reset the form in [submit](https://github.com/appirio-tech/connect-app/blob/dev/src/projects/detail/components/EditProjectForm.jsx#L124) method. To do that we have return promise object in [saveProject](https://github.com/appirio-tech/connect-app/blob/dev/src/projects/detail/containers/SpecificationContainer.jsx#L47) method and reset the form in `then` callback.
@Colinh84 Please let me know if that helps with your changes.

@gondzo Ideally a form should reset itself automatically once the submit is invoked, right? I think this problem is happening because of nested forms. We definitely need to get rid of them.

src/projects/detail/components/SpecScreens.jsx
Colinh84: @gondzo I added back the setValue call to the branch, everything should work as expected here now.

src/projects/detail/components/EditProjectForm.jsx
gondzo: @Colinh84 can you take over fixing 778 too as both of these are a priority ?

vikasrohit: @Colinh84 I guess, it would reset the form to the initial values with which the form was rendered. So, calling it in submit method after successful update to the server, should not cause the issue. Please let me know if I am missing something here.

vikasrohit:
@gondzo we should accept this PR only when it does not reintroduce #778. Please consult with @fnisen before merging the PR.

Colinh84:
@vikasrohit fine, I added back the line from the #778 commit. That commit is buggy though with or without my fix for this issue. If you do not see the problem, easy to reproduce:
- Start a new design wireframe project
- Click on "specification"
- Fill out first part of form completely, and click first save button.
- In App Screens section, add the first screen by clicking "Define another screen"
- Scroll back up to the first section and see that all the data you previously entered has now disappeared.
- Look at the sidebar and notice "Submit for Review" button is now available even though the required screen you just added is empty.

I don't have time to fix #778 which was assigned to someone else to fix, and I think it is a bigger issue to fix than simply moving the reset() call I previously commented out.

I also just noticed after just merging in the latest 5 commits into my local dev branch, that the "submit for review" button now becomes active after only saving the first part now, without even adding any of the required screens, so there are even more problems, none of which relate to my code in this fix.

@gondzo @fnisen

fnisen:
@Colinh84 thank you for finding that. You're totally right--we're losing data.
@gondzo this is serious enough to delay shipping. If it's not easy to fix, as a stop gap, can we remove any validation logic around the "Submit for Review" button and just keep it active at all times?

parthshah:
As a workaround, does forcing a full page refresh after saving the form fix the problem? Not a big fan of a refresh but it's defn better than us losing data.

vikasrohit:
@Colinh84 Thanks for sharing the steps for reproducing the issue. However, the enabled state of 'Submit for Review' button was never ever linked to the `App Screens` section before my changes to fix that. So now it is linked with valid state of `App Screens` section. However, earlier there was one more bug where it was saving empty screens as soon as user has clicked on 'Define another screen' button, fixed that as well. So, in nutshell, now user can not save an empty screen in any case and even if user is able to do so by some mean, 'Submit for Review' button won't be enabled. However, there is a catch. Right now 'Submit for Review' button is enabled if user has not defined even a single screen because current logic validates only for any invalid/incomplete screen to disable the button. Next target is to disable the button when user has not defined a good (within the range of answer of the `number of screens` question from Definition section) number of valid screens.

@gondzo As I can see you applied a temporary fix for that issue raised by @Colinh84, do you think it is now safe to merge this PR into dev?

fnisen:
@gondzo what do you think?

fnisen:
@gondzo how did you resolve this?

vikasrohit:
This bug was from our older bug hunt not this one. @gondzo Can you please check if the original issue #751 is still reproducible?

gondzo:
this isn't an issue anymore as we changed the component for defining screens. Also, you can't save the changes unless all screens pass validation.

PR #882, opened 2017-04-25
Permalink for status post and discussions message

vikasrohit:
-- Added permalink for status posts and discussion messages
-- Permalink for comments on both pages would be handled in separate issue
-- For dashboard, we are using hash based implementation for permalink instead of using REST based URL because there seems no logical entity which can be used after URL pattern `/projects/123/dashboard`. Still, implementation can be tweaked in future to load permalink from REST based URLs e.g. `/projects/123/status/486`


.travis.yml
```diff
@@ -26,7 +26,7 @@ deploy:
   skip_cleanup: true
   local_dir: dist
   on:
-    branch: dev
+    branch: ["dev" , "feature/permalink_for_feeds_messages"]
```
parthshah: please remove feature/... before merging to dev branch

vikasrohit: Yep, sure :)

PR #972, opened 2017-07-13
Feature/message edit delete

gondzo:


gondzo:
@vikasrohit can you check if the merge conflicts would revert any of the hot fixes deployed earlier to master.

PR #1112, opened 2017-08-11
small README fix

dmessing:
@vikasrohit just fixed a little of the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Install dependencies by running the following in the root of the project:
  - **Note:** You must use npm 3. Type `npm -v` to ensure you have a 3.x version.
 
 ## NPM Commands
-- To run locally, run `npm start` and head to `http://localhost:3000/search/challenges`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,7 +29,7 @@ Install dependencies by running the following in the root of the project:
  - **Note:** You must use npm 3. Type `npm -v` to ensure you have a 3.x version.
 
 ## NPM Commands
-- To run locally, run `npm start` and head to `http://localhost:3000/search/challenges`
+- To run locally, run `npm start` and head to `http://localhost:3000/new_project`
```
vikasrohit: It can be rather /projects which is kind of landing page for logged in user.

dmessing: ok whatever you want to make it, the only advantage to not logged in is someone can develop w/o accounts app, maybe we should add both?

vikasrohit:
Merged. Changes are anyway better than current content.

PR #1150, opened 2017-08-28
Sync up changes that include the Topic / Late user join fix...

mtwomey:


mtwomey:
On w/Phillip - were testing immediately after merge.

PR #1202, opened 2017-09-28
Feature/second level project category performance

vic-topcoder:
Adding styles for card fixes


circle.yml
```diff
@@ -19,7 +19,7 @@ compile:
 
 deployment:
   development:
-    branch: [dev]
+    branch: [dev, 'feature/second-level-project-category-performance']
```
vikasrohit: This needed to be removed we merge into dev. I have updated it directly in dev branch.

Issue #1064, opened 2017-07-27
[$30] Paragraphs are not divided in the 'PROJECT OUTLINE' section for 'Description'


sdgun:
Steps to Reproduce
-----------------------------------------------------------------

- Open the URL https://connect.topcoder-dev.com

- Login as a valid user: pshah_manager/topcoder123

- Click + New Project

- Fill all the required data

- Enter two paragraph description to the  'Description' section

- Check the 'PROJECT OUTLINE' section

Actual Result(s)
-----------------------------------------------------------------
- Paragraphs are not divided in the 'PROJECT OUTLINE' section for 'Description'

Expected Result(s)
-----------------------------------------------------------------
- Must divide the paragraphs in the 'PROJECT OUTLINE' section for 'Description'

URL where the Bug Appears
-----------------------------------------------------------------
https://connect.topcoder-dev.com/projects/?page=1&sort=createdAt+desc

Image/Video/JS Log/Console Log (If not attached here, Please check the comments section) 
-----------------------------------------------------------------
![37](https://user-images.githubusercontent.com/4592712/28668980-de50786c-72ef-11e7-8be1-9b0d621d4776.jpg)

Environment
-----------------------------------------------------------------
- **Device(s):** Laptop
- **Operating System:** Windows 7 Pro (64bit) 
- **Browser:** Google Chrome 59.0.3071.115 (Official Build) (64-bit) / IE11
- **User role (client, copilot or manager):** Manager
- **Account used:** pshah_manager/topcoder123

sirakovaDimitrina:
In order to get space between the two paragraphs in the right section like in the left section you need to separate the text in separate paragraphs. The text now is in one box and this is why there is no space between the paragraphs. 
![screen shot 2017-08-02 at 11 16 49 pm](https://user-images.githubusercontent.com/10530059/28893196-36792a46-77d9-11e7-8d26-d5a70a4b72e3.png)
See attached file

gondzo:
Changes merged.

Issue #1192, opened 2017-09-23
[$20] Code block styling


gondzo:
Using code blocks with large content will break page layout. We should probably force wrap on the <pre> tags in the posts, or add scrolling?
![image](https://user-images.githubusercontent.com/3051646/30773825-20bd5b6e-a079-11e7-962f-7a18c2e7df5e.png)

@vic-topcoder any ideas ?

vikasrohit:
Sorry to jump in @gondzo @vic-topcoder However, I do remember that we added horizontal scroll long time back to resolve this issue #619. I think it was done for comments only, may be we need to apply the same for post as well.

vic-topcoder:
Add horizontal overflow for code blocks.

HammadSiddiqui:
Can you give instructions to replicate this scenario?

gondzo:
In the project dashboard create a new post with a code block that has a
single long line (larger than page width)

HammadSiddiqui:
@gondzo 
While trying to create a new post, I am getting an error "Failed to create topic in Discourse: Unable to create discourse user"

gondzo:
Looks good

PR #1215, opened 2017-10-01
Fix for #1192 and #1064

HammadSiddiqui:
The following fix wraps issue #1192


HammadSiddiqui:
New commit also fixes issue #1064

Issue #1196, opened 2017-09-25
[$250] Upgrade Webpack to v3 (latest)


vic-topcoder:
We need to match the Webpack and stack to community-app as we currently use Webpack 1. There are great features like tree-shaking and ES6 native support, plus multiple optimizations. 

Created from Aha! https://topcoder.aha.io/features/TCCONNECT-36

gondzo:
Majority of changes should be made to https://github.com/appirio-tech/webpack-config repository (feature/connectv2 branch), as well as webpack.config.js in connect-app repo.

mhikeo:
@vic-topcoder @gondzo Can I request for access on the following?
* community-app
* https://topcoder.aha.io/features/TCCONNECT-36

gondzo:
What do you need community-app for ?
There is nothing in aha.io that isn't in this thread

mhikeo:
I thought we need to match how it is being done in community-app. Anyways, I'll proceed with the task. Thanks.

gondzo:
This is the community app repo > https://github.com/appirio-tech/topcoder-app

vikasrohit:
@gondzo community app which @vic-topcoder mentioned is https://github.com/topcoder-platform/community-app

gondzo:
@all prize increased and this is now available for pickup.

vic-topcoder:
@birdofpreyru - please advise if there's anything we can do to help align the system. I know we're already discussing it, would be nice to lay out what's needed in here for community
@mtwomey FYI

gondzo:
just following the webpack migration guide should be enough > https://webpack.js.org/guides/migrating/

birdofpreyru:
Well, just mechanical update to Wepback 3 can be done following the migration guide. Also, I'd say, break dependency on https://github.com/appirio-tech/webpack-config and do a local webpack config, properly split for dev / prod builds configurations.

But then, what we actually put into the config is related to other staff we discuss about styles, etc.

birdofpreyru:
Here is how Webpack is currently configured in Community App:
https://github.com/topcoder-platform/community-app/tree/develop/config/webpack

birdofpreyru:
Here is how Webpack is currently configured in Community App:
https://github.com/topcoder-platform/community-app/tree/develop/config/webpack

Most of its content is actually about making all stuff like CSS Modules, Hot Module Reloading in dev, inline svg, etc. to work together.

maxceem:
[...]

PR #1226, opened 2017-10-04
Fix Issue #1196

maxceem:
Fix for issue #1196 

- webpack updated to v3.6.0 and webpack-dev-server to v2.9.1
- `extract-text-webpack-plugin` replaced with `extract-css-chunks-webpack-plugin` from the fork of birdofpreyru
Seeing that webpack-config project has been using as a kind of universal config, there was a lot of stuff which is not used by connect-app, so I removed it:
- removed support for `jade`-like formats as they are not used in in the project
- removed `--tc` argument from webpack command, haven't found any usage in config
- removed `--silent` param inside webpack config as haven't found any usage in config
- removed `--test` param inside webpack config as webpack was never called with this param
- removed `DedupePlugin`
- removed `__MOCK__` variable from `DefinePlugin` as it never appears in the project

- `npm run test` and `npm run test:watch` don't work as it was before

- there is one small issue after migration:
  When we run the project in dev mode with hot reloading and make first change, hot reloading works, but first time it triggers components remount, next reloads don't
  it could be connected with the fact that react-hot-loader doesn't fully support react-router@3,
  migration to react-router@4 probably can solve it


gondzo:
@vic-topcoder @vikasrohit @mtwomey 
Looks like we need to update react-router from 3 to 4 to complete this migration to webpack 3 (otherwise hot reload doesn't work). IMO we shouldn't merge this until we have hot reload working properly. Any ideas ?

mtwomey:
@gondzo Can we do a paid task to do the upgrade for react-router?

gondzo:
Sure. Merging this one to test02 and I'll merge to dev once react-router is updated.

kevinkid:
Al be happy to work on it as private task .

gondzo:
It's assigned to a different member already.

kevinkid:
cool thanks

Issue #1212, opened 2017-09-29
Filter project types in admin view


vic-topcoder:
Fix all the missing project types listed in the admin view filter

![screen shot 1](https://user-images.githubusercontent.com/8484061/31104397-9ada89ca-a791-11e7-8868-865927bbcdc8.png)

Created from Aha! https://topcoder.aha.io/features/TCCONNECT-70

vikasrohit:
Changes done in feature branch. Waiting for the current milestone of 'chatbot' to be delivered first before getting this one merged into dev.

vic-topcoder:
Can we roll it with Chatbot release please?

vikasrohit:
Merged changes into dev for releasing it in current release.

PR #1238, opened 2017-10-09
Github issue#1212, Filter project types in admin view

vikasrohit:
— Fixed. Made the list of the project types to be dynamic. It was hard coded earlier because earlier we were showing different label in filters than what we have in the projectWizard/index.js


vikasrohit:
Merged into dev as per request here https://github.com/appirio-tech/connect-app/issues/1212#issuecomment-334523113

PR #1265, opened 2017-10-13
Promoting changes to production for release

vikasrohit:
"Real World Testing" updates and fixes

fyi @vic-topcoder @mtwomey


vikasrohit:
@vic-topcoder @mtwomey This release would have some of the community fixes as well. I am assuming that we are okay with those changes going to production.

PR #1291, opened 2017-11-06
Feature/lost ref code incomplete project

vikasrohit:
fyi @vic-topcoder @dmessing


vikasrohit:
@gondzo I think you copiloted the stuff where we removed webpack-config from dependencies.  Please let me know if the changes to `config/constants/index.js` makes sense to you.

gondzo:
Yeah, it makes sense to use dev for all branches except master and qa which have different values.   Good catch! @vikaroshit

vikasrohit:
Thanks @gondzo

PR #1293, opened 2017-11-08
Feature/project listing updates#1268

sachin-maheshwari:


src/components/LinksMenu/LinksMenu.scss
```diff
@@ -44,6 +44,3 @@
    color: $tc-dark-blue
   }
 }
-.panel {
-  box-shadow:none;
```
vikasrohit: Can this might have effect at other places where we are using LinksMenu.

vikasrohit:
LGTM 👍

PR #1329, opened 2017-11-16
Feature/project listing updates#1268

sachin-maheshwari:


vikasrohit:
@sachin-maheshwari I think we should keep `isManager` prop as well because there might be other components which are using this prop.

Issue #1317, opened 2017-11-15
[$30] Post composer/reply formatting options background issue


vic-topcoder:
The post composer formatting buttons have a white background on a slight gray background, but they should be transparent and blend with the background.
![screen shot 22](https://user-images.githubusercontent.com/8484061/32814274-24b6b23e-c963-11e7-8d32-e75f466b3148.png)

- [x] Remove background of formatting buttons

gondzo:
Thanks @Sande3p

Issue #1315, opened 2017-11-15
[$30] Project type icon should display appropriate tooltip on hover


vic-topcoder:
Hovering over the project type icon right now displays the icon name. It should be replaced by the appropriate project type - same as the project title from the project creation catalogue.

Current:
![screen shot 24](https://user-images.githubusercontent.com/8484061/32813450-1a667354-c95f-11e7-8f0a-af54e5ebe63e.png)

Expected: 
![screen shot](https://user-images.githubusercontent.com/8484061/32813572-ce96d51c-c95f-11e7-922c-a8045cd67a2d.png)

Must support all major browsers.

Sande3p:
verification: https://i.gyazo.com/2f5ddde72559c0291a92ef59d60db01b.png

gondzo:
Thanks @Sande3p

Issue #1311, opened 2017-11-15
[$50] Product cards are not responsive


vic-topcoder:
The product cards are fixed width and do not use the available space well. We might want to convert them to a responsive cards (min-width: 280px, max-width: 460px) to better utilize the space available.

sumitdaga:
do you mean these cards? https://goo.gl/RucM4i

Sande3p:
verification: https://i.gyazo.com/26c89000658bb3fbb6936c884492dca8.png

gondzo:
Thanks @Sande3p

Issue #1299, opened 2017-11-10
[$50] Add style and support for <sup> or <sub> text on Connect


vic-topcoder:
We need to format special characters as superscript/subscript in Connect and have no CSS styles for them. The case in question is the Watson Chatbot project.

- [ ] add support for `<sup>` and `<sub>` in Connect
- [ ] make sure the style works for any instance and makes the wrapped content 50% smaller:
`text-size: 0.5em;` and vertical position reflects the intended purpose`vertical-align: text-top`; or `vertical-align: text-bottom;`

Expected:
![screen shot 8](https://user-images.githubusercontent.com/8484061/32679879-f7774e9e-c61c-11e7-9d3a-454acdac3d89.png)
![screen shot 7](https://user-images.githubusercontent.com/8484061/32679886-fd842d3e-c61c-11e7-8f4f-ef05397df354.png)

vic-topcoder:
@gondzo For adding <sup> on the ® mark in Watson Chatbot we need to pass that in the string in Connect - can you make sure once we have the CSS support in place we can get this added to the title of the watson chatbot project? Thanks!

vic-topcoder:
@gondzo - what is the status of this one? Should we move it to the next milestone or it will be done within this one?

vic-topcoder:
Moved to track into a new release

gondzo:
@vic-topcoder Just to confirm, we need support for static and config strings only, not user input like project name, desctiption, etc ?

sumitdaga:
can you please explain what needs to be done ..i didn't get it exactly?

Sande3p:
verification: https://i.gyazo.com/58c9618b7d6c69070212e92337be42f5.png

gondzo:
@Sande3p We need to add sub/sup styles for the entire project (top level styles) and update the title at 
https://connect.topcoder.com/new-project/watson_chatbot to make the 'R' symbol a <sup>R</sup>. That config is in https://github.com/appirio-tech/connect-app/blob/602be8692eec5a36c493f3f239b3e439b11da553/src/config/projectWizard/index.js

Sande3p:
done.
verification: https://i.gyazo.com/4014614341b7af915a9e3047756b5934.png

Sande3p:
@gondzo please check the updated commit.

PR #1343, opened 2017-11-23
fix for #1317 #1311 #1315 #1299

Sande3p:


gondzo:
no need to create a pr to dev. In the future please fork the repo and create a PR to bugBash04 branch (so your work does not affect what other members are doing in parallel before it's merged)

PR #1354, opened 2017-11-24
fix updates for #1299

Sande3p:
This is the update for #1299. Next fix onwards would be using 'bugBase' as the base branch.
Thanks for the suggestion.


gondzo:
Thanks, looks good. I'll close the PR as the changes are already pushed to bugBash04 branch, and I'll merge all fixes to dev after the challenge.

PR #1360, opened 2017-11-27
Github issue#272, Validation error for custom validation is not working

vikasrohit:
— After this issue is fixed in react-components repo, updated project templates in connect app to get rid of unnecessary validationError field even though we have provided validationErrors object.

It does not add any feature or fixes any bug in connect app. Rather it adapts the uses of form component validations from react-components repo which got an issue about validation error, with custom validations, fixed via https://github.com/appirio-tech/react-components/issues/273.

**Verification:** After merging it in dev, to verify that it fixed the issue and didn't create any regression, we should be able to see validation error 'Please provide a description' when user clears out the description field after entering some value first.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -357,7 +357,7 @@
       }
     },
     "appirio-tech-react-components": {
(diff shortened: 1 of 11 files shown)


vikasrohit:
fyi @sachin-maheshwari @vic-topcoder @mtwomey

PR #1395, opened 2017-12-01
Issue#1359 spurious avatars render

sachin-maheshwari:


src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -4,6 +4,16 @@ import React, { PropTypes as PT } from 'react'
 import './AvatarGroup.scss'
 
 function AvatarGroup({ users }) {
+
+  const uniqueUserList = []
+  const uniqueUsers = _.filter(users, user => {
```
vikasrohit: @sachin-maheshwari we can use https://lodash.com/docs/4.17.4#uniqBy instead of writing our version. :)

sachin-maheshwari: done. Thanks @vikasrohit

Issue #1368, opened 2017-11-29
[$150] Fix text area auto-sizing


vic-topcoder:
We use a react plugin to auto-size text areas on project specification page. Right now the plugin doesn't work all the time.

 

**Goal** : All text areas should be sized to fit all the text without scrolling in all states (normal, active, disabled).

 

Current state:

Screen Shot.png (https://topcoder.aha.io/attachments/6493680258833768883/token/2f573aa94784c7f9e3b0e71e0e7ee39c402994845aec7a8cbc0331cefa1ddf88.download?size=original)

Created from Aha! https://topcoder.aha.io/features/TCCONNECT-111

zjuasmn:
PR submitted. https://github.com/appirio-tech/connect-app/pull/1413

vic-topcoder:
- [ ] - The text area is still smaller than the text and there's a tiny scroller. The more text lines there are, the bigger the difference - something is not quite right in pixel count.
![screen shot 4](https://user-images.githubusercontent.com/8484061/34190531-2d80c7fc-e4f6-11e7-83bc-d49cb2d9c758.png)


- [ ] - The size is lost after reloading the page - the text area should always keep the size consistent to its contents upon page reload/save/etc. events.
![screen shot 3](https://user-images.githubusercontent.com/8484061/34190523-26d9456e-e4f6-11e7-919d-3c0890c4e278.png)

vic-topcoder:
Do we have ETA on this one?

gondzo:
@ThomasKranitsas any ideas on the cause here ?

ThomasKranitsas:
Sorry, I'm still investigating... I tried few hacks to fix it but none of those worked.

vic-topcoder:
Ok, let's move it for CF6

ThomasKranitsas:
Ok, I'll focus on the other tickets

kevinkid:
I may have a solution for this @ThomasKranitsas @vic-topcoder on [this](https://github.com/appirio-tech/connect-app/pull/1574/commits/7892190cdc83d882abc8459989e50e462abbb09b) commit. Turns out the default React textarea already does the auto-resize.

ThomasKranitsas:
This breaks the `Notes` fields in the `New project` page.

ThomasKranitsas:
Actually, it breaks the `Notes` field everywhere 🙂 I'll see if I can do any hack to fix it

kevinkid:
[...]

PR #1413, opened 2017-12-05
#1368 Fix text area auto-sizing

zjuasmn:
Add a AutosizeTextArea to fix the resizing issue. It would be better to fix the original package if someone has access to that repo.


gondzo:
@zjuasmn this does fix it in active state, but there is still the same issue when viewing a disabled text area
image

vic-topcoder:
@gondzo what is the status on this one?

gondzo:
This one does not fix the issue and there is no further update by the author. closing this one.

Issue #1374, opened 2017-11-29
[150$] Update project filters


vic-topcoder:
1. Update the design of the project filters to match the new UI (darker background.
2. Replace the current drop-down with react-select which allows for multiple selection (see design).
3. Remove the project status (display it as tab); replace it with project segment (we don't have the data yet, but we are going to have a new tag per project - segment - that would be used to filter projects by.

Final design from here:  [https://drive.google.com/drive/folders/1AbIzCycuhTBh7WNCiaD0mYVMXjeGSdd3?usp=sharing](https://drive.google.com/drive/folders/1AbIzCycuhTBh7WNCiaD0mYVMXjeGSdd3?usp=sharing) This is a sketch live design, make sure to download the folder and open the index.html to explore all the artboards.

transition-updates-project-grid-full-list---admin-copy.png (https://topcoder.aha.io/attachments/6493701114636916906/token/1edaf3c0b0a39c59bd0cec742b977ceb2db90c66a5042aaa1a5a55965617f4ac.download?size=original)

Created from Aha! https://topcoder.aha.io/features/TCCONNECT-114

aretheregods:
@gondzo I can finish this, but it's nighttime in New York. Although I could do all of this tonight, it would be best if I finish it tomorrow morning (scheduled events tonight and a need for sleep), which would be outside of the 12 hour window. I neglected accepting the ticket, just to make sure I could actually do it expediently, but I'm not going to have time to deliver quality code unless I do the rest in the morning. So, I'm gonna accept the ticket and just finish in the morning and submit the pull request. I hope it's not unassigned before I get a chance to get back to it.

gondzo:
ok, just keep me updated

aretheregods:
@gondzo I'll get back to working on at 8 a.m. EST, and should be finished not too long after that. As you asked, I'll keep you updated. If I encounter anything prohibitive, I'll just self-unassign.

aretheregods:
@gondzo There are two things of which I'm not sure. First, what options if any should I add to the project segment dropdown? Basically, what are the project segments that should be included in this list? Second, since there's a separate issue for displaying project-statuses as tabs, am I correct in assuming adding that functionality isn't in scope for this issue? Am I just supposed to remove the status dropdown from the filters?

aretheregods:
It's taking me much longer to style the new select menus than I anticipated, so I've reopened this for pickup.

maxceem:
[...]

PR #1428, opened 2017-12-07
Issue 1374 - Update project filters

maxceem:
# Issue #1374 -  Update project filters

## Implementation notes

- I keep segments list empty so far. I cannot populate it with fake values because API requests will fail as `segment` is not available option for filtering.

- `react-select@0.9.1` is being used in the module `appirio-tech-react-components` though the latest version available is `1.1.0`.
  I have to keep using the older version as styles from the old version are included in `main.css` in `appirio-tech-react-components` module. If I try to use the latest version for this task, styles from old and new version will interfere and break each other.

  Old version works correctly, but produces the next warnings in the browser console:
  ```
  Warning: Unknown prop `minWidth` on <input> tag. Remove this prop from the element. For details

  warning.js:35 Warning: AutosizeInput: isMounted is deprecated. Instead, make sure to clean up subscriptions and pending requests in componentWillUnmount to prevent memory leaks.

  warning.js:35 Warning: Select: isMounted is deprecated. Instead, make sure to clean up subscriptions and pending requests in componentWillUnmount to prevent memory leaks.
  ```

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4,6 +4,18 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "Base64": {
+      "version": "0.1.4",
+      "resolved": "https://registry.npmjs.org/Base64/-/Base64-0.1.4.tgz",
+      "integrity": "sha1-6fbGvvVn/WNepBYqsU3TKedKpt4="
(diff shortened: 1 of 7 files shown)


gondzo:
Merged to bugBash05. Closing this one.

PR #1429, opened 2017-12-07
fixed #1352 taking constant for project listing chunk size.

### LINKED ISSUE #1352: Extract constant for page size of load projects calls
Right now its hardcoded anywhere we are passing page size. We need to extract it as constant, it would be easy to manager later when we would make page size dynamic as well.

sachin-maheshwari:


src/api/projects.js
```diff
@@ -1,13 +1,13 @@
 import _ from 'lodash'
 import { axiosInstance as axios } from './requestInterceptor'
-import { TC_API_URL } from '../config/constants'
+import { TC_API_URL, PROJECT_LIST_CHUNK } from '../config/constants'
```
vikasrohit: I would prefer `PROJECT_LIST_PAGE_SIZE` and for now we can use the same constant for card view page size.

sachin-maheshwari:
@vikasrohit , sure..I'll check and make changes accordingly. 

I observed some abnormal behavior of corresponding back-end API, which is giving maximum 20 results  . If it is known behavior, need to tune in front-end according - otherwise pagination functionality at front-end will be broken for value PROJECT_LIST_PAGE_SIZE > 20 (as we are using same constant in offset). https://api.topcoder-dev.com/v4/projects/?limit=50&offset=0&fields=id,name,description,members,status,type,actualPrice,estimatedPrice,createdAt,updatedAt,details&sort=updatedAt+desc

vikasrohit:
Yes 20 is the max page size that any one can pass to the API. I don't think that has any effect front end except that we should never set page size > 20. So, you can add comment to the newly introduced constant about the same.

vikasrohit:
LGTM 👍

Issue #1445, opened 2017-12-14
Update the customer tooltip component


vic-topcoder:
Right now we show some additional data that doesn't exist. Until that point let's remove the extra data from SFDC (it is hardcoded in the component).

Remove:
``` html
<div className="sf-data-bottom-container">
                    <div className="segment-data">
                      <span>Wipro Digital Jaipur / Topgear</span>
                    </div>
                    <div className="segment-data">
                      <span>0141-2211258</span>
                    </div>
                    <div className="segment-data">
                      <a href="https://google.com" target="_blank" rel="noopener noreferrer" className="ext-link">SDFC Lead Page</a>
                    </div>
````

Also, ProjectGridView should not hardcode the tooltip, but reuse a stand-alone component. At the moment it is like this:
![screen shot 10](https://user-images.githubusercontent.com/8484061/33971733-ecaedc44-e02e-11e7-8ba4-5c1a6d09bd2b.png)

Please use the tooltip components we have and simplify the components. Ideally this rendering component should be your container, and you'd reuse other components - tooltip, avatars, etc.

ThomasKranitsas:
Is there anything else pending on this from my side?

vic-topcoder:
We have to update the tooltips to use the stand-alone component instead of generating the content on the grid itself, I guess. But that'll be another task.

Issue #1444, opened 2017-12-14
Add tooltip hint for manager user avatar as well


vic-topcoder:
We don't have a hover hint tooltip for the project managers. Please reuse the tooltip hint for customer.

Depends on fixing #1445 first

ThomasKranitsas:
@vic-topcoder how do you want to show multiple managers?

ThomasKranitsas:
@vic-topcoder @gondzo 

1. Should I show one tooltip for all managers or one for each manager?

2. What details should I show? (avatar, first/last name, handle)?

gondzo:
1. I think one for each manager is better
2. Avatar, full name, handle should be enough

vic-topcoder:
We have double avatars after the tooltip, plus we want to have the email address on each user, especially customers. The email is the most needed piece of data for the connect managers.
![screen shot 5](https://user-images.githubusercontent.com/8484061/34226020-3a743ac0-e57d-11e7-8ccd-ec5e3379ce79.png)

@ThomasKranitsas - please add the email address back

ThomasKranitsas:
Oh, sorry! I've added double avatars to check how it looks with 3 managers... I forgot to remove it.

The email should be there if it's available.

ThomasKranitsas:
@vic-topcoder should be fixed now. Sorry about that! 🙂

The code for the email is here: https://github.com/appirio-tech/connect-app/blob/bugBash05/src/components/User/UserTooltip.jsx#L38
If the user details object contains the email, the tooltip will show it.

vic-topcoder:
But aren't all users registered with an email? I don't see it still. @gondzo - why the user details object may not contain email?

gondzo:
we have to add email to the fields we want from the api >
https://github.com/appirio-tech/connect-app/blob/dev/src/api/projectMembers.js#L8
- I think it's just add 'email' to the list but I'm trying to find the api
docs to confirm

vic-topcoder:
This is the tooltip:
![screen shot 6](https://user-images.githubusercontent.com/8484061/34228764-bcefe518-e586-11e7-90b7-2d435221a91e.png)
This is the data object:
![screen shot 1](https://user-images.githubusercontent.com/8484061/34228791-d4203c24-e586-11e7-9bcd-6b57c69ef290.png)

I would expect to see "connect_dev_manager@mailinator.com"

vic-topcoder:
I think we were getting the email before this changes - check the old tooltip code

[...]

PR #1492, opened 2017-12-20
Update UserTooltip component

ThomasKranitsas:
related issue #1445

@gondzo can you check this one so I can move to #1444 ?


ThomasKranitsas:
Actually, I'll push another commit here to fix both tickets

gondzo:
looks good.

Issue #1508, opened 2017-12-22
Dismissing a notification from the dropdown closes the menu


vic-topcoder:
### Expected behavior
Dismissing a notification within the dropdown menu should reorder the items without closing the menu. The items underneath the notification should move up to take the available space.

### Actual behavior
Dismissing a notification closes the notifications dropdown menu

### Steps to reproduce the problem
- Open the notifications menu
- Dismiss a notification

maxceem:
Quick fix. Just bring back the old custom implementation of Dropdown which has performance issues.

vic-topcoder:
@maxceem - what are the performance issues of the old Dropdown? Is is safe to bring it back? How can we fix the dismissal of the menu with the component currently used?

vikasrohit:
I would have suggested to fix the optimized implementation as it was implemented to reduce tech debt and now we have again introduced tech debt.

PR #1520, opened 2017-12-22
issue #1508 - Dismissing a notification from the dropdown closes the menu

maxceem:
issue #1508 - Dismissing a notification from the dropdown closes the menu
Quick fix. Just bring back old custom implementation of Dropdown, which has performance issues.


vic-topcoder:
Discussed with @gondzo

Issue #1519, opened 2017-12-22
[$100] Screen flicker in dropdown


gondzo:
Found a weird issue with notifications dropdown.
When hover with mouse to the dropdown list global browser scrollbar disappears which causes flickering http://take.ms/g53EJ
@vikasrohit Is there anything in changes to the Dropdown component that could cause this (the one that was moved to react-components repository)?

Reported by @maxceem

vikasrohit:
It can be @gondzo However, I would like to know if this does not happen without those changes.

gondzo:
It doesn't > http://connect-test02.topcoder-dev.com
@maxceem is looking into possible (quick) fixes to get this before the release

maxceem:
I've checked it in the notification branch where the dropdown component was in the `connect-app`. Before moving to `react-components`.
This issue with the flickering was not there.
Also, this issue was fixed https://github.com/appirio-tech/connect-app/issues/1508 in that version.

This issue https://github.com/appirio-tech/connect-app/issues/1497 presents in the version before moving.

vikasrohit:
I have observed that when you move out from the header of the notifications dropdown, it does not flicker. So, IMO, it is something related with internal scroll bar and its corresponding div.

vikasrohit:
Its a class `noScroll` that is being added as soon as mouse enters the `notifications-dropdown-body` section.

maxceem:
Oh yes, so this issue is not related to the replaced dropdown component.
The new code was added to `NotificationsDropdownContainer` to prevent scrolling browser window when scrolling inside the notifications drop-down reaches the end http://take.ms/QBp1Z

Though now this flickering is introduced.

I propose to remove the code which adds `noScroll` class for the current release.

After release, this could be implemented different ways. In my opinion, the best way is to `preventDefault`/`stopPropagation` when we reach the end scrolling area inside the notifications drop-down.
Also, there is a decision to make. If there is no scroll in notification dropdown, should we block global scrolling or no?

vikasrohit:
Agree with @maxceem for getting rid of `noScroll`.

However, @maxceem I think we can always do `stopPropagation` in notification container to prevent scroll event from moving up.

maxceem:
@vikasrohit amazing idea with `stopPropagation`, though I've checked it and it doesn't work :-(.
In a good case, it's just because how react handles `onWhell` event. So we still have a chance to add a native listener with `addEventListener` and control `useCapture`.
[...]

PR #1521, opened 2017-12-22
issue #1519 - Screen flicker in dropdown

maxceem:
issue #1519 - Screen flicker in dropdown
Removed noScroll class


vic-topcoder:
The reason for adding the "noScroll" is to prevent double scrolling when the notifications dropdown is shown. Before that reaching at the end of the notifications on scroll would start scrolling the page under the menu. 

The jumping is caused by the showing/hiding of the browser scrollbar. I'll put the top bar above everything so the container would never need to have a scrollbar and thus the behavior would be fixed. This is a CSS problem to have, I'm on it. Let's hold with reversal of the code for the moment.

Issue #1497, opened 2017-12-21
Dropdown menu doesn't take viewport bounds into consideration.


vic-topcoder:
The drop-down should take into account the viewport bounds. Right now it is positioned via static CSS and gets cut at the end of the project grid.

Current:
![screen shot 4](https://user-images.githubusercontent.com/8484061/34233684-bd636f28-e59b-11e7-8c9e-0028dcd373bd.png)

Expected:
The menu would show on top of the click target when the viewport bounds would clip it with standard position.
![screen shot 5](https://user-images.githubusercontent.com/8484061/34233704-d9b6080c-e59b-11e7-8a74-d7b3c45773be.png)

Desired behavior: https://reactstrap.github.io/components/dropdowns/ - check the menu when the trigger (button) is very close to the browser borders.

Another example is https://tympanus.net/codrops/2013/05/23/tooltip-menu/; check the demo.

Ideally the solution for positioning the dropdown menu would be the same used for displaying tooltips.

Requirement - follow TC SCSS/HTML/JS best practices.

maxceem:
As a quick solution, I propose to make the next. When we click in the top half of the screen - open dropdown **down**. If we click in the bottom half of the screen - open dropdown **up**, check video http://take.ms/QJfs5

To make it work more elegant way when it's open up only if it cannot fit down it's better to use some existent implementation of dropdown as there are a lot of edge cases usually.

PR #1523, opened 2017-12-22
Issue #1497 - Dropdown menu doesn't take viewport bounds into consideration.

maxceem:
Issue #1497 - Dropdown menu doesn't take viewport bounds into consideration.

Quick solution: always dropdown “upside” if click in the bottom half of the window.

### DIFF
--- a/src/components/ProjectStatus/editableProjectStatus.js
+++ b/src/components/ProjectStatus/editableProjectStatus.js
@@ -24,7 +24,7 @@ const hocStatusDropdown = (CompositeComponent) => {
       return false
     }
 
-    render(props) {
+    render() {
       const { canEdit, isOpen, handleClick, onItemSelect, showText, withoutLabel, unifiedHeader, status } = this.props
       const selected = PROJECT_STATUS.filter((opt) => opt.value === status)[0]
 
--- a/src/components/ProjectStatus/editableProjectStatus.js
+++ b/src/components/ProjectStatus/editableProjectStatus.js
@@ -11,6 +11,68 @@ import {
   PROJECT_STATUS_CANCELLED
 } from '../../config/constants'
 
+const hocStatusDropdown = (CompositeComponent) => {
+  class StatusDropdown extends Component {
+    shouldDropdownUp() {
+      if (this.refs.dropdown) {
+        const bounds = this.refs.dropdown.getBoundingClientRect()
+        const windowHeight = window.innerHeight
+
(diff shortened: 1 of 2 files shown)


src/components/ProjectStatus/editableProjectStatus.js
```diff
@@ -11,6 +11,68 @@ import {
   PROJECT_STATUS_CANCELLED
 } from '../../config/constants'
 
+const hocStatusDropdown = (CompositeComponent) => {
+  class StatusDropdown extends Component {
+    shouldDropdownUp() {
```
vikasrohit: @maxceem I know it is late as changes are already merged. However, is it possible to move this logic, of showing dropdown up or down, can be moved to the generic dropdown component or its HOC? That way all dropdowns would get benefit from it.
fyi @vic-topcoder

maxceem:
Sorry @vikasrohit I'm not sure I'll have time untill 2nd January for this.
That was done just to make quick fix for release.

Currently there are three implementations in use: from react-components,
Dropdown component inside connect-app, and some simple inline
implementations, like in this case.

I believe the best way to go with dropdowns further is to use some existent
implementation for all dropdowns cases. There are usually a lot of edge
cases with custom UI components like Dropdown. Also we can get support for
mobile devices, keyboard control and accessibility if we incorporate some
existent solution with long history and active development.

On Dec 23, 2017 13:14, "vikasrohit" <notifications@github.com> wrote:

*@vikasrohit* commented on this pull request.
------------------------------

In src/components/ProjectStatus/editableProjectStatus.js
<https://github.com/appirio-tech/connect-app/pull/1523#discussion_r158576667>
:

> @@ -11,6 +11,68 @@ import {
   PROJECT_STATUS_CANCELLED
 } from '../../config/constants'

+const hocStatusDropdown = (CompositeComponent) => {
+  class StatusDropdown extends Component {
+    shouldDropdownUp() {

@maxceem <https://github.com/maxceem> I know it is late as changes are
already merged. However, is it possible to move this logic, of showing
dropdown up or down, can be moved to the generic dropdown component or its
HOC? That way all dropdowns would get benefit from it.
fyi @vic-topcoder <https://github.com/vic-topcoder>

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/appirio-tech/connect-app/pull/1523#pullrequestreview-85417137>,
or mute the thread

Issue #1532, opened 2017-12-23
Console errors


vic-topcoder:
### Expected behavior
No console errors

### Actual behavior
I see 2 errors and 1 notification

![screen shot 18](https://user-images.githubusercontent.com/8484061/34316419-1b144980-e74a-11e7-8f70-cda3c82b5ca0.png)
![screen shot 19](https://user-images.githubusercontent.com/8484061/34316420-1b25d628-e74a-11e7-9dea-bbc70b1e8c83.png)
![screen shot 17](https://user-images.githubusercontent.com/8484061/34316421-1b36f476-e74a-11e7-9801-96cf7f5e907b.png)

Latest Localhost. Please verify

ThomasKranitsas:
I replaced the `PropTypes` with from the core `react` module with the `prop-type` module in 137 files but still getting those warnings... Those are from some outdated components from the `appirio-tech-react-components` module.

I'm fixing the errors right now and I'll create a PR

vic-topcoder:
Cool, thanks!

PR #1552, opened 2017-12-26
Use PropTypes from prop-types instead of react core module and fix errors

ThomasKranitsas:
#1532

@gondzo I fixed most of the errors I found and I changed all components to use `PropTypes` from 'prop-types` instead of the `react` core module.


gondzo:
good job 👍

Issue #1424, opened 2017-12-06
[$100] Remove dependency of old/obsolete technology


vic-topcoder:
We still have dependencies for Coffescript, Jade, etc. We should update the repo to not use any obsolete technologies

Created from Aha! https://topcoder.aha.io/features/TCCONNECT-42

maxceem:
Connect-app depends on [appirio-tech-react-components](https://github.com/appirio-tech/react-components/tree/feature/connectv2) which uses coffee scripts.

`appirio-tech-react-components` is being utilized in 58 places in the connect-app so we cannot stop using it.

It looks like the best way to go here is to implement building CoffeeScript to JavaScript inside `appirio-tech-react-components` and publish it as an npm module. So when it's being used as an npm module it only exposes js files. 
Or we can simply keep built js files in the git repo without publishing to npm.

What do you think?

gondzo:
if we're still using coffescript then let's not remove the dependency at all for now.
@vikasrohit any drawbacks to publishing as npm module? can we make such a change to the repo, as there are other apps using it?

vikasrohit:
First things first. I am not sure why we are considering coffeescript and jade as old and legacy. They in fact relatively new. And I found `jade` in particularly concise and helpful in development. `coffeescript` is good way to work on Javascript for `python` guys.

Now if we think it is still necessary, publishing it as npm module has one problem. We started this repo to have common components across all topcoder apps. But later on we deviated from having common and consistent components across topcoder applications. So, essentially it started to have separate branches for different applications. Right now it has two different branches which are used, one is master which is used for community facing apps and connectv2 branch used for Connect. So, if we publish it as npm module we might have to publish two different major versions e.g. 1.x for connect and 2.x for community apps.

vic-topcoder:
@vikasrohit @gondzo @birdofpreyru @mtwomey Can we migrate towards the components used in Community-app, as it is actively under development, and unify with them? This would require us to evaluate both apps and synchronize the code between the two. I guess that would be a bigger project than just refreshing the component repo that we have right now. I would love if we end up with a npm module or component library which is unified across all apps. The fragmentation of space and languages used creates a lot of troubles maintaining and supporting the development.

As I see it, we can start the transition with unification of components in smaller batches, until we replace all components. This could be done in several bug hunts.

I'm interested to hear your thought on this guys.

birdofpreyru:
[...]

PR #1590, opened 2018-01-06
Fix for Issue #1424

kevinkid:
Coffee script isn't a local project dependancy, please let me know if there are any issues.


gondzo:
didn't upgrade react version and build is failing

Issue #1469, opened 2017-12-17
[$400] User mentions in the posts


gondzo:
Should we use user handles for generating mentions or full names ?

vic-topcoder:
It should be `@username` format that ideally would be replaced with the `Full Name` to be readable - Facebook like. The `@username` mentions are not yet available in the system - save for later, but let's think about how we can solve this.

gondzo:
If we use @handle for user mentions and display full name in the notifications, I'm not sure it will be obvious for the users that we have this feature. maybe we need to add mentions support to the post editor, similar to how it's implemented in github ?

vic-topcoder:
I was aiming for the way it is implemented in Facebook and Dribbble - upon using the `@username` the system replaces/renders the mention as `First name Last name`. Check this demo from draftjs - https://www.draft-js-plugins.com/plugin/mention - I'm fine if we use this plugin as is. Remember - in Connect the names are more important for our customers than @username, and they don't use the latter that often, or at all. The goal is to make it easy for people to communicate :)

gondzo:
This requires several steps:
- add mentions plugin to post editor (allow mentions for project members only)
- add mentions parsing to notifications server
- add mentions notifications to connect app

vikasrohit:
@gondzo I would suggest if we can keep `@username` in the actual message being posted to the message api. And replace it with user's full name when fetching the messages from the API. This is to allow showing latest value of the user's full name. I know user's full name is not something which can change frequently, however, I am thinking from the perspective of having a generic approach to replace placeholders in the messages e.g. we can have `${project_title}` as one of placeholder which can be resolved to the latest project title when fetching the messages.

gondzo:
that would be fine in the app, but the the emails would be sent with @username instead of full name

vikasrohit:
We have to apply the same process of replacing the placeholders, which we would do in API while fetching messages via API, before sending the emails as well. I know right now it is not possible because discourse is doing that, but we can do this once we get rid of discourse and take control of sending emails.
I guess, for now you can proceed with your current approach, we can revisit it later after getting rid of discourse.

passimm:
I have made a PR: https://github.com/appirio-tech/connect-app/pull/1594
1) use draft-js-mention-plugin in post editor
2) use "@F_NAME L_NAME" in textbox
3) replace with @handle before sending to backend

passimm:
notifications server code change will be sent out later..

[...]

PR #1594, opened 2018-01-09
Issue #1469 - User mentions in the posts

passimm:
add draft-js-mention-plugin to post text editor.
use @'F_NAME L_NAME' in the text box. 
replace them with @handle before posting to backend


passimm:
updated.
select handle and render its fullname

PR #1597, opened 2018-01-10
Issue #1469 - User mentions in the posts

passimm:
- use draft-js-mention-plugin in post text. select project member by handle and render in fullname
- add new type of notification for 'you're mentioned in a post'


passimm:
@gondzo updated per your comment.
added new event type: notifications.connect.project.post.mention

Issue #1615, opened 2018-01-19
Adding member without image (avatar component) isn't rendered correctly in auto-suggest box


vic-topcoder:
### Expected behavior
The avatars should always be rendered correctly.

### Actual behavior
Auto-suggest doesn't list the avatars (doesn't use the avatar component) and thus when a user with no image is selected, the avatar is broken

### Steps to reproduce the problem
- Search for user to add
- Select one without a picture
- The avatar is broken

Also observe - the auto-suggest doesn't use the correct avatar

After adding the user, the avatar renders correctly.

### Screenshot/screencast
![screen shot 2018-01-18 at 5 56 31 pm](https://user-images.githubusercontent.com/8484061/35131148-48fc63c0-fc79-11e7-93a0-7a453d4b70b5.png)
![screen shot 2018-01-18 at 5 59 12 pm](https://user-images.githubusercontent.com/8484061/35131156-4d04c142-fc79-11e7-8a10-c7ef33e3205b.png)

PR #1642, opened 2018-01-26
Issue #1615 Adding member without image (avatar component) isn't rendered correctly in auto-suggest box

tpwwu:


src/components/TeamManagement/AutoCompleteInput.jsx
```diff
@@ -64,7 +66,11 @@ class AutoCompleteInput extends React.Component {
 
         <span>
           <InputIcon className="input-icon"/>
-          {selectedNewMember && <img src={selectedNewMember.photoURL} />}
+          {selectedNewMember && (
+            selectedNewMember.photoURL
+              ? <img src={selectedNewMember.photoURL} />
+              : <img style={defaultMemberPhotoStyle} src={require('../../assets/icons/users-16px_single-01.svg')} />
```
gondzo: let's not use inline require for images. please use `import icon from ...svg ` on top of the file for all imports

tpwwu:
@gondzo fixed

Issue #1437, opened 2017-12-13
[$200] Page update reload suggestion


vic-topcoder:
Detect when a page has been updated while the user is browsing, and display a prompt to reload the page to update with the latest content. Specifically useful for discussions on the home feed.

Created from Aha! https://topcoder.aha.io/features/TCCONNECT-9

gondzo:
we don't have any strategy for this yet, but with changes for #1525 we could refresh the project page if there is a new notification for the project being displayed

gondzo:
@vic-topcoder does the above approach look fine? any suggestions on what to do if user has unsaved changes in a form (posting comment, editing spec, etc)?

vic-topcoder:
Here is the prototype for the interaction: https://invis.io/B9997WQJW#/202782585_Refresh_Feed_-_Step_01

Should function the same way as in Facebook/Google. 

- When on top of a page and new update is present, reload page
- When scrolled in a page, display reload suggestion button
- When posting or using dropdown menus, do not change UI until action is completed, after that follow the above rules.

gondzo:
Prize increased

passimm:
hi gondzo, just sent a PR for this: https://github.com/appirio-tech/connect-app/pull/1647/commits
your comments are welcome. thanks

passimm:
just realized i forget to take care of
"When posting or using dropdown menus, do not change UI until action is completed, after that follow the above rules."

i will update the PR later...

passimm:
PR updated.
have one pending issue. it's difficult to track whether Notification/user profile dropdown is expanded. they are using appirio-tech-react-components:Dropdown, which neither allows add click listener, nor registered to react-redux... could you shed light on how to track whether Dropdown is opened?

passimm:
hi @gondzo any idea about that?

gondzo:
I think we can ignore the dropdowns for now and fix that later

passimm:
@gondzo PR updated. please help review. thanks!

gondzo:
merged to bugBash07. closing

PR #1647, opened 2018-01-27
Issue #1437 - Page update reload suggestion

passimm:


gondzo:
@passimm 
- The feed is reloaded even if I'm composing a new topic or comment - it should not do anything in that case
- if I scroll the page, the reload button is "displayed" but is not visible as it's on top of the feed. it should have absolute positioning in the feed container
- if there are new notifications when I load the page, it will still trigger a reload after the first timer

passimm:
@gondzo . fixed above issues.

passimm:
@gondzo . fixed above issues.

passimm:
Hi @gondzo  any update about this?

PR #1675, opened 2018-02-04
Minor tweak to status textarea

arun1595:
Adds a bit of top padding to the textarea.

### Before

image

### After

image


vic-topcoder:
This is not to our UI Kit standards. When you use padding/margin/size/line-heihgt, it is always a multiple of our base unit (= 5px).

vic-topcoder:
I'd just use $base-unit for padding inside input/text area elements. For the time being I don't think it is needed, I want to review the post composer at a later stage and fix it as a whole. Thank you for looking into it! We have more design-relates issues coming in.

arun1595:
@vic-topcoder Sounds good, thanks!

PR #1712, opened 2018-02-28
Used Common method defined for finding files section title

RishiRajSahu:
Removed redundant code from generic_design template.


vikasrohit:
We can push this change later with next release.

Issue #1738, opened 2018-03-01
[$50] Window-Manger – When we select the same project status application should not show message to user as “Project updated”


vic-topcoder:
_From @youdare on February 18, 2018 6:19_

Summary:
Window-Manger – When we select the same project status application should not show message to user as “Project updated”
### Reported URL ###
 https://connect.topcoder-dev.com/
### Reproduction Steps ###
1. Go to  https://connect.topcoder-dev.com/
2. Log in with pshah_manager/topcoder123 and Verify the filter order
3.Go to any project and select the same status
Example: If project in draft select draft

### Expected Result ###
Window-Manger – When we select the same project status application is shows message to user as “Project updated”
### Actual  Result ###
Window-Manger – When we select the same project status application should not show message to user as “Project updated”

 ### Screenshots ###

See comments

### Issue Information ###

Frequency: Repeated

Device: Window

OS: Windows 10
Browser: Chrome 

_Copied from original issue: topcoderinc/connect-test-cycle-1#180_

vic-topcoder:
_From @youdare on February 18, 2018 6:21_

https://youtu.be/6FRDMb8GzAc

gondzo:
we're not showing any messages when updating project status. closing.

PR #1839, opened 2018-03-08
issue #1738 fixes

anonymousjaggu:
fixes for  #1738 and indent fixes


gondzo:
The expected fix here is not to update the project at all if the selected status is the same - not update the project and not show the alert

gondzo:
@anonymousjaggu can you take a look and submit a new PR?

PR #1840, opened 2018-03-08
fix #1708: Tooltip inside project dashboard is cropped by the sidebar

### LINKED ISSUE #1708: [$30] Tooltip inside project dashboard is cropped by the sidebar
### Expected behavior
The tooltip should be above all elements and not be clipped by the parent container (position: absolute; z-index: 100) 

@gondzo - what was the z-index order table? Let's put that as a reference on the Wiki please.

### Actual behavior
Hover over the avatar on dashboard to observe.
screen shot 2018-02-26 at 6 25 04 pm

zjuasmn:
fix #1708

### DIFF
--- a/src/projects/detail/Dashboard.scss
+++ b/src/projects/detail/Dashboard.scss
@@ -18,7 +18,6 @@
   .sideAreaWrapper {
     border: 1px solid $tc-gray-20;
     border-radius: 4px;
-    overflow: hidden;
   }
 
   .dashboard-container {


gondzo:
@zjuasmn I think the proper solution here is to change the z index for the tooltip, not remove the overflow property

zjuasmn:
@gondzo I try change the z index to 99999 but not works, since one of its offset parent set 'overflow hidden'. Changing the z-index could show one element covered by non ancestor I think the proper way is to remove the overflow property.

gondzo:
@zjuasmn .left-area also has that overflow:hidden property but still doesn't solve the issue - right container will still hide part of the tooltip. can you please take a look at that
image

PR #1877, opened 2018-03-09
fix #1683: Don't hide "Old" notifications when there are no notificat…

### LINKED ISSUE #1683: [$60] Don't hide "Old" notifications when there are no notifications displayed in the notifications drop-down
### Expected behavior
I want to see all notifications I haven't dismissed in the notifications drop-down.

If there are older notifications that are not displayed, when I dismiss my notifications I want to see them into the drop down instead.

The dropdown menu should never be empty when there are notifications that are not dismissed.

### Actual behavior
The dropdown menu shows nothing, but there is a link for "Older" notification.

screen shot 2018-02-01 at 7 29 54 am

At the notification center I can't see any of the notifications - all are hidden.
screen shot 2018-02-01 at 7 24 29 am

zjuasmn:
…ions displayed in the notifications drop-down

fix #1683


gondzo:
I don't think this one is correct - we should simply show 5 latest notifications in the dropdown and the rest can be viewed in the notifications center by clicking the view older notifications

gondzo:
@zjuasmn can you please update the PR?

PR #1949, opened 2018-04-10
v5 implementation for tc-notifications endpoints change.

sachin-maheshwari:


vikasrohit:
@sachin-maheshwari I guess we need to move v5 related changes as hotfix instead of a regular dev => master merge because dev contains many in flight changes and may be unstable. I would request you create hotfix PRs (one against master and one against dev), let me know if you need help there.

PR #1951, opened 2018-04-10
[Production Deployment] hotfix: v5 deployment - changes in tc-notifications endpoints.

sachin-maheshwari:


vikasrohit:
Can we create a another PR against dev for the same changes? And merge that in dev and verify that it works in dev env.

vikasrohit:
I think we are okay @sachin-maheshwari. @RishiRajSahu reminded me that these changes are already in dev and seems to work fine.

Issue #2036, opened 2018-06-13
Phase card header background color is not correct


vic-topcoder:
Change the color to match the design specification, it is light gray. Right now the card is white.

![screen shot 2018-06-13 at 9 47 13 am](https://user-images.githubusercontent.com/8484061/41365670-df94d000-6eee-11e8-9403-a4fba85b7944.png)

There are also visual artifacts in the styling that need to be cleared - rounded corners, missing drop shadows, lines.

Current implementation:
![screen shot 2018-06-13 at 9 48 39 am](https://user-images.githubusercontent.com/8484061/41365897-7f6d0a02-6eef-11e8-8736-3470a0afe966.png)

Expected (design):
![screen shot 2018-06-13 at 9 49 55 am](https://user-images.githubusercontent.com/8484061/41365905-86b25f1a-6eef-11e8-9a03-9359117fb452.png)

gondzo:
looks good. closing

PR #2067, opened 2018-06-20
correct phase card header background and other styles #2036

sumitdaga:
@gondzo  @RishiRajSahu


RishiRajSahu:
LGTM. @gondzo please review from your side too.

Issue #2059, opened 2018-06-18
Verify if very old discussions tab is visible after project plan changes


vikasrohit:
We used to have a separate tab for non customer facing discussions which was discarded later on, however, few old projects are still using that tab for communication between managers and copilots. We need to validate that if that tab is still available after refactoring done for project plan.

fyi @RishiRajSahu

vikasrohit:
Verified and it is working however, style for the edit icon is off.

https://connect.topcoder-dev.com/projects/6404/discussions can be used for testing in dev.

<img width="1280" alt="screen shot 2018-06-18 at 3 30 08 pm" src="https://user-images.githubusercontent.com/2417632/41530075-9cfdff56-730c-11e8-8ea6-2600836b9fb1.png">

sumitdaga:
@vikasrohit  @RishiRajSahu 
![discussions tab test 18-06-2018 15_26 - topcoder - google chrome 2018-06-20 15 19 00](https://user-images.githubusercontent.com/7769148/41651151-7d573a40-749d-11e8-9e48-897e702d4e05.png)

this is how it looks to me and the dropdown menu position also seems wrong to me when the icon is clicked

sumitdaga:
i am hoping my PR fixes this issue without causing any side effect

RishiRajSahu:
@sumitdaga replied on PR

RishiRajSahu:
Closing. Pushed to Dev.

PR #2070, opened 2018-06-20
fix edit icon in old discussions tab #2059

sumitdaga:
@gondzo  @RishiRajSahu


RishiRajSahu:
screen shot 2018-06-21 at 1 06 32 pm

@sumitdaga I think this broke the dropdown for Topic

Issue #2039, opened 2018-06-13
Messages with a time gap of 10+ min should show the avatar and time stamp.


vic-topcoder:
### Expected behavior
When a user posts multiple messages and the time difference between them is bigger than 10 minutes, the messages should not be bundled. The messages with a time gap of 10+ min should show the avatar and time stamp.

### Actual behavior
All messages by the same author are bundled regardless of time interval between them

![screen shot 2018-06-13 at 4 23 59 pm](https://user-images.githubusercontent.com/8484061/41383584-4ed8e730-6f26-11e8-917b-ee43ea321536.png)

Expected (mock design):
![screen shot 2018-06-13 at 4 27 45 pm](https://user-images.githubusercontent.com/8484061/41383670-c0a61c34-6f26-11e8-9916-8d7006ba25a8.png)

sumitdaga:
@vikasrohit  @RishiRajSahu @gondzo  
also in mobile no matter the time difference between successive posts by the same user, they are all rendered separately with user avatar and time stamp as of now ...
should this remain as it is on mobile for now? or something else ?

gondzo:
I think we should follow the same logic in mobile. Can you confirm @RishiRajSahu

vikasrohit:
Yes, same logic of grouping messages in 10 minutes should be applicable to both desktop and mobile versions.

sumitdaga:
ok, i will update the PR accordingly ...
just to be sure ...it was not at all grouped on mobile till now (all posts were separate) ...
not sure if it was by mistake or intentional ...
but we want the same behaviour on both mobile and desktop?

RishiRajSahu:
Yes, same behaviour for mobile as well.

gondzo:
looks good. merged to dev. closing

vikasrohit:
@gondzo what made us to reopen this issue?

RishiRajSahu:
was partially committed earlier, now its fixed and merged to dev. Closing.

PR #2071, opened 2018-06-20
messages by same user with time gap of 10min + show separately with a…

sumitdaga:
…vatar and time stamp #2039 

@gondzo  @RishiRajSahu


RishiRajSahu:
@sumitdaga Test all the cases for bundling

Issue #2034, opened 2018-06-13
New post card doesn't display all formatting correctly.


vic-topcoder:
### Expected behavior
Post cards should display the formatting correctly, supporting all the options we enabled with the new composer (old composer with extended formatting)

### Actual behavior
Only Underline works, check the post with formatting on https://connect.topcoder-dev.com/projects/6398
![screen shot 2018-06-13 at 9 42 38 am](https://user-images.githubusercontent.com/8484061/41365522-73e5ba22-6eee-11e8-98a8-f1e53c0b48eb.png)

gondzo:
Here are the styles we had previously https://github.com/appirio-tech/connect-app/blob/b31dd8a8d3fc1b2ec74c61cc053c4dbfc266663b/src/components/Feed/draftjs.scss#L86

can you check why are they not applied anymore and bring them back if necessary

sumitdaga:
@gondzo  
also it would be helpful if you could may be point me to a commit when this was working

gondzo:
It was working at commit b31dd8a8d3fc1b2ec74c61cc053c4dbfc266663b

RishiRajSahu:
@rmjmccormick working fine in dev now. closing.

PR #2074, opened 2018-06-21
fixes post card formatting, both desktop and mobile #2034

sumitdaga:
@gondzo @RishiRajSahu  
i think this should do it ...for both desktop and mobile


RishiRajSahu:
LGTM.

Issue #1731, opened 2018-03-01
[$60] Creating a new project from continuation does not redirect to Dashboard.


vic-topcoder:
_From @macs054 on February 18, 2018 12:17_

### Reported URL ###
https://connect.topcoder-dev.com/new-project/app

### Reproduction Steps ###
Go to https://connect.topcoder-dev.com/
Login as pshah_manager/topcoder123
Open new project window (Create new project app)
Select App
Type anything on the Project Name
Close the new project window
Open new project window
Click 'Continue where I left off'
Complete the required fields
Click Continue
Message will appear that project is created

### Expected Result ###

After the project is created the user must be redirected to the new project's dashboard.

### Actual  Result ###

The page stays put after the project is created which is not the usual behavior when creating a new one.

### Screencast ###

https://youtu.be/clhuUpg3tws

### Issue Information ###

Frequency: Always

Device: Desktop

OS: Windows 10

Browser: Chrome 64.0.3282.140, Edge 38.14393.0.0, Firefox 58.0.2

_Copied from original issue: topcoderinc/connect-test-cycle-1#217_

Huiqi:
I would like to work on it. But I can't assign it to myself. Any one can give me the permission please?

Huiqi:
But I found this bug can not be reproduced. Maybe it has been fixed?

Saoskia:
Definitely not fixed, I just reproduced it myself.

narekcat:
I couldn't find solution, something mysterious.

PR #2136, opened 2018-07-05
fix #1993 Seems like project is not being removed from the local storage …

### LINKED ISSUE #1993: [$60] Seems like project is not being removed from the local storage after creation
### Expected behavior
Ideally, it should remove the project from the local storage once the new project is created.

### Actual behavior
Even after successful creation of a project, we can see the same project still present in the local storage.

- Create a new project
- Click on new project button immediately after successful creation of the first project.
- Notice that you are represented with incomplete project screen which signifies that we still have the project data in local storage.

--

#### Environment
- OS: Any
- Browser (w/version): Any
- User role (client, copilot or manager): Any
- Account used: N/A

zjuasmn:
…after creation

Related to #1731.
When user navigate from project detail to create new project page, these two page refer to the same project object in redux store, So the current condition would not consider it successfully created.

The delay execution of `history.push` is to let user see notification after they creating project.

### DIFF
--- a/src/projects/create/containers/CreateContainer.jsx
+++ b/src/projects/create/containers/CreateContainer.jsx
@@ -63,7 +63,8 @@ class CreateContainer extends React.Component {
   constructor(props) {
     super(props)
     this.state = {
-      creatingProject : false,
+      creatingProject: false,
+      createdProject: false,
       isProjectDirty: false,
       wizardStep: 0,
       updatedProject: {}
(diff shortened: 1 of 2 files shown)


zjuasmn:
also fix #2151

gondzo:
@zjuasmn when clicking the create project button, this will show the loading indicator, then briefly show the new project wizard again and then finally the project details page (loading indicator and then the project page). 
What we need is just loading indicator and target page.

zjuasmn:
@gondzo done.

gondzo:
@zjuasmn looks better. there is some issue with alerts being shown without styling in between the transition, see https://www.screencast.com/t/bf65IcdC

Also, let's remove 
```
if (this.state.createdProject) {
+      return null
+    }
```
and set 
```
<EnhancedCreateView ...
processing={ this.state.creatingProject || this.state.createdProject}
```

zjuasmn:
@gondzo I try trigger `Alert.success` in create view and found that the style is same with that in your video. The alert box style is customized in <https://github.com/appirio-tech/connect-app/blob/dev/src/styles/vendors/s-alert-default.scss>.

Shall I change the style to base on default style in `react-s-alert`?

gondzo:
good catch. can you investigate & fix alerts as well?

zjuasmn:
@gondzo Alert works now.
image

PR #2140, opened 2018-07-05
fix #1954 Search results are not visible on any status filter other...

### LINKED ISSUE #1954: [$30] Search results are not visible on any status filter other than 'All Projects'
### Expected behaviour
Should show search results in all project status filters.

### Actual behavior
It does not apply the text filter with any specific project status filter except 'All Projects'.

- Login into Connect as manager.
- Make sure you have a project status filter, other than 'All Projects', selected.
- Now do a text search from the top bar.
- Notice there is no effect of search on the projects shown in the grid. Now if you select 'All Projects' filter, it would show you the search results.

#### Environment
- OS: Any
- Browser (w/version): Any
- User role (client, copilot or manager): manager
- Account used: vikasrohit

fyi @vic-topcoder @jpoarch @gondzo I think we should fix the text message in case of no results found something like 'No results found with current search criteria, please try relaxing the filters' because without such information user might be confused when he/she searches from top bar while being on particular status filter and there is no result for that particular status but there are results for different project status. User might miss to click on appropriate status filter unless we notify them with more detailed message.

jankyzhang:
screencapture-localhost-3000-projects-2018-07-05-16_24_11

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -341,10 +341,21 @@
         "react-textarea-autosize": "^5.2.1",
         "react-transition-group": "^2.2.1",
         "redux-thunk": "^2.1.0",
-        "tc-ui": "git+https://github.com/appirio-tech/tc-ui.git#8dcc15b5c2aca91cd8deecb523523dae2856e6ef",
         "uncontrollable": "^4.0.1"
       },
       "dependencies": {
+        "fbjs": {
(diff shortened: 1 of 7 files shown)


jankyzhang:
screencapture-localhost-3000-projects-2018-07-05-23_03_12

Issue #2126, opened 2018-07-03
Project listing page is rendered in project creation process


vikasrohit:
### Expected behavior
Project listing page should not be visible

### Actual behavior
Project listing page is visible for some time before redirecting to the newly created project.

### Steps to reproduce the problem
- Create a new project
- Observe the page after hitting the `Continue` button the creation form

#### Environment
- OS: Any
- Browser (w/version): Any
- User role (client, copilot or manager): Any
- Account used:  Any

gondzo:
it's not fixed yet. also, if the project isn't created (api error), we close the project wizard and move to project listing page - should we fix this as well?

vikasrohit:
@gondzo is it moving to listing page or the page from where user started the project creation wizard? I think we are okay if we taking user back to the page from where user started the project creation. @rmjmccormick could you please suggest a better way to handle the error use case?

andela-jomadoye:
I can do this, I'll work on it.

vikasrohit:
@gondzo I know we haven't merged the PR for this issue yet, but we have started seeing login page instead of project listing page now when user clicks on a project from the listing page.

gondzo:
@vikasrohit let me know if you still see the login page when opening project details

vikasrohit:
Thanks @gondzo. It seems to be working on dev now. 👍

PR #2144, opened 2018-07-05
#2126 Stop project listing page from displaying when redirecting

andela-jomadoye:
The PR stops project listing page from displaying when redirecting after creating a project #2126


gondzo:
this is already fixed in dev

PR #2172, opened 2018-07-09
fix #2038 Do not display error until the user has completed the URL a…

### LINKED ISSUE #2038: [$40] Do not display error until the user has completed the URL address when adding a link
### Expected behavior
I would like to see an error only after I completed entering my  URL address in add link section. While I'm typing, I don't want to see an error. Once I'm done typing, if the link is valid, the button should become blue. If I lose focus and there is error, the error message should appear. Let's be smart about the user experience

### Actual behavior
Error is displayed as soon as I start typing in the field. That's bad UX
screen shot 2018-06-13 at 9 57 25 am

zjuasmn:
…ddress when adding a link

### DIFF
--- a/src/components/MobileExpandable/MobileExpandable.scss
+++ b/src/components/MobileExpandable/MobileExpandable.scss
@@ -40,7 +40,7 @@
   }
 
   :global {
-    input[name="address"] + .error-message {
+    input[name="address"]:focus + .error-message {
       display: none;
     }
   }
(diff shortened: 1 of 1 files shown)


gondzo:
@zjuasmn this does not fix the issue - error message should not be displayed while typing - once the field loses focus the error message would be displayed

zjuasmn:
@gondzo Please check my updated PR

PR #2212, opened 2018-07-19
#2011 fix Bring back the welcome post to the customer upon project cr…

### LINKED ISSUE #2011: [$40] Bring back the welcome post to the customer upon project creation
When the user lands on the dashboard after their new project is created, the system used to create a welcome post from Coder.

It is really important to maintain the welcome feel of the user experience, thus let's reinstate the initial post that drives the user to go and complete their specification, and explains what they should expect next.

jankyzhang:
fix #2011

### DIFF
--- a/src/api/messages.js
+++ b/src/api/messages.js
@@ -121,7 +121,7 @@ export function getTopicsWithComments(reference, referenceId, tag, removeCoderBo
       if (removeCoderBotTopics) {
         //remove coderBot posts
         const rTopics = _.remove(topics, i =>
-          [DISCOURSE_BOT_USERID, CODER_BOT_USERID, TC_SYSTEM_USERID].indexOf(i.userId) > -1
+          i.title.indexOf('Your project has been created, and we') === -1 && [DISCOURSE_BOT_USERID, CODER_BOT_USERID, TC_SYSTEM_USERID].indexOf(i.userId) > -1
         )
         totalCount -= rTopics.length
       }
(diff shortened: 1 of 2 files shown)


src/api/messages.js
```diff
@@ -121,7 +121,7 @@ export function getTopicsWithComments(reference, referenceId, tag, removeCoderBo
       if (removeCoderBotTopics) {
         //remove coderBot posts
         const rTopics = _.remove(topics, i =>
-          [DISCOURSE_BOT_USERID, CODER_BOT_USERID, TC_SYSTEM_USERID].indexOf(i.userId) > -1
+          i.title.indexOf('Your project has been created, and we') === -1 && [DISCOURSE_BOT_USERID, CODER_BOT_USERID, TC_SYSTEM_USERID].indexOf(i.userId) > -1
```
vikasrohit: We should avoid doing string comparison. Rather, we should not hide the system messages now as we are only posting for project creation now. fyi @gondzo

jankyzhang: @vikasrohit ,@gondzo How to change about this bug ?
Just make it not hide the system messages now ?
Please make a decision .
Thanks

jankyzhang: @vikasrohit , the new commit is about the requirement '**not hide the system messages**'
Please check it.
Thanks

vikasrohit: Thanks for the changes @jankyzhang . Please exclude the package-lock.json file from the PR as it should not have any changes for this fix. fyi @gondzo

jankyzhang: @vikasrohit , thanks for the reply.
Sorry for the package-lock.json file.
I just submit a new commit .
Please check it.
Thanks.

Issue #2227, opened 2018-07-24
[$80] Remove individual calls for creating milestones for a timeline


RishiRajSahu:
Right now we are calling milestone's POST call multiple times to create milestones for a timeline. Now we have backend supported with implicit creation of milestone when calling POST for milestone with `templatedId` param passed, so this single call will create timeline and all the associated milestone using value of `templatedId`

narekcat:
@gondzo @RishiRajSahu I start connect-app locally by running `npm start`, but when I try to create new app, it return HTTP error 500. In development console it says HTTP error 404 when requests to this URL  https://api.topcoder-dev.com/v4/projects/6704/phases by method GET. Are there any known issues with development back end ?

gondzo:
Which account are you testing with?
available accounts are posted in the challenge forums

narekcat:
@gondzo Ok, I have started to work on this issue before you posted available accounts in the form and create my own (username: testcat, email: test-cat@mail.ru). But ok no problem I'll try with accounts posted in the forum. Thank you.

narekcat:
@gondzo @RishiRajSahu The description is ambiguous, I can't understand. How many endpoints we have in back end now ? One for creating timeline and one for creating milestones ? Or only one endpoint for creating timeline and all milestones together ?

narekcat:
@gondzo @RishiRajSahu The description is ambiguous, I can't understand. How many endpoints we have in back end now ? One for creating timeline and one for creating milestones ? Or only one endpoint for creating timeline and all milestones together ? And if endpoint is only one, which one is /timelines or /milestones ?

gondzo:
@narekcat create timeline endpoint will accept templateId parameter and will create all the necessary milestones from the templates in the backend. You just need to pass the templateId (productTemplateId) and and remove calls to create individual milestones. 
Here is the relevant backend change - https://github.com/topcoder-platform/tc-project-service/commit/827b2252947eb2e3a18f5732ebc09dc7d3187963#diff-7a3be78f0970e178e8ace2b5c70b17baR43

narekcat:
@gondzo Yes I already understand it from repository https://github.com/topcoder-platform/tc-project-service/tree/feature/timeline-milestone, but there are some issue, because I did as you say, but milestones didn't created.

gondzo:
Can you post api call details here (url, body) so we can investigate further

narekcat:
@gondzo Api call details are:
[...]

PR #2238, opened 2018-07-25
Issue #2227: Remove individual calls for creating milestones for a timeline

narekcat:
**Description** #2227 
Right now we are calling milestone's POST call multiple times to create milestones for a timeline. Now we have backend supported with implicit creation of milestone when calling POST for milestone with templatedId param passed, so this single call will create timeline and all the associated milestone using value of templatedId.


narekcat:
@RishiRajSahu Ok, sorry, I couldn't understand what you mean at first. But way update comments to the method updated ? They says exactly what method does (Create timeline and milestones for a product).

PR #2239, opened 2018-07-25
issue_2187 ffixed update end date when marked as delivered

Gauravseta:
End date will be updated to the date when it is marked as delivered.


gondzo:
@Gauravseta this will update the date any time the phase is updated in the delivered state. it should be updating only when previous state wasn't delivered

Gauravseta:
@gondzo to my assumption..once phase reaches delivered state, it is no more possible to edit the phase , hence no way of changing the phase once it is in delivered state. So i checked if phase is going to be in delivered state then change end date. correct me if my assumption wrong

gondzo:
ah, didn't notice that. good catch.

vikasrohit:
@gondzo I think we should still apply the logic only when previous state was not delivered. There might be cases when admin would be updating some field in phase even after phase is marked as delivered, and that should not cause delivered date to be updated.

PR #2252, opened 2018-07-30
Issue_2235:Confirmation modal before accepting final design derlivera…

meggamind:
…bles


gondzo:
no need for this change, already working in dev

Issue #2352, opened 2018-08-13
After adding a new phase, don't call loading timeline again as we get it during creation process


maxceem:
We don't have to make this dispatch https://github.com/appirio-tech/connect-app/blob/dev/src/projects/actions/project.js#L287
As this line already returns timeline https://github.com/appirio-tech/connect-app/blob/dev/src/projects/actions/project.js#L321

same action `CREATE_PROJECT_STAGE` could be caught in two reducers

Aha! Link: https://topcoder.aha.io/features/TCCONNECT-329

PR #2353, opened 2018-08-13
issue #2352 - After adding a new phase, don't call loading timeline again as we get it during creation process

maxceem:
issue #2352 - After adding a new phase, don't call loading timeline again as we get it during creation process


RishiRajSahu:
closing as the change is already pushed to dev.

Issue #2365, opened 2018-08-15
Menu popup for a post is not completely visible


vikasrohit:
<img width="788" alt="screen shot 2018-08-15 at 11 59 21 am" src="https://user-images.githubusercontent.com/2417632/44134477-c924c31e-a082-11e8-80b1-356df139dc19.png">

@maxceem Can you please get it resolved quickly (EOD today)?

maxceem:
@vikasrohit yes, will fix it asap

maxceem:
Fixed by PR https://github.com/appirio-tech/connect-app/pull/2367, keeping in mind this one https://github.com/appirio-tech/connect-app/issues/2310

See video http://take.ms/mDdYn

vikasrohit:
Thanks @maxceem for keeping #2310 in mind, otherwise it could have created a regression of a high priority issue.

vikasrohit:
Working fine. Thanks.

PR #2367, opened 2018-08-15
issue #2365 - Menu popup for a post is not completely visible

maxceem:
issue #2365 - Menu popup for a post is not completely visible


vikasrohit:
LGTM. I was bit unsure by the name of the class where we have applied `overflow-x:auto`, however, as it is working fine in the demo video, assuming that is the correct place to fix.

Issue #2366, opened 2018-08-15
Root post should not have delete post option


vikasrohit:
@maxceem please hide(ideally it should be disabled with appropriate tooltip, but as tooltip in side a dropdown is not working right now, opting for hide option) the delete button for first/root post of the thread. Right now it is hiding the button only when there is only single post in the thread.

<img width="789" alt="screen shot 2018-08-15 at 12 01 38 pm" src="https://user-images.githubusercontent.com/2417632/44134628-50cebd6a-a083-11e8-88d2-8bfb811b3962.png">

maxceem:
Done, so Delete post is not shown for root post.

I can make it look like disabled instead, but when user will click it, the dropdown will be closed anyway even it look disabled. I cannot prevent closing in current implementation as it uses `<DropdownItem>` component from `appirio-tech-react-components` which doesn't let us prevent closing `<Dropdown>` when we click it.
Also, styling of the disabled option would be done not the best and slightly hacky way because of the limitations of `<DropdownItem>` component from `appirio-tech-react-components`.

Please, let me know if you still want to keep it with disabled styles, even though when user will click it dropdown will be closed (delete action wouldn't be called, just closed dropdown).

vikasrohit:
> Done, so Delete post is not shown for root post.

That is perfect for now. We won't make it look disabled as of now because we can't show tooltip for that disabled item at present.

vikasrohit:
Working fine now. Thanks.

vikasrohit:
@maxceem in dev I am seeing strange behaviour for 'Delete Thread' option though. I think that is not caused by this commit, and it is rather happening because of missing data in member service in dev env. If you can devote 5 minutes to debug that, that would be great.

maxceem:
@vikasrohit I've tried to delete a thread at it did work well for me. Could you provide any details about the strange behavior you experience?

vikasrohit:
I would update the thread once see it again.

PR #2368, opened 2018-08-15
issue #2366 - Root post should not have delete post option

maxceem:
issue #2366 - Root post should not have delete post option

Done, so Delete post is not shown for root post.

I can make it look like disabled instead, but when user will click it, the dropdown will be closed anyway even it look disabled. I cannot prevent closing in current implementation as it uses `<DropdownItem>` component from `appirio-tech-react-components` which doesn't let us prevent closing `<Dropdown>` when we click it.
Also, styling of the disabled option would be done not the best and slightly hacky way because of the limitations of `<DropdownItem>` component from `appirio-tech-react-components`.

Please, let me know if you still want to keep it with disabled styles, even though when user will click it dropdown will be closed (delete action wouldn't be called, just closed dropdown).

### DIFF
--- a/src/components/Feed/FeedComments.jsx
+++ b/src/components/Feed/FeedComments.jsx
@@ -301,7 +301,7 @@ class FeedComments extends React.Component {
           hasError={item.error}
           allMembers={allMembers}
           noInfo={item.noInfo}
-          canDelete={comments.length>1}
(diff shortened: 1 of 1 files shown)


vikasrohit:
LGTM 👍

PR #2436, opened 2018-08-25
Issue 2305 fixed

Gauravseta:


gondzo:
@Gauravseta this will actually just move the bar down by two pixels, so it's not centered. also, the slider bar now overflows the container and conflicts with the bellow text
image

image

Gauravseta:
@gondzo will look into it immediately.

Gauravseta:
@gondzo Pr updated. screenshot:
issue_2305

Issue #2139, opened 2018-07-05
[100$] Support categories for product templates


vikasrohit:
As of now we are not supporting categories for the product templates. To achieve following UI for product catalog, we need to support the categories for the product templates.

<img width="908" alt="screen shot 2018-07-05 at 12 06 15 pm" src="https://user-images.githubusercontent.com/2417632/42306229-de98c3a0-804b-11e8-8963-960fa26826bb.png">

@rmjmccormick @vic-topcoder Should we use exact same categories as we are using for project templates or do you think there can a case where there is new category or a missing category for product template?

Aha! Link: https://topcoder.aha.io/features/TCCONNECT-395

rmjmccormick:
@vikasrohit @vic-topcoder To be clear, based on version in prod, current project template categories are App, Website, Chatbot, Design, Software Development and QA. Product (do you mean phase?) template categories should presumably follow the designations in the new Workbook, namely Design, Development, Quality Assurance and Data Science.

vikasrohit:
> Product (do you mean phase?)
Essentially yes as right now we are supporting single product per phase. However, our backend supports multiple products per phase but that is a topic for a different day.
From the possible values of categories of products, I think we need to keep two separate list of categories, one for project templates and another for product templates.

vic-topcoder:
Project templates categories ≠ phase categories. 

- **Project templates** are purely a project plan structure and can be organized in categories by the desired outcome.
- **Phases** are containers of work and we're not worried about products at that level.
- **Products** in essence are standardized challenge formats, and they are different from Phases. Products are used for delivery - we will have to talk more about delivery plans. A phase can have different delivery plan, but stays the same (timeline, milestones, price, duration, managed hours).
- **Addons** are phases that are added as a secondary extension to the project plan, they're not products.

This means we should support `category` for project templates, phases, and products.

vikasrohit:
Yes @gondzo. Every product template now has a `category` associated, right now every product in the dev env has `generic` as category value. I would update couple of them to reflect the correct categories so that changes for this issue can be tested.

nahidshahin:
@gondzo [PR](https://github.com/appirio-tech/connect-app/pull/2441) issued

PR #2441, opened 2018-08-27
issues-2139: Support categories

nahidshahin:
Category names at design is left aligned but cards are centre aligned. So I keep both centre aligned. See the screenshot:  

issues-2139


nahidshahin:
#2139

vikasrohit:
@nahidshahin Please keep them as per design, i.e. left align both heading and cards.

nahidshahin:
Did the CSS adjustment at a484eaa. The UI should follow the design now.

PR #2442, opened 2018-08-27
Issue 2401

vikasrohit:
Looks good to me.

Merging to test out in dev. fyi @RishiRajSahu


vikasrohit:
@maxceem Seems like there are lint errors.

maxceem:
@vikasrohit that's very strange. These errors happen either at my local machine or if fixed, happen on CI. probably some version mismatches, I'll check.

vikasrohit:
@maxceem the difference between local and CI is that CI might be caching the old version of a dependency because of it `package-lock.json` based caching unless this file is updated. OR there might be case where the cache is bursted on CI because we have a change in `package-lock.json` file but it is not bursted at our local.

maxceem:
Fixed.

vikasrohit:
Thanks @maxceem, Merged in dev.

Issue #2488, opened 2018-09-10
[$40] Consider Planned phases for calculating Project progress


RishiRajSahu:
Current implementation of calculation project progress doesn't consider planned phases its only considering `Active` and `delivered`. It should consider planned phases as well. Please find the screen shot for existing behaviour.

![screen shot 2018-09-10 at 2 31 37 pm](https://user-images.githubusercontent.com/11248478/45287702-8e803c00-b506-11e8-94be-c0cea4b01d8d.png)

**Current Behaviour -** 

Project Progress = (100 + 100 + 100 + 82)/4 = 96

**Expected Behaviour -** 

Project Progress = (100 + 100 + 100 + 82 + **0**)/**5** = 76

RishiRajSahu:
fyi @maxceem

iversonLv:
@maxceem I would like to work on it

iversonLv:
It’s done and ready for review

PR #2491, opened 2018-09-10
Issue 2488

iversonLv:
issue #2488

### DIFF
--- a/src/helpers/projectHelper.js
+++ b/src/helpers/projectHelper.js
@@ -161,7 +161,7 @@ export function formatProjectProgressProps(project, phases, productsTimelines) {
   const spentAmount = _.sumBy(activeAndCompletedPhases, 'spentBudget') || 0
   const labelSpent = spentAmount > 0 ? `Spent $${formatNumberWithCommas(spentAmount)}` : ''
   const progressPercent = phases.length > 0 ? Math.round(totalProgress/activeAndCompletedPhases.length) : 0
-  const labelStatus = `${progressPercent}% done ${totalProgress} ${activeAndCompletedPhases.length}`
+  const labelStatus = `${progressPercent}% done`
 
   return {
     labelDayStatus,
--- a/src/helpers/projectHelper.js
+++ b/src/helpers/projectHelper.js
@@ -6,6 +6,7 @@ import {
   PROJECT_ROLE_OWNER,
   PHASE_STATUS_ACTIVE,
   PHASE_STATUS_COMPLETED,
+  PHASE_STATUS_REVIEWED,
   PHASE_STATUS_DRAFT,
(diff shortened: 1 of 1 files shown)


src/helpers/projectHelper.js
```diff
@@ -144,7 +145,7 @@ export function formatProjectProgressProps(project, phases, productsTimelines) {
 
     // override project progress if status is delivered
     if (phase.status === PHASE_STATUS_COMPLETED) {
-      progress = 100
+      progress = 99
```
maxceem: @iversonLv Could you please provide some background why there should be `99` instead of `100` as it feels there should be some issue with the logic if need to put `99` here.

src/helpers/projectHelper.js
```diff
@@ -87,7 +88,7 @@ export function formatProjectProgressProps(project, phases, productsTimelines) {
 
   // phases where start date is set and are not draft
   const nonDraftPhases = _.filter(phases, (phase) => (phase.startDate && phase.status !== PHASE_STATUS_DRAFT))
-  const activeAndCompletedPhases = _.filter(phases, (phase) => (phase.startDate && (phase.status === PHASE_STATUS_ACTIVE || phase.status === PHASE_STATUS_COMPLETED)))
+  const activeAndCompletedPhases = _.filter(phases, (phase) => (phase.startDate && (phase.status === PHASE_STATUS_ACTIVE || phase.status === PHASE_STATUS_COMPLETED || phase.status === PHASE_STATUS_REVIEWED)))
```
maxceem: @iversonLv As the meaning of the variable `activeAndCompletedPhases` has changed after adding new phase status, could you please consider renaming it to reflect the new meaning.

src/helpers/projectHelper.js
iversonLv: i just found if it is 100(deliver) final total will be > actual total 1， if there is no deliver phase there is correct. I do/can not test if all phases are deliver. 

So why I decrease 1 for deliver phase from 100 to 99.

src/helpers/projectHelper.js
iversonLv: Good suggestion

src/helpers/projectHelper.js
maxceem: Thanks. Regarding the final total > 1. I think if this happens there has to be some issue with logic so I would prefer to create a separate issue to fix it. If I understood it correctly such behavior is not caused by the changes in this PR so you can just leave `100` here for now.

iversonLv:
@maxceem I invite you to my fork repository, you might need to accept?

maxceem:
@iversonLv it looks like I don't need access to the fork. I'll merge in a moment.

Issue #2480, opened 2018-09-07
[$35]  Align user avatar in the post composer correctly


RishiRajSahu:
User avatar in the post composer (footer) is not aligned with the avatar from the actual post.

https://github.com/appirio-tech/connect-app/issues/2301

maxceem:
@RishiRajSahu There is one more thing. When user doesn't have and image avatar defined. It's not displayed next to composer at all, see for example as `pshah_manager`:

![image](https://user-images.githubusercontent.com/146016/45215152-28f63a80-b2cf-11e8-9801-7f1fa2a40fc8.png)

I don't remember what was the reason to remove avatar in such case. But probably it's not desired, and we have to put it back even if user doesn't have an image avatar. So intials has to be displayed instead of the image as at other places.

gurmeetb:
@maxceem Is this open for pickup ?

maxceem:
@gurmeetb not yet, but will be open soon after the relative challenge run.

gurmeetb:
Thanks @maxceem for quick response.

maxceem:
@gurmeetb just want to let you know that challenge is planned to be started on Monday. The Bug Bash process will be slightly changed.

maxceem:
@gurmeetb  The challenge is on.
Important! There is a change in the Bug Bash process, please read the new rules in the challenge spec carefully, before starting work on issue https://www.topcoder.com/challenges/30070926/?type=develop&noncache=true
Thank you!

arun1595:
@maxceem I would like to tackle this issue.

gurmeetb:
@maxceem Can I work on this now ?

gurmeetb:
Thanks

gurmeetb:
@maxceem I am new to this app, could you guide how do I land the page to see the issue ?

iversonLv:
@maxceem I think I could get the ticket as per Bug bash rule: 12hr is gone and there is no PR

gurmeetb:
It’s done and ready for review

gurmeetb:
@iversonLv when it was assigned I was sleeping - logically I started working only 3 hrs back and there is PR moved in. 

@maxceem @vikasrohit  12 hrs rule will not work. You need to take into consideration ppl from around the globe take part so we should have 24 hrs rule.

maxceem:
@gurmeetb got your PR, thanks.

[...]

PR #2493, opened 2018-09-10
#2480 - Align user avatar in the post composer correctly

gurmeetb:
Fix for #2480

### DIFF
--- a/src/components/RichTextArea/RichTextArea.jsx
+++ b/src/components/RichTextArea/RichTextArea.jsx
@@ -303,7 +303,7 @@ class RichTextArea extends React.Component {
         <div className="modal-row">
           {avatarUrl &&
             <div className="portrait">
-              <Avatar avatarUrl={getAvatarResized(avatarUrl, 30)} userName={authorName} />
+              <Avatar size={40} avatarUrl={getAvatarResized(avatarUrl, 30)} userName={authorName} />
             </div>
           }
           <div className={cn('object', {comment: disableTitle}, 'commentEdit')}>
--- a/src/components/RichTextArea/RichTextArea.jsx
+++ b/src/components/RichTextArea/RichTextArea.jsx
@@ -303,7 +303,7 @@ class RichTextArea extends React.Component {
         <div className="modal-row">
           {avatarUrl &&
             <div className="portrait">
-              <Avatar size={40} avatarUrl={getAvatarResized(avatarUrl, 30)} userName={authorName} />
+              <Avatar size={40} avatarUrl={getAvatarResized(avatarUrl, 40)} userName={authorName} />
             </div>
           }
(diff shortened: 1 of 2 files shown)


gurmeetb:
@maxceem i see we have component (Avatar) coming from other repo - `appirio-tech-react-components` - I don't have access to the same. Could you guide me how do I get handle over it ?

src/components/RichTextArea/RichTextArea.jsx
```diff
@@ -303,7 +303,7 @@ class RichTextArea extends React.Component {
         <div className="modal-row">
           {avatarUrl &&
             <div className="portrait">
-              <Avatar avatarUrl={getAvatarResized(avatarUrl, 30)} userName={authorName} />
+              <Avatar size={40} avatarUrl={getAvatarResized(avatarUrl, 30)} userName={authorName} />
```
maxceem: It's almost done. But the image by itself is still 30x30 so it still smaller than avatars of comments.  Most likely function `getAvatarResized` has to be called with the second param as `40` to generate a URL for image 40x40.

gurmeetb:
@maxceem Kindly validate☝️

maxceem:
@gurmeetb the image now comes in a proper size, but this change is not enough to make image same size, see

image

Please, make all neccesary changes so that avatar is displayed in a proper size here.

gurmeetb:
is it fine if we show tooltip too? else we might need to replicate bit of logic from UserToolTip here - i did try out with it and it comes fine - only issue being it comes with tooltip

maxceem:
I think we don't need a tooltip here, and we can implement it without using a tooltip. But let me ask first if we can add a tooltip for this case.

@RishiRajSahu I think we shouldn't show a tooltip for the avatar of our own user here. Please, let me know if you think we may need it here.

RishiRajSahu:
@maxceem yeah tooltip for avatar beside post-composer is not needed.

maxceem:
@gurmeetb 

> will need to tweak the logic a bit.

I'm not sure if it really needs to tweak the logic here. As I see it adjusting or adding some CSS styles has to be enough. Maybe adding some className to some component if really necessary.

Do you have some difficulties to adjust the size only by CSS styles? Let me know if you do, maybe I could help with some ideas.

gurmeetb:
@RishiRajSahu thanks for confirming.

@maxceem let me get back on this in few hrs, will let know if i get stuck. I will look into this in 1-2 hrs, moved out for some work - away from computer right now..let me get back asap

maxceem:
@gurmeetb I cannot find any user I know without an avatar, but you can log out and after register a new user using your e-mail. So it will be a new user on the DEV environment.

gurmeetb:
Thanks @maxceem for Firefox Inspector help. It was quick to find where I was going wrong. It is fixed and pushed in now. 

Thanks :)

PR #2497, opened 2018-09-11
Fix #2489 Hide the the Delete phase warning when marking the phase In Progress

### LINKED ISSUE #2489: Don't show two alerts on phase card when activating a phase
Hide the the Delete phase warning when marking the phase `In Progress` . 

screen shot 2018-09-07 at 3 01 01 pm

amitpatra:


src/projects/detail/components/PhaseCard/EditStageForm.jsx
```diff
@@ -372,17 +373,28 @@ class EditStageForm extends React.Component {
...
+        )}
       </div>
     )
   }
 }
 
 EditStageForm.defaultProps = {
-  cancel: () => {}
+  cancel: () => {},
+  deleteProjectPhase: () => {}
```
maxceem: Defining the default value for `deleteProjectPhase` as `() => {}` is not a good practice here as it would hide the warning if it's not defined, but actually, it's required by the component.
I know it's done this way for `cancel: () => {}` which I think is not good also but better not to create more places like this.

As `deleteProjectPhase` is actually needed for the `EditStageForm` work and it's marked as required in the upper component https://github.com/appirio-tech/connect-app/blob/dev/src/projects/detail/components/ProjectStage.jsx#L280

Could you please remove default value for `deleteProjectPhase` and mark `deleteProjectPhase` as required instead.
Thanks.

amitpatra: Will fix this. Thanks :)

amitpatra:
Hi, I have modified the pull request. Kindly review. Thanks :)

Issue #2339, opened 2018-08-10
[$75] Add tooltip to the disabled `Delivered` item in the phase status


maxceem:
## Task

Add tooltip to the disabled `Delivered` item in the phase status select, same like disabled duration has:

![image](https://user-images.githubusercontent.com/146016/43939835-0c514ffa-9c9f-11e8-8473-404ef31df7e9.png)

## Note

It appeared that `ToolTip` component doesn't work inside `Dropdown` component, probably it has to be fixed inside `appirio-tech-react-components` first.

@vikasrohit
Continues https://github.com/appirio-tech/connect-app/issues/2248#issuecomment-411692163

Aha! Link: https://topcoder.aha.io/features/TCCONNECT-319

vikasrohit:
Moved to next release.

jomadoye:
I am working on this, please assign me @maxceem

Gauravseta:
@maxceem on how many issues will jomaday work? multiple allowed?

jomadoye:
@Gauravseta I am done working on my last issues, I just need to raise PR from my fork as maxceem has informed me, I think that is a new rule, I am already working on this. thanks.

jomadoye:
PR has been raise @maxceem, Please assign this ticket to me, thanks.

maxceem:
@jomadoye okay, here we kind of on the edge of rules but as jomadoye already finished another issue and even tried to push just was confused that this time doesn't have rights to do that and raised a question about providing his work. I think it would be fair enough this time to let him work on this one. I know it's a little bit harder with new rules to follow the things for me also. 

@jomadoye the issue is yours. But to actually assign you to this issue please accept an invitation to this repo. Thanks.

jomadoye:
I have accepted the invitation to this repo, thanks. @maxceem

jomadoye:
@maxceem I just noticed the `Delivered` is not disabled 
<img width="785" alt="screen shot 2018-09-10 at 12 15 02 am" src="https://user-images.githubusercontent.com/26272732/45276174-a4521a80-b48e-11e8-827e-ced17f9e23ec.png">
Is there a specific user I should login with?

jomadoye:
@maxceem login with copilot and visit https://connect.topcoder-dev.com/projects/6605 the `delivered` should be enabled

maxceem:
[...]

PR #2500, opened 2018-09-11
fix issue #2339

suppermancool:
fix issue #2339


src/projects/detail/components/PhaseCard/EditStageForm.jsx
```diff
@@ -255,6 +255,12 @@ class EditStageForm extends React.Component {
       disabled: hasTimeline && status.value === PHASE_STATUS_COMPLETED
     }))
 
+    for (let i = 0; i < activePhaseStatuses.length; i++) {
+      if (activePhaseStatuses[i].disabled) {
+        activePhaseStatuses[i].toolTipMessage = 'Once activated, phase delivery is controlled by the milestones.'
```
maxceem: This code will add this tooltip to any option which is disabled. But this tooltip only for the Delivery option. For now it's the only one disabled option, but it can be changed in future.

For example, you may consider adding `toolTipMessage` property together with the `disabled` property inside `map` https://github.com/appirio-tech/connect-app/blob/a4693ff966eea685d95e66bd555623a6f613b98a/src/projects/detail/components/PhaseCard/EditStageForm.jsx#L253-L256

suppermancool: @maxceem done in https://github.com/appirio-tech/connect-app/pull/2500#discussion_r217309529

PR #2501, opened 2018-09-11
Refactor redux store

### LINKED ISSUE #2474: Do not make additional calls for project/product templates on project page
After this [issue](https://github.com/appirio-tech/connect-app/issues/2472) is completed we can remove some additional calls for project and product templates on the project page and reuse already loaded data in the Redux Store.

Current Redux Store
```js
{
   projectState: {
      ...
      projectTemlate: {}, // current project's projectTemplate
      productTemplates: [],  // current project's productTemplates
      allProductTemplates: [], // all possbile productTemplates
   },
   templates: {
      projectTemplates: [...]   // all possible projectTemplates
      projectCategories: [...]  // all possible projectTypes
      isLoading: false
   }
}
```
What we can do:
1. Currently, project page loads one project template assosiated with it to `projectState.projectTemplate`making separate API call. As after previous issue is done we already have all `projectTemplates` in the store, we can find needed project template by its id. We have to make sure that templates are already loaded in the store at this point, and if not, then call action to load all templates.
   Note: after this we shouldn't have `projectState.projectTemplate` in the store anymore.

2. On the project page we load `projectState.allProductTemplates` with ALL product templates for `add phase` functionality. Also, we load  SOME product templates used in the project to  `projectState.productTemplates. Instead we should get product templates from `projectState.allProductTemplates` for a project using ids.
    Note: after this we shouldn't have `projectState.productTemplates` in the store anymore.

3. After previous two steps we still have `allProductTemplates` in `projectState` which is not a good place for it. As `projectState` suppose to keep some concrete project data, but `allProductTemplates` doesn’t depend on the concrete project. So it's better to move it to the`templates` section of the store and rename to `productTemplates`.
    Note: after this we shouldn't have `projectState.allProductTemplates` in store anymore, and have `template.productTemplates` instead.

After these steps Redux Store should look like this:
```js
{
   projectState: {
      ...
      // don't keep templates data anymore
   },
   templates: {
      // stay as before
      projectTemplates: [...]   // == /v4/projects/metadata -> projectTemplates
      projectCategories: [...]  // == /v4/projects/metadata -> projectTypes
      isLoading: false 
      // moved from projectState, before it was allProductTemplates
      productTemplates: [...] // == /v4/projects/metadata -> productTemplates
   }
}
```

fyi @RishiRajSahu

arun1595:
Fixes #2474


src/projects/detail/containers/ScopeAndSpecificationContainer.jsx
```diff
@@ -123,12 +124,18 @@ SpecificationContainer.propTypes = {
...
-const mapStateToProps = ({projectState, loadUser}) => {
+const mapStateToProps = ({projectState, loadUser, templates}) => {
+  const { projectTemplates, productTemplates } = templates
+
   return {
     processing: projectState.processing,
     error: projectState.error,
     currentUserId: parseInt(loadUser.user.id),
-    productTemplates: projectState.productTemplates,
+    productTemplates: getProjectProductTemplates(
```
RishiRajSahu: nit : Can we have non-null check for templates object like we did at other places to be consistent and avoid any runtime issues ? one more reason behind that is `getProjectProductTemplates` doesn't have a non-null check for productTemplates.

src/projects/detail/containers/SpecificationContainer.jsx
```diff
@@ -21,9 +22,17 @@ const SpecificationContainer = (props) => {
...
 }
 
-const mapStateToProps = ({ projectState }) => ({
-  productTemplates: projectState.productTemplates,
-})
+const mapStateToProps = ({ projectState, templates }) => {
+  const { projectTemplates, productTemplates } = templates
+
+  return {
+    productTemplates: getProjectProductTemplates(
```
RishiRajSahu: same here !

src/projects/detail/containers/ScopeAndSpecificationContainer.jsx
arun1595: @RishiRajSahu Done.

src/projects/detail/containers/SpecificationContainer.jsx
arun1595: @RishiRajSahu Done.

Issue #2185, opened 2018-07-11
Feed components are being re-rendered on typing


maxceem:
This issue is created as a part of this one https://github.com/appirio-tech/connect-app/issues/2138

Typing is very slow when there are several topic threads. Every time we type a symbol all the threads are rerendered. In the next video, I quickly type the same latter with the same interval but letters appear in batches rather one by one http://take.ms/wA1u2

See for example this project https://connect.topcoder-dev.com/projects/2871 and try to type a new topic title.

Aha! Link: https://topcoder.aha.io/features/TCCONNECT-418

vikasrohit:
@maxceem are you working on it? If you haven't started it yet, please wait on this until we close the 2.4.4 milestone.

maxceem:
@vikasrohit it's on my list, but I'm not working on it as this one is quite heavy, will most likely require to refactor the Feed and/or FeedContainer components.

vikasrohit:
Okay, makes sense. Already moved it to 2.4.5.

vikasrohit:
@maxceem Moving it one more release further.

maxceem:
There are few things to keep in mind:
- We don't wanna use `shouldComponentUpdate` for fixing this as it would be hard to support it in the future. Re-renders has to be fixed in a natural way. Most likely by refactoring `FeedContainer.js` to take care about only one topic rather then the list of topics as done now. As a reference we can use `PhaseFeedHOC.jsx` which implements feeds for pahses (maybe we don't need HOC in this case, just container).
- Make sure that all existent functionality works, in particular:
    - When user started editing post and try to leave the page, confirmation prompt is appearing https://github.com/appirio-tech/connect-app/blob/dev/src/projects/detail/containers/FeedContainer.js#L423.

sumitdaga:
please assign this to me if it is open for pick up!

sumitdaga:
@maxceem  
ok so, so far from my analysis so far what i think may be a good idea is to not propagate "newTitle" all the way down from FeedContainer to RichTextEditor as the RichTextEditor is anyway maintaining its own title state ...this i think will avoid all re-renderings happening due to props change when only the title is being edited and not yet saved?

does it make sense?
may to have to dig a little more deeper and actually implement to see if it works

maxceem:
@sumitdaga Yeah, I was considering before the way you proposed `not propagate "newTitle"` but it would break the work of the prompt which has to be shown when we are trying to leave the page where we started editing some post:
```
[...]

PR #2508, opened 2018-09-12
Issue #2185: Feed components are being re-rendered on typing version 2

KevinLeigh:
#### Description

Please see issue description here -> https://github.com/appirio-tech/connect-app/issues/2185#issuecomment-418233790

#### Changes

- Intelligently use `shouldComponentUpdate`

### DIFF
--- a/src/projects/detail/containers/FeedContainer.js
+++ b/src/projects/detail/containers/FeedContainer.js
@@ -77,6 +77,46 @@ class FeedView extends React.Component {
     this.init(this.props)
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
+    if (this.props.currentUser !== nextProps.currentUser) {
+      return true;
+    }
+    if (this.props.isLoading !== nextProps.isLoading) {
+      return true;
+    }
+    if (this.props.isCreatingFeed !== nextProps.isCreatingFeed) {
+      return true;
+    }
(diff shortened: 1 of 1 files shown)


src/projects/detail/containers/FeedContainer.js
```diff
@@ -81,6 +82,14 @@ class FeedView extends React.Component {
     this.init(nextProps, this.props)
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
+    if(nextState.shouldNotRender) {
+      nextState.shouldNotRender = false
```
maxceem: It's very not good to modify the state of the component directly. Method `this.setState()` has to be always used to modify state, see the official React documentation https://reactjs.org/docs/state-and-lifecycle.html#do-not-modify-state-directly

maxceem:
Closing this PR because it doesn't follow requirements from https://github.com/appirio-tech/connect-app/issues/2185#issuecomment-418233790

src/projects/detail/containers/FeedContainer.js
vikasrohit: Also, is it good to change the state in `shouldComponentUpdate` method? I mean wouldn't it is prone to infinite loop when if by mistake we change the state in a way which again requires the re-render.

Issue #2476, opened 2018-09-07
[$20] Post mention should not be link


RishiRajSahu:
@mention should not lead to the profile page; let's remove the profile link - we don't want customers to go to the community profile page at this time.

continued from https://github.com/appirio-tech/connect-app/issues/2301

maxceem:
@RishiRajSahu in another issue there is a task to make mentions bold https://github.com/appirio-tech/connect-app/issues/2201.

After we make mentions not links in this task, should they stay bold? Other words if they are not links anymore, should they be highlighted by styles in some way or look just like a regular text?

RishiRajSahu:
@maxceem so what we need is - 
1.In #2201 it say mentions needs to be bold, which should not be changed
2.As part of this issue we want bold mentions not to contain a href of user profile.

maxceem:
This is one is open for pick up now.

PrakashDurlabhji:
@maxceem please assign it to me

gondzo:
@maxceem @RishiRajSahu post mention being a link is expected in notifications so we should be careful not to break that feature https://github.com/topcoder-platform/tc-notifications/blob/d2e0c71f56a3744cd11516044bd9f9beec7b92a7/connect/connectNotificationServer.js#L72

maxceem:
Thank you @gondzo for your comment.

@PrakashDurlabhji just to be clear on this we only have to remove links from @menitions inside posts:

![image](https://user-images.githubusercontent.com/146016/45423390-c84b7100-b6c5-11e8-9cba-db44b65b918a.png)

There is `@Parth Customer` is a links now, but shouldn't. You may find this example in the project with id `7298` on the dashboard tab.

@RishiRajSahu fyi

vikasrohit:
Good catch @gondzo. Thanks for saving us from a regression. 👍

PrakashDurlabhji:
sure will implement same.

nahidshahin:
@maxceem Want to work on it

PrakashDurlabhji:
@maxceem releasing it, docs are not proper of markdown ,trying to understand since long. wont be able to do it. open for pickup now.

suppermancool:
@maxceem pls assign to me

RishiRajSahu:
@suppermancool assigned to you !

suppermancool:
@RishiRajSahu @maxceem  ready for review in 
https://github.com/appirio-tech/connect-app/pull/2510/files

[...]

PR #2510, opened 2018-09-13
implement issue #2476

suppermancool:
implement issue #2476


suppermancool:

Issue #2466, opened 2018-09-04
[$30] Complete milestone for customer shouldn't have blue line


maxceem:
### How to reproduce:
- Create project with type **App**
- Complete all milestones except of the milestone with title **Complete**
- View project with customer user and he will see:
   **Actual behaviour**:
   ![image](https://user-images.githubusercontent.com/146016/45009611-b7eb2480-b03b-11e8-934f-bd6654616637.png)
   There shouldn't bee blue line along description of this milestone.

    **Expected**:
    ![image](https://user-images.githubusercontent.com/146016/45009666-0ac4dc00-b03c-11e8-9a0c-3e4a17e8f800.png)

- Note. For manager user this milestone is displayed correctly.
   ![image](https://user-images.githubusercontent.com/146016/45009634-dea95b00-b03b-11e8-9971-06832d0d554b.png)

PrakashDurlabhji:
@maxceem can you please assign it to me thanks

maxceem:
@PrakashDurlabhji  just in case, you don't have to wait to be assigned to the issues and can already work on it as described in the new Bag Bush rules. Thanks.

gurmeetb:
@PrakashDurlabhji are you working on this ? If not, please flag @maxceem to release it to me. 

@maxceem Kindly assign me if this gets free.

PrakashDurlabhji:
@gurmeetb PR raising within 20mins @maxceem actively working on it.

PrakashDurlabhji:
@maxceem done Pr raised

maxceem:
@PrakashDurlabhji There is an issue with the PR, I've logged it here https://github.com/appirio-tech/connect-app/pull/2495#pullrequestreview-154598957.

Also, the way of implementing it now is not desirable. Please reuse the current approach of controlling the blue line by reusing component `<DotIndicator>`.

So there is no need to edit CSS stules. The way to do it as it's already done for other places is the next.
All the changes has to be done inside component `MilestoneTypeDelivery` https://github.com/appirio-tech/connect-app/blob/dev/src/projects/detail/components/timeline/milestones/MilestoneTypeDelivery/MilestoneTypeDelivery.jsx which is responsible for showing this milestone.
[...]

PR #2512, opened 2018-09-14
fix issue #2466

suppermancool:
fix issue #2466


src/projects/detail/components/timeline/milestones/MilestoneTypeDelivery/MilestoneTypeDelivery.jsx
```diff
@@ -329,9 +366,11 @@ class MilestoneTypeDelivery extends React.Component {
           Completed status
          */}
         {isCompleted && (
-          <div>
-            <LinkList links={links} />
-          </div>
+          <DotIndicator>
```
maxceem: Look like no need to wrap here with `DotIndicator` as it causes the additional dot to be shown without line, when the final delivery was provided and milestone is complete.

image

suppermancool: @maxceem  done in latest commit

PR #2520, opened 2018-09-15
issues-2471: Adding field `details.metadata.waitingForCustomer`

nahidshahin:
Addressed all 5 cases.

### DIFF
--- a/src/projects/detail/components/timeline/Milestone/Milestone.jsx
+++ b/src/projects/detail/components/timeline/Milestone/Milestone.jsx
@@ -95,7 +95,7 @@ class Milestone extends React.Component {
     updateMilestone(milestone.id, values)
   }
 
-  updateMilestoneContent(contentProps) {
+  updateMilestoneContent(contentProps, metaDataProps) {
     const { updateMilestone, milestone } = this.props
 
     const updatedMilestone = {
@@ -105,6 +105,10 @@ class Milestone extends React.Component {
           ..._.get(milestone, 'details.content', {}),
           ...contentProps,
         },
+        metadata: {
+          ..._.get(milestone, 'details.metadata', {}),
+          ...metaDataProps
+        }
       }
     }
 
@@ -114,12 +118,28 @@ class Milestone extends React.Component {
   completeMilestone(updatedProps) {
     const { completeMilestone, milestone } = this.props
 
-    completeMilestone(milestone.id, updatedProps)
+    const updatedMilestone = {
+      ...updatedProps,
+      details: {
+        metadata: {
+          waitingForCustomer: false
+        }
+      }
+    }
+    completeMilestone(milestone.id, updatedMilestone)
   }
   completeFinalFixesMilestone(updatedProps) {
     const { completeFinalFixesMilestone, milestone } = this.props
 
-    completeFinalFixesMilestone(milestone.id, updatedProps)
+    const updatedMilestone = {
+      ...updatedProps,
+      details: {
+        metadata: {
+          waitingForCustomer: false
+        }
+      }
+    }
+    completeFinalFixesMilestone(milestone.id, updatedMilestone)
   }
 
   extendMilestone(extendDuration, updatedProps) {
--- a/src/projects/detail/components/timeline/Milestone/Milestone.jsx
+++ b/src/projects/detail/components/timeline/Milestone/Milestone.jsx
@@ -118,28 +118,28 @@ class Milestone extends React.Component {
   completeMilestone(updatedProps) {
     const { completeMilestone, milestone } = this.props
 
-    const updatedMilestone = {
-      ...updatedProps,
-      details: {
-        metadata: {
-          waitingForCustomer: false
-        }
+    const details = _.get(updatedProps, 'details.metadata', {})
+    _.set(updatedProps, 'details.metadata', {
+      ...details,
+      metadata: {
+        ..._.get(details, 'metadata', {}),
+        waitingForCustomer: false
       }
-    }
-    completeMilestone(milestone.id, updatedMilestone)
+    })
+    completeMilestone(milestone.id, updatedProps)
   }
   completeFinalFixesMilestone(updatedProps) {
     const { completeFinalFixesMilestone, milestone } = this.props
 
-    const updatedMilestone = {
-      ...updatedProps,
-      details: {
-        metadata: {
-          waitingForCustomer: false
-        }
+    const details = _.get(updatedProps, 'details.metadata', {})
+    _.set(updatedProps, 'details.metadata', {
+      ...details,
+      metadata: {
+        ..._.get(details, 'metadata', {}),
+        waitingForCustomer: false
       }
-    }
-    completeFinalFixesMilestone(milestone.id, updatedMilestone)
+    })
+    completeFinalFixesMilestone(milestone.id, updatedProps)
   }
 
   extendMilestone(extendDuration, updatedProps) {
--- a/src/projects/detail/components/timeline/Milestone/Milestone.jsx
+++ b/src/projects/detail/components/timeline/Milestone/Milestone.jsx
@@ -115,30 +115,30 @@ class Milestone extends React.Component {
     updateMilestone(milestone.id, updatedMilestone)
   }
 
-  completeMilestone(updatedProps) {
+  completeMilestone(updatedProps = {}) {
     const { completeMilestone, milestone } = this.props
 
-    const details = _.get(updatedProps, 'details.metadata', {})
-    _.set(updatedProps, 'details.metadata', {
+    const details = _.get(updatedProps, 'details', {})
+    updatedProps.details = {
       ...details,
       metadata: {
         ..._.get(details, 'metadata', {}),
         waitingForCustomer: false
       }
-    })
+    }
     completeMilestone(milestone.id, updatedProps)
   }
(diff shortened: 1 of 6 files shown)


src/projects/detail/components/timeline/Milestone/Milestone.jsx
```diff
@@ -114,12 +118,28 @@ class Milestone extends React.Component {
   completeMilestone(updatedProps) {
     const { completeMilestone, milestone } = this.props
 
-    completeMilestone(milestone.id, updatedProps)
+    const updatedMilestone = {
+      ...updatedProps,
+      details: {
```
maxceem: This line `details: {` overwrites the `details` property which is defined in `updatedProps` and thus all the updates passed for `details` in `updatedProps` are not applied.

This breaks some functionality, for example:
- In `Visual Design` phase I've added links in `Adding LInks` milestone
- That list of links has to appear in `Design Review` milestone to choose winners
- But there are no links in `Design Review` because they has to be inside `details`

image

src/projects/detail/components/timeline/Milestone/Milestone.jsx
```diff
@@ -114,12 +118,28 @@ class Milestone extends React.Component {
...
+    }
+    completeMilestone(milestone.id, updatedMilestone)
   }
   completeFinalFixesMilestone(updatedProps) {
     const { completeFinalFixesMilestone, milestone } = this.props
 
-    completeFinalFixesMilestone(milestone.id, updatedProps)
+    const updatedMilestone = {
+      ...updatedProps,
+      details: {
```
maxceem: Here is same as above `details: {` overwrites the details property which is defined in `updatedProps`.

nahidshahin: Fixed at 31c9af1

nahidshahin:
@maxceem the problem fixed at 31c9af1, please check once more.

maxceem:
@nahidshahin the previous issue is fixed so I continue testing.

So far I see that case 2 is not supported.

Here is the screenshot:
image

When milestone type `checkpoint-review` (on the screenshot) or `final-designs` become active, and links were intially provided (see on the screenshot in the Redux Store `details.prevMilestoneType==='add-links'` which means links are provided) we have to set `details.metadata.waitingForCustomer=true`.

Most likely the right place to implement this functionality is inside `completeProductMilestone` action around the next lines https://github.com/appirio-tech/connect-app/blob/dev/src/projects/actions/productsTimelines.js#L174

 I guess that idea would be the next
- Check the the `nextMilstone` type is `checkpoint-review` or `final-designs`
- Check that `prevMilestoneType` for the next milestone is `add-links`
- If both of these conditions are met we set `details.metadata.waitingForCustomer=true`.


Meanwhile I'm checking other cases...

maxceem:
@nahidshahin 
Case 5 is not supported for the case in point 1. 
So when we request an extension `details.metadata.waitingForCustomer` becomes `true` which is correct.
But when someone either `approve` or `declined` extension `details.metadata.waitingForCustomer` has to become `false` but for now it's not implemented.

Most likely changes has to be done in 
https://github.com/appirio-tech/connect-app/blob/dev/src/projects/detail/components/timeline/MilestoneExtensionRequest/MilestoneExtensionRequest.jsx#L55
and https://github.com/appirio-tech/connect-app/blob/dev/src/projects/detail/components/timeline/MilestoneExtensionRequest/MilestoneExtensionRequest.jsx#L63

src/projects/detail/components/timeline/milestones/MilestoneTypeCheckpointReview/MilestoneTypeCheckpointReview.jsx
```diff
@@ -126,7 +127,9 @@ class MilestoneTypeCheckpointReview extends React.Component {
 
     updateMilestoneContent({
       links
-    })
+    }, (isActive ? {
```
maxceem: @nahidshahin This won't work as expected and has to be removed. Instead of this approach described in the next comment has to be used https://github.com/appirio-tech/connect-app/pull/2520#issuecomment-421951565

maxceem:
@nahidshahin case 2 is not supported properly.

See screenshot 

image

We didn't press button `Ready for review` yet, to pass it to the customer but `details.metadata.waitingForCustomer` is already `true`. It's caused by this lines https://github.com/appirio-tech/connect-app/blob/31c9af1e38f9df7078a564600e1a93b029243bb5/src/projects/detail/components/timeline/milestones/MilestoneTypeCheckpointReview/MilestoneTypeCheckpointReview.jsx#L130-L132

which make `details.metadata.waitingForCustomer` = `true` for any milstone update. But instead it has to become true only when we press `Ready for review` button.

src/projects/detail/components/timeline/milestones/MilestoneTypeCheckpointReview/MilestoneTypeCheckpointReview.jsx
nahidshahin: removed at a0ddab7

nahidshahin:
Hi @maxceem address all your feedback at a0ddab7

src/projects/detail/components/timeline/MilestoneExtensionRequest/MilestoneExtensionRequest.jsx
```diff
@@ -41,30 +41,48 @@ export const withMilestoneExtensionRequest = (Component) => {
...
     }
   
     approveExtension() {
       const { extendMilestone, milestone } = this.props
       const content = _.get(milestone, 'details.content')
       const extensionRequest = _.get(milestone, 'details.content.extensionRequest')
-  
+
+      milestone.details = milestone.details || {}
+      milestone.details.metadata = {
```
maxceem: These lines:
- `milestone.details` 
- `milestone.details.metadata`
modify the `milestone` object in Redux Store directly which in this case doesn't bring any harm but in general, it can cause some issues.
Please, avoid modifying `milestone` object directly.

maxceem:
@nahidshahin case 3 has some undesired behavior:

See screenshot:
image

We didn't press button `Ready for review` yet, but the milestone is already marked as `details.metadata.waitingForCustomer` = `true`.

Most likely it's caused by these changes, and they has to be discraded:

image

And instead, it has to be done the same way as for `checkpoint-review`. So only when we press `Ready for review` button, we start waiting for customer action.

image

src/projects/detail/components/timeline/MilestoneExtensionRequest/MilestoneExtensionRequest.jsx
nahidshahin: right, changed at 0740a96

nahidshahin:
@maxceem Fixed logical error of final-designs at 0740a96

src/projects/detail/components/timeline/milestones/MilestoneTypeDelivery/MilestoneTypeDelivery.jsx
```diff
@@ -163,13 +165,16 @@ class MilestoneTypeDelivery extends React.Component {
...
-    const { submitFinalFixesRequest } = this.props
+    const { updateMilestoneContent, submitFinalFixesRequest } = this.props
     const { finalFixRequests } = this.state
 
     submitFinalFixesRequest(
       // submit only non-empty requests
       finalFixRequests.filter((finalFixRequest) => !!finalFixRequest.value)
     )
+    updateMilestoneContent({}, {
+      waitingForCustomer: false,
```
maxceem: This will send duplicate request to update milestone. We can set `waitingForCustomer: false,` inside method `submitFinalFixesRequest` which called above, see https://github.com/appirio-tech/connect-app/blob/dev/src/projects/actions/productsTimelines.js#L286

In particular, backend currently doesn't support requests in parallel, so the second requests will fail.

src/projects/detail/components/timeline/MilestoneExtensionRequest/MilestoneExtensionRequest.jsx
```diff
@@ -77,15 +77,15 @@ export const withMilestoneExtensionRequest = (Component) => {
       const content = _.get(milestone, 'details.content')
       const extensionRequest = _.get(milestone, 'details.content.extensionRequest')
 
-      milestone.details = milestone.details || {}
-      milestone.details.metadata = {
-        ..._.get(milestone.details, 'metadata', {}),
+      const details = milestone.details || {}
+      details.metadata = {
```
maxceem: This line `details.metadata = {...}` still mutates Redux Store.

Here `const details = milestone.details` we put to `details` a link to the same object as `milestone.details` and in next line we do `details.metadata = {...}`.
After this line `milestone.details.metadata` will be updated.

maxceem:
@nahidshahin PR is accepted.

Regarding increasing $. In general, it can be done if scope appears to be bigger than it was evaluated from the beginnig. For this task it doesn't feel like a proper case for increasing. 
The situations when to update `waitingForCustomer` were described in the issue initially and no new situations where added. Also, provided PRs have multiple funcitional issues and required repeating testing for several times.

If the description of the issue was not clear, please ask for details and I would do my best to provide the proper explanation for the task. If during clarification of the task it would appear that there are more things to take care about than it looked from the beginning, it would be also a proper reason for increasing the task prize.

PR #2523, opened 2018-09-17
demo fix

sumitdaga:


sumitdaga:
@maxceem  
like i said it was a demo fix ....it's just to show you the approach ...the issues you mentioned can be fixed ...
but you are correct about this :
> It looks like by current logic if we edit topic title FeedContainer will stop all re-renders in shouldComponentUpdate.
if this is not desired, i can't think of any other approach (except may be adding 1 or 2 more conditions)...let me know if i should continue to work on this PR and make a full PR

thanks

maxceem:
@sumitdaga 

>It looks like by current logic if we edit topic title FeedContainer will stop all re-renders in shouldComponentUpdate.

> if this is not desired

It's just fyi. It's not that it's not desired by itself, but it means that functional inside container wouldn't work (mostly).

> except may be adding 1 or 2 more conditions in shouldComponentUpdate

I think it's ok to to add a couple of conditions. But I have a big concerns that a couple of conditions would be enough. Most likely after `forbidding` to render on edit, now we would have to `allow` again for any other case to be rendered when we need it. Plus in all these cases the whole `FeedContainer` would be re-rendered means all topics would be still re-rendered. Which again would be slow.

So for me looks like this approach wouldn't bring us desired outcome: not re-render all topics when we editing only one topic.

maxceem:
I've tested the current implementation. Most things work and there are no re-renders on typing. There are some issues though.

1. [Major] When we change one comment/topic and after change another comment/topic and after discard our changes for one of these comment/topics, we can leave the page without warning, see demo https://monosnap.com/file/fPohvBXZyBah7z4UUP6f3BZLnpBLbw

2. [Major] When we didn't change anything, scrolled browser window and created a post by another user, popup with a suggestion to refresh the page should be shown. But it doesn't.

3. [Major] After 2. if we scroll up, the page will be reloaded and reloading will happen again and again in a loop.
    Demo video for case 2. and 3. https://monosnap.com/file/SHUcxny6mDScnRwE7PvdSQiZfaPLmV (please ignore black screen in the video)

4. [Minor] When we create a post or comment, update a topic or comment all the topics are being re-rendered: https://monosnap.com/file/B1ZRi0gGFm2yWfv6jA9VsubCGJnZq3

5. [Trivial] There is waring in console in some cases:
    ```
    warning.js?6327:33 Warning: Failed prop type: Invalid prop `preventShowing` of type `number` supplied to `PostsRefreshPrompt`, expected `boolean`.
    ```

maxceem:
There are many things appears to be tights together. And it could be that direct refactoring would be too wast and thus prone to errors.
So I guess we will try to approach it from the other side and will try to move out things one by one by smaller portions keeping the functionality.
And only at last refactor FeedContainer to remove re-renders.

So I close this PR and we will use it only for references and probably some solutions could be used from this PR.

Issue #2541, opened 2018-09-25
Cannot find notification rule for eventType


maxceem:
There is an error in console: 
```
Failed to load notifications. Cannot find notification rule for eventType 'notifications.connect.project.post.deleted' version '2'.
```

NOTE. Because of this error all notifications are not shown.

There are two things to fix:
1. Add rule for `notifications.connect.project.post.deleted` version `2`.
2. I think code shouldn't fail when there is no appropriate rule for notification. Probably better to log warning using `console.warn()` instead of throwing an exception in such cases.

maxceem:
@vikasrohit @RishiRajSahu 

After looking closer into this, I'm not sure, that above I suggested the correct way to handle it.

This notification type `notifications.connect.project.post.deleted` previously wasn't added `tc-notifications`. Could you please share any details what should we do with this notification type in connect app?

vikasrohit:
@maxceem I was doing some changes around events in `tc-notifications` and `tc-project-service`. Also, made some changes in `tc-message-service`, however, didn't explicitly made any changes for this particular event. I would check and let you know how we want to handle this.

maxceem:
@vikasrohit 

> I think we can handle it similar to what we do for topic deleted

we don't have notifications for topics deleted either.

> we can we handle it like post updated

we can, but not sure if that would be desired way. For post updates we show notifications like this:
- `<strong>{{userFullName}}</strong> edited post`
- or when bundled: `{{#showMore __history__ 3}}<strong>{{fallback userFullName userHandle}}</strong>{{/showMore}} edited {{bundledCount}} posts`

Do you want it to be shown simillar way?
- `<strong>{{userFullName}}</strong> deleted post`
- or when bundled: `{{#showMore __history__ 3}}<strong>{{fallback userFullName userHandle}}</strong>{{/showMore}} deleted {{bundledCount}} posts`

vikasrohit:
> Do you want it to be shown simillar way?

vikasrohit:
@maxceem closing this issue for now so as you have handled the missing notifications already, we should create separate issues for specific missing notifications to discuss more specifically. Please feel free to create new ones as you find them.

PR #2542, opened 2018-09-25
issue #2541 -  log warning using console.warn() instead of throwing an exception when a rule for notification is not found

maxceem:
I propose to log warning instead of throwing an exception when a rule for notification is not found.
Otherwise, notification functionality just got broken.

I think it's better to merge this PR without waiting for the second part of issue #2541 so notifications are not broken in `dev` branch.


maxceem:
@vikasrohit just in case, I think this one has to be included into the release, as it fixes an issue which may break notifications functionality.

Issue #2522, opened 2018-09-17
Need bundling support for email group


vikasrohit:
### Expected behavior
User should be able to configure different bundling behaviour per event group e.g. `Post and replies` could be immediate delivery events while `Project plan` events could be bundled by 24 hours.

### Actual behavior
As of now it allows same bundling setting for all events i.e. if user chose to bundle by 24 hours, it would bundle all events by that duration.

### Steps to reproduce the problem
- Open [Notification Settings](https://connect.topcoder.com/settings/notifications) page after logging in and observe that we have only global level of bundling support

### Screenshot/screencast
<img width="855" alt="screen shot 2018-09-17 at 10 59 30 am" src="https://user-images.githubusercontent.com/2417632/45607250-c9cfbd00-ba68-11e8-95c5-fcfced526f7a.png">

--

#### Environment
- OS: Any
- Browser (w/version): Any
- User role (client, copilot or manager): Any
- Account used:  N/A

Aha! Link: https://topcoder.aha.io/features/TCCONNECT-481

vikasrohit:
fyi @acshields @vic-topcoder @RishiRajSahu  @gondzo

acshields:
@vikasrohit - Love the design and logic outline above.  I just went into Connect to try & find how to access the Notification Settings page from a project and I couldn't locate it.  I assumed it would be under Profile/Settings.  Where is it?

gondzo:
@acshields click the notifications icon, click all notifications and then click the notifications settings button

acshields:
Hi @vikasrohit @vic-topcoder - Per our conversation this morning:

1. I liked the second option you showed me (where the user can select which notification types they want bundled, but they can only set one cadence for the bundles to be sent). See design above.

2. We need to add language/direct link to the notifications tab that indicates to the customer that they have the option to modify notification settings, as the current workflow is not intuitive.

<img width="1420" alt="screen shot 2018-09-18 at 10 21 54 am" src="https://user-images.githubusercontent.com/41118149/45705362-4082be00-bb2e-11e8-9028-c29543ff9301.png">

3. @vic-topcoder - Do you think we need the "Web" notifications selection? Matt raised a good point stating that now that we will have email notification on/off/bundled options we should make web notifications required.  Thoughts?

vic-topcoder:
[...]

PR #2550, opened 2018-09-26
Issue #2522 f2f - fix cutting last dropdown

maxceem:
Issue #2522 f2f - fix cutting last dropdown


vikasrohit:
Thanks @maxceem for quick fix.

PR #2551, opened 2018-09-26
issue #2522 - additional fixes

maxceem:
Additional fixes for issue #2522

@vikasrohit 
> btw, it seems we can not save `every 10 minutes` setting in notification settings now

I think the issue was not connected with `every 10 minutes` but with value for `bundlePeriod`, as a server doesn't accept `null`. Also, I've noticed not correct behaviour of default value for email settings of messaging notifications.

- fixed setting `bundlePeriod` to `off` by using an empty string `''` instead of `null`
- fixed default value for messaging notifications by email:
   - only for notifications related to messagin, if settings is not defined, then by default email notifications are `off` (bundlePeriod is '')
   - for the rest of notifications, default value is `daily`

### DIFF
--- a/src/routes/settings/routes/notifications/components/NotificationSettingsForm.jsx
+++ b/src/routes/settings/routes/notifications/components/NotificationSettingsForm.jsx
@@ -146,18 +146,22 @@ const initSettings = (notInitedSettings) => {
       if (!notifications[type][serviceId]) {
         notifications[type][serviceId] = {}
       }
-      if (_.isUndefined(notifications[type][serviceId].enabled)) {
-        notifications[type][serviceId].enabled = 'yes'
-      }
-      if (_.isUndefined(notifications[type][serviceId].bundlePeriod)) {
-        notifications[type][serviceId].bundlePeriod = 'daily'
+
+      // for all messaging events, email notifications are off by default
+      if (serviceId === 'email' && _.includes(messagingTypes, type) && _.isUndefined(notifications[type][serviceId].enabled)) {
+        notifications[type][serviceId].enabled = 'no'
+        notifications[type][serviceId].bundlePeriod = ''
+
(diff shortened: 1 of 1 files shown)


vikasrohit:
@maxceem we explicitly enabled the messaging emails enabled by default because that is something users's can not afford to miss. Email is enabled for all events by default and bundling is set to `daily` for all events except messaging events where we assuming it to be immediately sent.

maxceem:
@vikasrohit ok I'll adjust the fix, so to sum up:
- all notifications by email have to be enabled by default
- the bundling period has to be set to `daily` for all email notifications except of email notifications related to messages, for them `bundlePeriod` has to be `immediately`, 
did I get it right?

vikasrohit:
yes @maxceem, your understanding is correct. However, what I want to convey is that current dev branch has that thing almost working i.e. we are assuming missing setting for bundlePeriod, in case of messaging events, to be immediately sent. However, I think on the front end is not showing the `Immediately` option as selected one.

maxceem:
@vikasrohit it seems for me, currently on DEV for all messaging email notifications no matter if it has value from backend or no, `bundlePeriod` is set to `null`: https://github.com/appirio-tech/connect-app/blob/dev/src/routes/settings/routes/notifications/components/NotificationSettingsForm.jsx#L157-L160
```
   // for all messaging events, defaults emailBundling to off
    if (_.includes(messagingTypes, type)) {
      notifications[type]['email'].bundlePeriod = null
    }
```

vikasrohit:
> it seems for me, currently on DEV for all messaging email notifications no matter if it has value from backend or no, bundlePeriod is set to null

Just to close the thread @maxceem, that piece of code actually disables bundling only for messaging events. I have made an line comment as well 😬 . Setting it null was effectively disabling bundling and assuming `immediately` as bundling value. However, as I said previously that it does not render the correct option in dropdown though.

That said, I think final UI looks good. Merging.

PR #2558, opened 2018-09-28
Feature/login flow

gondzo:


package.json
```diff
@@ -78,7 +78,7 @@
     "webpack-merge": "^4.1.1"
   },
   "dependencies": {
-    "appirio-tech-react-components": "github:appirio-tech/react-components#feature/connectv2",
+    "appirio-tech-react-components": "github:appirio-tech/react-components#feature/loginFlow",
```
vikasrohit: I guess this should be reverted now back to connectv2

src/components/Home/Home.jsx
```diff
@@ -31,7 +31,6 @@ class Home extends React.Component {
             <p>Companies and agencies, from the world’s largest to the Valley’s newest, use crowdsourcing to deliver high-quality assets, faster.</p>
             <p>Connect guides you through the entire crowdsourcing process, from entering requirements to receiving final deliverables, and it facilitates collaboration between your project team and Topcoder community members working on your project.</p>
             <div className="button-bar">
-              <Link to="/new-project" className="tc-btn tc-btn-primary">Start a project</Link>
```
vikasrohit: Is there any requirement for removing this link from home page? I think instead it should now point to the new connect registration page.

src/config/constants.js
```diff
@@ -439,6 +439,7 @@ export const SEGMENT_KEY = process.env.CONNECT_SEGMENT_KEY
  */
 export const DOMAIN = process.env.domain || 'topcoder.com'
 export const CONNECT_DOMAIN = `connect.${DOMAIN}`
+export const CONNECT_MAIN_PAGE_URL = `http://connect.${DOMAIN}`
```
vikasrohit: Could we reuse the `CONNECT_DOMAIN` (declared just above this one) variable instead?

src/projects/create/components/ProjectWizard.jsx
```diff
@@ -403,8 +426,9 @@ class ProjectWizard extends Component {
   }
 
   render() {
-    const { processing, showModal, userRoles, projectTemplates, projectTypes } = this.props
+    const { processing, showModal, userRoles, projectTemplates, projectTypes, projectId } = this.props
     const { project, dirtyProject, wizardStep } = this.state
+    const params = this.props.match.params
```
vikasrohit: Minor change, we can move `match` to the line#429

src/projects/create/components/ProjectWizard.jsx
```diff
@@ -46,19 +49,32 @@ class ProjectWizard extends Component {
...
   componentDidMount() {
-    const { onStepChange, projectTemplates } = this.props
+    const { onStepChange, projectTemplates, createdProject } = this.props
     const params = this.props.match.params
 
     // load incomplete project from local storage
     const incompleteProjectStr = window.localStorage.getItem(LS_INCOMPLETE_PROJECT)
-    if (incompleteProjectStr) {
+    
+    if ((params && params.project === 'submitted') || createdProject) {
```
vikasrohit: Just curious, what makes us using two different ways of determining that project is submitted.

src/projects/create/containers/CreateContainer.jsx
```diff
@@ -307,6 +344,30 @@ class CreateContainer extends React.Component {
       />
     )
   }
+
+  render() {
+    const { wizardStep, projectType } = this.state
+    let type = 'unknonw'
```
vikasrohit: typo `unknonw` => `unknown`

src/projects/create/containers/CreateContainer.jsx
```diff
@@ -307,6 +344,30 @@ class CreateContainer extends React.Component {
...
+    } else if (wizardStep === ProjectWizard.Steps.WZ_STEP_PROJECT_SUBMITTED) {
+      type = ViewTypes.projectSubmitted
+    }
+
+    if (projectType && projectType.icon) {
+      projectType.iconUI = <ProjectTypeIcon type={projectType.icon} />
+    }
+
+    return (
+      <Wizard wrapperClass="WizardCreateProject" type={type} vm={{userHandle: this.props.userHandle, projectType: projectType || {}}} >
```
vikasrohit: I guess `vm` is used because we are sharing the stuff between Angular and React. Could we better separate it into separate props i.e. `userHandle` and `projectType`? If we can not do it in a generic way which can be shared between Angular and React, can we do it specifically for React?

package.json
gondzo: Yes if we merge the PR at react components repo, IMO we can leave it as is to check if there are any regressions with the new styling introduced there.

src/components/Home/Home.jsx
gondzo: good point. will fix

src/config/constants.js
gondzo: sure

package.json
vikasrohit: @gondzo we are deploying the release today, so I was thinking of merging react-components as well today. Anyway, I would start with PR only without this change and see if things works as expected.

vikasrohit:
@gondzo I have fixed only required things and created a new PR https://github.com/appirio-tech/connect-app/pull/2562.

PR #2593, opened 2018-10-10
fixes #2150

### LINKED ISSUE #2150: Display the full project name if there is enough space on the top bar
Due to the interface changes we have more space on the top bar. Let's use it to our full advantage. Right now projects with longer name don't quite fit as the name space is too constrained.

- [ ] Expand the container for the project name to fit on the entire top bar.

screen shot 2018-07-05 at 7 32 01 pm

sumitdaga:


sumitdaga:
@maxceem 
ok i will take a look!

Issue #2518, opened 2018-09-14
[$40] Jumbled/Incorrect Words Appearing in Projects


acshields:
Words are becoming jumbled when creating a Connect project on a customers behalf.
<img width="1103" alt="screen shot 2018-09-14 at 11 18 07 am" src="https://user-images.githubusercontent.com/41118149/45559825-5aec2d00-b7f8-11e8-8a29-803875bffa02.png">

(https://connect.topcoder.com/projects/7060)

Aha! Link: https://topcoder.aha.io/features/TCCONNECT-480

maxceem:
It's on you, you can work on it.

gets0ul:
How to reproduce this issue?  
Any screencast?

maxceem:
[Create a new project](http://local.topcoder-dev.com:3000/new-project) include in all the fields special characters, for example, `&`.

After the project is created go to the `Scope` tab and `&` will be replaced with `&amp;`.

See screencast https://monosnap.com/file/QTSn1Q1tFMLI4XunpVgCdWH0CuaJow

You can check the sceenshot of issue in the top comment, to get a hint which other characters are replaced with their HTML numeric entities.

gets0ul:
@maxceem PR is done and ready for review.  
Thanks

maxceem:
@gets0ul I've reviewed the PR, please have a look https://github.com/appirio-tech/connect-app/pull/2602#pullrequestreview-164484153
Thanks.

maxceem:
@RishiRajSahu fyi, it's done and has to work good for all the characters. Though the solution is to handle it in every place individually https://github.com/appirio-tech/connect-app/pull/2602/files. I don't see a better way for this so far.

RishiRajSahu:
@maxceem agree and looks good to me.

PR #2602, opened 2018-10-10
issue_2518 Fix Jumbled/Incorrect Words Appearing in Projects

gets0ul:
Fix for issue #2518

### DIFF
--- a/src/components/TopBar/ProjectToolBar.js
+++ b/src/components/TopBar/ProjectToolBar.js
@@ -69,10 +69,10 @@ class ProjectToolBar extends React.Component {
             </div>
           </div>
           {project && project.name && <div className="bar-column project-name">
-            <span ref="name" onMouseEnter={this.onNameEnter} onMouseLeave={this.onNameLeave}>{project.name}</span>
-            {isTooltipVisible && <div className="breadcrumb-tooltip">{project.name}</div>}
+            <span ref="name" onMouseEnter={this.onNameEnter} onMouseLeave={this.onNameLeave}>{_.unescape(project.name)}</span>
+            {isTooltipVisible && <div className="breadcrumb-tooltip">{_.unescape(project.name)}</div>}
           </div>}
-          <div className="bar-column project-name mobile"><span>{project.name}</span></div>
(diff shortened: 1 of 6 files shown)


gets0ul:
Added fix for project list.

vikasrohit:
@maxceem Just out of curiosity, could these changes cause the script injection in the platform?

vikasrohit:
Also, there might be one more place where the same cause is causing the issue. It is in ref code, if you try adding `R&D` as ref code, it renders as `R&amp;D`. It might be a backend issue as well.

maxceem:
@vikasrohit yeah, I've also was curious about that. I've checked all the places and think it's safe as we use this text inside
- react components which don't let us insert HTML
- values of input/textarea
- property of `React-Text-Truncate` which also inserts text using react components https://github.com/ShinyChang/React-Text-Truncate/blob/master/src/TextTruncate.js#L233

maxceem:
> Also, there might be one more place where the same cause is causing the issue. It is in ref code, if you try adding R&D as ref code, it renders as R&amp;D

Oh, I've missed that, thanks. I've fixed it in this commit 85f18529dc84fe70ee348507577219a5df4272c0

> It might be a backend issue as well.

Yes, encoding is done by the backend, but I think it's a good thing for us, as it keeps us safe from injections.

vikasrohit:
> Oh, I've missed that, thanks. I've fixed it in this commit 85f1852

Thanks @maxceem However, I am worried about ref code thing because if it is encoded in database, it is tricky and confusing to query projects by ref code in our reporting system because we are rendering `R&D` while in database it is `R&amp;D`. Is it possible, on front end, to keep it `R&D` at both places?

maxceem:
@vikasrohit I think we cannot handle it frontside.

Even if I send API request using postman with & presented it would be returned from the backend as `&amp;`

So I think we have to fix it on backend.
Though we can check if backend encodes symbols before saving data, or before returning it. So there is small probability that data is stored original way and searching would work correctly.

Otherwise we can check if it's possible to disable encoding in the backend for one field REF only.

Issue #2524, opened 2018-09-19
[$50] Use productCategories instead of projectTypes for product classification


maxceem:
## Overview

When we add a new phase to a project we see the list of products which is divided into categories:

![image](https://user-images.githubusercontent.com/146016/45727864-1ac1eb80-bbf8-11e8-9f3f-ef7bc74536a4.png)

It suppose to use `productCategories` to devide products into these sections and show titles of the sections. But instead it uses `projectTypes` now.

![image](https://user-images.githubusercontent.com/146016/45728204-9e300c80-bbf9-11e8-94f6-9c1ce4ceace8.png)

This most likely happens due to a confusion caused by the current code. Inside code we use term `projectCategories` instead of `projectTypes` so it's easy to confuse **project***Categories* with **product***Categories*.

### 1. Products on the "add a new phase" popup should be categorised using `productCategories`.

- Action `loadProjectsMetadata` already loads all the metadata using `/v4/projects/metadata`. The only thing it doesn't populate Redux Store with `productCategories` https://github.com/appirio-tech/connect-app/blob/cf12/src/reducers/templates.js#L21. So it has to be adjusted to populate `templates.productCategories` in Redux Store.
- After that we have to adjust code of "add a new phase" [popup](https://github.com/appirio-tech/connect-app/blob/cf12/src/projects/create/components/SelectProductTemplate.jsx) to use `templates.productCategories` instead of `templates.projectCategories`.

### 2. Rename `projectCategories` to `projectTypes` everywhere in project code

This has to be done in the Redus Store `templates.projectCategories -> templates.projectTypes` and in all relative places to avoid future confusion.

## References:
- https://github.com/appirio-tech/connect-app/issues/2472 - implementation of loading metadata using `/v4/projects/metadata`
- https://github.com/appirio-tech/connect-app/issues/2139 - implementation of categories on "add a new phase" popup

Aha! Link: https://topcoder.aha.io/features/TCCONNECT-482

vikasrohit:
> After that we have to adjust code of "add a new phase" popup to use templates.productCategories instead of templates.projectCategories.

@maxceem  last instance of `templates.projectCategories` in above statement should be `template.projectTypes`

I didn't understand the second requirement to rename projectCategories to projectTypes. I think we need both projectTypes and productCategories.

maxceem:
@vikasrohit yeah, it all about confusion brought by the terms we are using `projectCategories` vs `projectTypes`.

[...]

PR #2613, opened 2018-10-11
issue_2524 Add productCategories, projectCategories -> projectTypes

gets0ul:
Fix for issue #2524


vikasrohit:
Thanks @gets0ul. It is always tricky to do refactoring.

Issue #2511, opened 2018-09-14
[$50] Sidebar jumps when we mark milestone as completed


maxceem:
When we mark some milestone as completed and loading indicator is being shown, the sidebar jumps, see demo video https://monosnap.com/file/V09BUP4V3RHO3nHKwdnv5TEsNSCH8G

![image](https://user-images.githubusercontent.com/146016/45526215-db1a8e80-b808-11e8-848f-4d0c4596a86a.png)

It happens not always, probably when browser window has been scrolled and we change the last phase's milestone.

Possible related with https://github.com/appirio-tech/connect-app/issues/2149, but that one I couldn't reproduce.

Aha! Link: https://topcoder.aha.io/features/TCCONNECT-479

vikasrohit:
@maxceem I think we got this fixed in CF 11. Can you please check taking latest from dev?

maxceem:
@vikasrohit I can still reproduce it for this project https://connect.topcoder-dev.com/projects/7301/plan if I complete milestone in the last phase.

Also, I've pulled the latest div locally and can reproduce the issue too, with the mentioned project.

vikasrohit:
Ohk. Then this might be an edge case for #2422

sumitdaga:
@maxceem 
please assign this to me!

sumitdaga:
@maxceem 
please unassign me!
i would like to take a dig at other issues first

Gauravseta:
@maxceem pls assign to me thnaks

maxceem:
As the issue hasn't been completed during 12 hours and no comment was provided, this issue is now open for pick up.

zjuasmn:
works on it

zjuasmn:
done, please check #2621

RishiRajSahu:
Closing merged to dev.

PR #2621, opened 2018-10-11
#2511 Sidebar jumps when we mark milestone as completed

zjuasmn:
This is a bug from `react-stickynode` so I create a wrapper component to update it after each scroll event.

<https://github.com/yahoo/react-stickynode/issues/102#issuecomment-362502692>


vikasrohit:
Awesome. Thanks @zjuasmn @maxceem

Issue #2091, opened 2018-06-26
[$30] There should be only 1 message to dismiss on a project dashboard.


vic-topcoder:
### Expected behavior
When the user comes back and there's a project status update message, they can dismiss it. There should be only 1 message at all times. In effect if there have been several messages, we want to replace them with only the latest.

### Actual behavior
Upon dismissal of the last status update message, the previous one is loaded and the user has to click unspecified number of times to clear them old.

### Steps to reproduce the problem
Change a project status several times and view the project as a member. Try dismissing the messages to observe the behavior.

### Screenshot/screencast
![screen shot 2018-06-25 at 7 56 06 pm](https://user-images.githubusercontent.com/8484061/41886980-8d42368e-78b3-11e8-8f2c-efda4c39a0aa.png)

Aha! Link: https://topcoder.aha.io/features/TCCONNECT-474

maxceem:
Note, to test it, we can mark two milestones as completed at the same time which will produce two notifications.

maxceem:
This is one is open for pick up now.

khushimehta03:
i wish to work on it and complete this time @maxceem

maxceem:
This issue is now opened for pick due to inactivity.

sumitdaga:
Can you assign me ?

sumitdaga:
@maxceem  
how to mark a milestone complete?

maxceem:
@sumitdaga you can use this project or create a new one http://local.topcoder-dev.com:3000/projects/7300/plan

- Go to Plan tab (if not yet)
- Open two phases which are `In Progress` state
- There would be milestones which have `Mark as completed` or `Complete Review` buttons which will complete milestones. Note some buttons require one more click to confirm the action.
- Click these two buttons one by one fast to complete two milestones at the same time. There would be two notifications.

See demo video https://monosnap.com/file/OORAzuTp9CwtBD3CpdB7fA6nS3eCBQ

Also if you already don't have active phases for a project, you may need to add a new phase using the `Add New Phase` button at the bottom.
After you've added phase, you have to activate it to be able to complete milestones. To activate phase, click on the pencil on the phase, and after choose status `In Progress`. After that, you could complete milestones inside.

Let me know if it helps.

sumitdaga:
@maxceem  
thats for your input
ok i was able to reproduce just like the demo video you shared  (initially i think the red boxes in the screenshot shared by vic-topcoder confused me)  
[...]

PR #2631, opened 2018-10-15
Issue #2091 - There should be only 1 message to dismiss on a project dashboard. 

maxceem:
@RishiRajSahu, during CF-12 issue https://github.com/appirio-tech/connect-app/issues/2091 has been accomplished in a separate branch.

I've created this PR to not forget about it.


vikasrohit:
@maxceem @RishiRajSahu merging the PR to reduce outstanding things. The changes are very minimum and in case we need to revert back to current state, we can easily do that by updating the count in alert config.

Issue #2642, opened 2018-10-18
Invalid progress data in left panel


vikasrohit:
<img width="1280" alt="screen shot 2018-10-18 at 11 00 20 am" src="https://user-images.githubusercontent.com/2417632/47133149-08090800-d2c5-11e8-998b-c996fee95a77.png">

Project: https://connect.topcoder-dev.com/projects/7210

@maxceem @RishiRajSahu it might be because of bad data during testing, however, it would be great if we can have a quick look into it by debugging this project in local env.

maxceem:
@vikasrohit looking into it.

RishiRajSahu:
fyi @maxceem it is actually happening because of phase are delayed, and its considering the duration so far(till today) if phase is active and gone beyond its expected endDate.
Also, I found one more issue, that is with backend, For `App Design` the correct duration is 31 while the backend is returning 33

maxceem:
@RishiRajSahu as I see the reason is, that it takes data from `phase.duration` for completed phases instead of using milestones to calculate the actual duration of the completed phases.

By current logic actual duration is restricted by planned duration of the phase. So let's say some active phase is delayed. The actual duration for such phase will be returned as a planned duration for that phase, not as time until today.

![image](https://user-images.githubusercontent.com/146016/47134516-a5713580-d2df-11e8-8283-fdf10faa265e.png)

This leaves the question open how should we calculate current duration? If we include delayed time the value will be bigger than planned duration. I guess in such cases we should update the progress bar to show delayed time instead `x days of y`, what do you think?

RishiRajSahu:
@maxceem Agree, looked more into it, it has been due to old data, right now backend is working fine, Tested by adding new phase to the project and its duration is coming out to be correct. So to handle it changes we need are 
1. Show the message `Delayed by (109-41) days` having progress bar as red coloured

maxceem:
@RishiRajSahu so we shouldn't restrict the maximum duration of phases by its planned endDate as it's done now, correct? When we calculate the total spent time.

maxceem:
@RishiRajSahu I think we have to agree on a few moment regarding calculating the total duration:
1. **Planned duration**
   1. Imagine we have 2 planned phases with a gap between them:
       - Sep 1 - Sep 3 (3 days)
       - a gap
       - Sep 5 -Sep 7 (3 days)

[...]

PR #2643, opened 2018-10-18
issue #2642 - Invalid progress data in left panel

maxceem:
issue #2642 - Invalid progress data in left panel

Logic now looks pretty clean and we summed up with our approach for duration calculation, so hopefully this time it would be our last fix for this :-)

### DIFF
--- a/src/helpers/projectHelper.js
+++ b/src/helpers/projectHelper.js
@@ -84,7 +84,11 @@ export function formatProjectProgressProps(project, phases, productsTimelines) {
   // duration in days
   const plannedDuration = getProjectPlannedDuration(phases, productsTimelines)
   const actualDuration = getProjectActualDuration(phases, productsTimelines)
-  const labelDayStatus = `Day ${actualDuration} of ${plannedDuration}`
+  const delay = actualDuration - plannedDuration
+  const labelDayStatus = delay > 0 
+    ? `Delayed by ${delay} day${delay === 1 ? '' : 's'}`
+    : `Day ${actualDuration} of ${plannedDuration}`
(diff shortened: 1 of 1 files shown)


src/helpers/projectHelper.js
```diff
@@ -194,6 +146,44 @@ export function projectPlannedDuration(nonDraftPhases, productsTimelines) {
...
+  const today = moment()
+  const phasesActualData = activeAndCompletedPhases.map((phase) => {
+    const product = _.get(phase, 'products[0]')
+    const timeline = _.get(productsTimelines, `[${product.id}].timeline`)
+    const { startDate, endDate } = getPhaseActualData(phase, timeline)
+
+    return { 
+      startDate,
+      // for all active phases use today as endDate
+      endDate: phase.status === PHASE_STATUS_ACTIVE ? today : endDate,
```
RishiRajSahu: @maxceem does this check takes care of avoid considering overlapping days ? there could be scenario when some phase is active and `today` and `startDate` of date phase is accommodating some delivered phase ?

maxceem: @RishiRajSahu I'm not sure which particular situation do you mean in this particular case. But generally, I take care about overlapping between phases. As basically we need to find minimum date startDate of all phase - we count the time since that. And we have to find maximum endDate - we count the time until that. And after just calculate the difference. So any overlapping wouldn't make any harm to such calculations.

RishiRajSahu: @maxceem ohh, i think the my comments got published :) , I was in the process of reviewing and thought like my comments would publish after I approve/review the changes from top. Never mind I got that, so we are good.

Issue #1985, opened 2018-05-01
Not all notifications are dismissed when the user visits the target page


vic-topcoder:
### Expected behavior
When I click/tap on a notification, and the target page loads, the notification should be dismissed.

### Actual behavior
Currently works for posts, but not for project status, members joining a project, probably others. Let's try to unify the behavior of all notifications in the system to be the same.

Aha! Link: https://topcoder.aha.io/features/TCCONNECT-445

maxceem:
Also, I see notifications are not dismissed for `added a link to your project`.

vikasrohit:
@maxceem Do you think this issue is same as https://github.com/appirio-tech/connect-app/issues/2123

gondzo:
We should implement this in a generic way so it's easy to extend when we add notifications for different target pages in the future (ex #2178). 
IMO a good approach would be to listen for route changes in react-router and dismiss notifications that match current route. Any better ideas @vikasrohit ?

vikasrohit:
Yep, handling it in a generic way would be good, however, I am not sure what exactly that would be. How are we dismissing the notifications right now?

> IMO a good approach would be to listen for route changes in react-router and dismiss notifications that match current route

I am not sure if all notifications would have separate routes i.e. there might be multiple notifications sharing the same route, so it would be difficult to determine which notification has to be dismissed.

vikasrohit:
But wouldn't it dismiss notifications for all phases when there are posts in multiple phases and user clicks on one of the notifications.

gondzo:
@vikasrohit yes, it would dismiss notifications for all phases with the above approach. Would it make sense to add some kind of 'event emitter' logic to components that display the target for a notification? For example we can have Feed container emit "message seen" event for each displayed message, files list component would emit "file seen" event for all seen files. Then we would have some main logic to process those 'events' and dismiss notifications.
any other ideas?

maxceem:
@vikasrohit @gondzo FYI
For implementing [Add notification visual cues to project plan cards](https://github.com/appirio-tech/connect-app/issues/2401) we have to mark post notifications as read, when user opens the tab with posts for a particular phase.

So far I've implemented some helper component to dismiss notifications for some mounted i. e. visible component: https://github.com/appirio-tech/connect-app/blob/issue-2401/src/components/NotificationsReader/NotificationsReader.jsx

[...]

Issue #2649, opened 2018-10-23
[$50] Hide dismissed notifications without waiting for server response 


maxceem:
Follow up from https://github.com/appirio-tech/connect-app/issues/2607#issuecomment-429810521

To make it looks smoother we have to additionally implement `hide notification without waiting for the end of server request (if it fails we can show red alert popup and show notification back)`
For consistency, we have to implement it for cases when we show animation and we don't show animation:
1. Notifications dropdown: mark one notification as read
2. Notifications dropdown: mark all notifications as read
3. [Notifications page](http://local.topcoder-dev.com:3000/notifications): mark one notification as read
4. [Notifications page](http://local.topcoder-dev.com:3000/notifications): mark all notifications of one project as read

Implementation thoughts:
- we will need to save data to the Redux store about notifications which are preliminarily marked as read, and based on that data hide notifications in Dropdown and Notifications page
   1. maybe we can add some additional property to notifications objects in Redux store to keep track of preliminary read notifications, this will require updating existent componets so they will hide such notifications with this property
   2. or we can directly change `read` property of notifications in the redux store as they are really already `read` and in `FAILURE` action we can revert value for `read` property. This is possible as inside action unction we know the whole list of notification to mark, so we can pass this whole list to the `PEDNING`, `SUCCESS` and `FAILURE` actions as we dispatch them all inside one function https://github.com/appirio-tech/connect-app/blob/dev/src/routes/notifications/actions/index.js#L50. In this case we don't have to update other code I think. But theoretically it could affect other code which could detect such cases as `new unread notiifcations arrive`.

maxceem:
@RishiRajSahu here is the answer to your question:

> @maxceem first to understand it, so your point is if we toggle read state of notification again in case of failure from backend it will have effect on other components as treating them as new notifications, right ? and using approach 1 we can avoid it, right ?

yes, that's right, though for now I think we don't have such cases, but theoretically it's possbile to add code like this:
```
componentWillReceiveProps(nextProps) {
   if (_.filter(this.props.notifications, 'read').length === 0 && _.filter(nextProps.notifications, 'read').length > 0) {
      // do something in case there are new unread notifications appears
   }
}
```

> what are the other components sharing the notifications with notification drop-down ?
- Notification page
- https://github.com/appirio-tech/connect-app/blob/dev/src/projects/detail/components/PostsRefreshPrompt/PostsRefreshPrompt.jsx
[...]

Issue #2455, opened 2018-08-29
Add notification visual cues to phase timeline tab


maxceem:
This is the second part of this issue https://github.com/appirio-tech/connect-app/issues/2401

Add the red dot to highlight that `Timeline` has some updates.
When we open such tab we have to mark the related notifications as read and hide the red dot as we already saw the updated content.

![image](https://user-images.githubusercontent.com/146016/44770657-4c1c3e00-ab9b-11e8-8c2a-315b6ad67685.png)

This issue requires support of additional event types which is tracked by the next issue https://github.com/appirio-tech/connect-app/issues/2426

continued from https://github.com/appirio-tech/connect-app/issues/2301

**Update:** Restricting the scope of the issue to timeline tab only. We would handle the specification tab separately later as that tab might be revamped soon or later.

Aha! Link: https://topcoder.aha.io/features/TCCONNECT-408

maxceem:
@RishiRajSahu 
I've tried to run connect-app using branch https://github.com/appirio-tech/connect-app/tree/feature/plan-timeline-events-settings where new notifications are implemented.

But new notifications never come when I activate phases or complete milestones by manager user and check for notifications by customer user. I used this project for testing http://local.topcoder-dev.com:3000/projects/7279

Is there any chance these events weren't added to Kafka, or currently some version of `tc-project-service` deployed which doesn't support these events yet?

RishiRajSahu:
@maxceem yes there could be two possibilities - 
- These events weren't added to kakfa
- Notifications service is not able to listen to kafka

I have to check, will let you know soon.

vikasrohit:
@maxceem I am able to see the phase completed/activation and milestone completed/activation web notifications for my project when I perform the action as copilot and check the notifications as customer.

maxceem:
To sum up.

The whole list of notification types when we have to add a cue to `Timelines` tab is:
- `connect.action.timeline.adjusted`
- `connect.action.timeline.milestone.transition.active`
- `connect.action.timeline.milestone.transition.completed`
- `connect.action.timeline.milestone.waiting.customer`

All of these notififications with mentioned types we will mark as read when we open `Timelines` tab.

So far it can be implemented similar way as it done now for `Discussions` tab.

RishiRajSahu:
@maxceem yeah the list looks good, however other than above events, i found few milestones events as well - 
`connect.action.timeline.milestone.added` NOT NEEDED
[...]

Issue #2410, opened 2018-08-22
Enforce minimum design selection at checkpoint review milestone 


vikasrohit:
### Expected behavior
It should not enable the `complete review` button until the required number (right now it is 5) of designs are selected. Only exception is when the available number of options is lower than the required number e.g. if there are only 3 design options to select from, it should enable the `complete review` button only when all 3 are selected.

### Actual behavior
It enables the button even without selecting a single design.

### Steps to reproduce the problem
- Follow the milestones for a design phase and observe the behaviour at checkpoint review milestone.

### Screenshot/screencast
N/A

--

#### Environment
- OS: Any
- Browser (w/version): Any
- User role (client, copilot or manager): Copilot or Manager
- Account used: N/A

vic-topcoder:
I think we should disable "submit for review" for customers as the specification is hidden on the phase cards. @acshields please chime in

vic-topcoder:
Yeah, poor wording. Ignore my previous comment. The button should not be enabled until the customer selects the required number of designs (review, final designs).

PR #2659, opened 2018-10-30
Issue #1985 - notifications reader

maxceem:
Issue #1985 - notifications reader

- Implemented general `NotificationsReader` which marks notifications as read by providing a criteria:
  - `NotificationReader` is not connected to store directly
  - `NotifcationsReader` only dispatches action when there are any notifications to be marked as read, otherwise it doesn't dispatch any actions
  - notifications are being marked as read, only at the time when `NotifcationsReader` is mounted. This apeears to cover all our current cases. There is no need to `listen` for new notifications.
- `NotificationsReader` marks notifications as `read` per this table https://docs.google.com/spreadsheets/d/1olFRCgXyZy6DTBWmFCJF-MXuhxh80y5JWOKvlSpgguU/edit#gid=1005864840. Note, that there are notifications marked by red color. They don't come to Connect App currently.
- **Important**. The way how we store notifications in the redux store was refactored.
   - before: we stored already bundled notifications with rendered texts
   - after: we store notifications as they come from the server (with preprocessing) but without bundling and without texts rendered. 
   - Reasons:
      - `NotificationReader` mark notifications as per notification NOT per bundle. If we don't have original notifications it becomes complicated to properly updated state.
      - We wanted to have ALL notifications in the store, not only notifications which are visible. The new implementation separates notifications storing and notifications displaying. Now we store all the notifications in the store, while displaying only the notifications which has rules of how to render their texts. 
- added missed `goTo` for some notifications


maxceem:
In general, NotificationReader markes notifications as read when it's rendered i. e. visible for user. 

In particular for project plan page, when user opens it, notifications related to project plan and phase cards are marked as read, but not for timelines as timelines are not visible initially. Only when user open timeline tab related notifications will be marked as read. 

Note, that timeline tab suppose to have red cue shown if there are some unread notifications related to timeline, so if we mark such notifications before user openes timeline, he will never see a red cue.

RishiRajSahu:
>Note, that timeline tab suppose to have red cue shown if there are some unread notifications related to timeline, so if we mark such notifications before user opens timeline, he will never see a red cue.

Correct we should not mark them as read on load of project-plan. I am just thinking how the cue will be handled/behave when we open timeline, will it disappear immediately, anyway we can take care that in our other issue

>In particular for project plan page, when user opens it, notifications related to project plan and phase cards are marked as read

Actually why i raised the first review comment is, I am curious how we are going to handle visual cues for phases due to notifications regarding phase-cards ?
screen shot 2018-10-31 at 2 05 38 pm

Cue on phase-card will be result of both the phase-card and timeline notification

maxceem:
> I am just thinking how the cue will be handled/behave when we open timeline, will it disappear immediately, anyway we can take care that in our other issue

So far cue will disappear NOT immediately, but after server request is finished. But it will be fixed after the issue [#2649 - Hide dismissed notifications without waiting for server response](https://github.com/appirio-tech/connect-app/issues/2649) is done. During that issue I'm going to slightly refactor actions, so we will get this behavior for all the cases. I mean everywhere notifications will be marked as read with immediate effect in Connect App. And with reverting `read` state if an error happened during server request.

> Actually why i raised the first review comment is, I am curious how we are going to handle visual cues for phases due to notifications regarding phase-cards ?

So far per current implementation it should work as expected, i. e.:
- notifications regarding phase card will be marked as read on page render - no red cue for such cases
- red cue for a phase will be shown if any of tabs have unread notifications: `timeline`, `discussions` or `specification` this will be completed in the [issue #2455 - Add notification visual cues to phase timeline and specifications tabs](https://github.com/appirio-tech/connect-app/issues/2455). And it already works for `discussions`.

RishiRajSahu:
Okay sounds good to me, could we create a separate issue to keep track of phase specific notifications ?
>notifications regarding phase card will be marked as read on page render - no red cue for such cases

maxceem:
> Okay sounds good to me, could we create a separate issue to keep track of phase specific notifications ?

Sure, what is the issue there?

RishiRajSahu:
>red cue for a phase will be shown if any of tabs have unread notifications: timeline, discussions or specification


```
EVENT_TYPE.PROJECT_PLAN.PHASE_PAYMENT_UPDATED 
EVENT_TYPE.PROJECT_PLAN.PHASE_PROGRESS_UPDATED 
```

should we include the above notifications as well to show red-cue or not ?

maxceem:
> should we include the above notifications as well to show red-cue or not ?

I guess this is a common question for many places.

maxceem:
> should we include the above notifications as well to show red-cue or not ?

So far there were not such requirements in the related issues:
- [issue #2455 - Add notification visual cues to phase timeline and specifications tabs](https://github.com/appirio-tech/connect-app/issues/2455)
- [issue #2410 - Add notification visual cues to project plan cards](https://github.com/appirio-tech/connect-app/issues/2401)

Note, that per issue #2401 only notifications inside tabs affect red cue on the phase card. As per requirements, when we open phase, red cue on the phase disappears and is shown on the particular tab instead. Which implies that phase card by itself doesn't have a red cue.

> should we include the above notifications as well to show red-cue or not ?

Coming back to your question. I understand your intention to show red cue for a phase due to phase related notifications. But I don't have a strong opinon on whether we should show it here or no.
I think this can be general question of how we higlight places which were updated but user actually see them immediately when he comes to the page.

RishiRajSahu:
@maxceem thanks, I have created a issue to discuss further on this https://github.com/appirio-tech/connect-app/issues/2660

PR #2726, opened 2018-12-20
[HotFix]CS-Enabled file upload only for Managers

RishiRajSahu:
CS forms specific change to enable only managers to upload files and that too from specifications tab under phase.


RishiRajSahu:
Dropping this PR

PR #2734, opened 2019-01-03
team management feature

gondzo:


src/projects/list/components/Projects/ProjectsGridView.jsx
```diff
@@ -135,6 +135,22 @@ const ProjectsGridView = props => {
         // </div>
         // Hiding the user segment for the momemnt
       }
+    },
```
vikasrohit: I don't think we need this button for all users in all cases. I mean, this column was supposed to fill the gap created by complicity in implementing the use case where an unregistered user is accepting an invite. So, if we keep this column for that case, I guess we should show this column only when there exists at least project where there is a pending invite for the logged in user. Further, we should directly add the user to the project on click of `Join` button, right now it is simply redirecting to the details page which user can go by clicking on project name any way. Also, to have all these things working, in our REST api we need to allow listing and get operations on project resource  on projects where requesting user is not part of the team yet but has an invite. So, in nutshell, I think this button right now is not of much use and unnecessarily consuming the space on the viewport.

What I would suggest rather is:
1. In our REST api we need to allow listing and get operations on project resource  on projects where requesting user is not part of the team yet but has an invite
2. In our loadProjectDashboard action, add the logged in user to the team by making a REST call, when the project does not have the user in members list but in invites list.
And I guess that would do the trick.

gondzo: looks like I didn't commit the changes before creating the PR, please check the updated PR
We're showing this button only when there is a pending invite (and on click showing the join dialog) - it's the same as if user would click on the project title, just an additional indicator.

I wouldn't allow project get operations for invited users as that would allow those users to view/post messages. 
The tricky part is to not show the error page for users that can't view the project (show the join dialog instead), show the join dialog for members that can see the project anyway (managers/admins) but also show the error page when there is no invite. This is handled for all /projects/:id routes in https://github.com/appirio-tech/connect-app/pull/2734/files#diff-2b7e6f55e2297d58b8bd2191b8d19733R152

So even if the user goes to /projects/:id/plan, the invite dialog would be shown.

vikasrohit: Okay. So, essentially we can remove the new column if we want as loading the project details page would automatically show them the join dialog and we can send the link for joining the project in the email notification. It is just a quick way to join a project where user has pending invite. Reason I am asking this is because personally it seems we are consuming space on listing page for a not so frequent use case.
Am I right?

vikasrohit: @gondzo @RishiRajSahu resolving the conversation for now. We need to think about keeping the Join button in the list view, after testing the feature in dev env.

PR #2755, opened 2019-01-17
Corrected error message for server error during invites

RishiRajSahu:
fyi @vikasrohit


RishiRajSahu:
Closing this PR as the scope of this change has become broader as per discussion here https://github.com/appirio-tech/connect-app/issues/2758. Will push it as a new PR once req. are done.

Issue #2555, opened 2018-09-28
[$30] When we edit milestone but didn't make any changes, "Update milestone" button has to be disabled.


maxceem:
### Expected behavior
When we edit a milestone but didn't make any changes, `Update milestone` button has to be disabled.

### Actual behavior
Now when we just click edit button and open edit form, we already can click `Update milestone` even without making any changes.

![image](https://user-images.githubusercontent.com/146016/46189947-ebac2800-c323-11e8-93e8-5cd8077561c2.png)

@RishiRajSahu if it's a desired fix, I think we can include it to next CF.

afrisalyp:
Hi @maxceem, Can you please assign this ticket to me?
Thanks.

maxceem:
Done

afrisalyp:
Hi maxceem,
It’s done and ready for review
Here is the PR https://github.com/appirio-tech/connect-app/pull/2762

Please review. Thanks.

maxceem:
Hi @afrisalyp I've checked you PR, please have a look https://github.com/appirio-tech/connect-app/pull/2762#pullrequestreview-195870471

afrisalyp:
@maxceem Thanks for the review, I'll check the comments.

RishiRajSahu:
Working fine on dev.

PR #2762, opened 2019-01-22
Disable update milestone button if no changes.

afrisalyp:
https://github.com/appirio-tech/connect-app/issues/2555

### DIFF
--- a/src/projects/detail/components/timeline/Milestone/Milestone.jsx
+++ b/src/projects/detail/components/timeline/Milestone/Milestone.jsx
@@ -45,12 +45,14 @@ class Milestone extends React.Component {
     this.completeFinalFixesMilestone = this.completeFinalFixesMilestone.bind(this)
     this.extendMilestone = this.extendMilestone.bind(this)
     this.submitFinalFixesRequest = this.submitFinalFixesRequest.bind(this)
+    this.milestoneEditorChanged = this.milestoneEditorChanged.bind(this)
 
     this.state = {
       activeMenu: '',
       isHoverHeader: false,
       isEditing: false,
-      isMobileEditing: false
+      isMobileEditing: false,
+      disableSubmit: true
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/projects/detail/components/timeline/Milestone/Milestone.jsx
```diff
@@ -103,6 +105,26 @@ class Milestone extends React.Component {
...
+    for (const key in values) {
+      if (values.hasOwnProperty(key)) {
+        const element = values[key]
+        let compareElement = this.props.milestone[key]
+        if (!(compareElement instanceof String)) {
+          compareElement = compareElement.toString()
+        }
+        
+        if (element !== compareElement) {
+          return this.setState({ disableSubmit: false })
```
maxceem: Here the state is updated in a cycle `for` multiple times. Do you think it can affect the performance as we call this method on each keypress?
It feels that it would be safer for performance to update state only once after cycle if we find in the cycle that some of the values have been updated.

afrisalyp:
@maxceem,
I'm working on the issue.
And I've also replied the comment on the line.

src/projects/detail/components/timeline/Milestone/Milestone.jsx
maxceem:
```
if (element !== compareElement && this.state.disableSubmit) {
```
This one wouldn't work, as `setState` doesn't change `this.state` immediately.

The best way would be to add a new variable, before the cycle, something like this:
```
let disableSubmit = true
for (...) {
   if (element !== compareElement) {
      disableSubmit = false
      break // stop cycle but don't stop fuction so we can apply the result
   }
}
// apply the result
this.setState({ disableSubmit })
```

This code is just an idea, not exact.

maxceem: I see you've fixed it another way, I'm checking...

maxceem: your approach also fixes a possible performance issue

Issue #2623, opened 2018-10-12
[$15] Failed to mark notifications seen


maxceem:
From time to time there is error popup `Failed to mark notifications seen` when I navigate to the project dashboard.

![image](https://user-images.githubusercontent.com/146016/46844144-65561280-ce08-11e8-9501-9f4630f742cb.png)

RishiRajSahu:
@maxceem as per me this appears when we try to mark already seen notifications as seen again. I tried reproducing it by opening the notifications dropdown in different browsers then trying to mark the notification as seen in each. For later it gave the same error as notification was marked as seen already.

maxceem:
@RishiRajSahu as the user never calls the method to mark notifications as `seen` by himself, it's always done in the background automatically.

I think we can just remove showing this message in Connect App for this error. The only thing we would still have an error message in the browser console.

What do you think?

maxceem:
As no PR was provided for 12 hours, this issue is now open for pick up.
The one who first comments can start working immediately without waiting to be assigned.

maxceem:
Sure, assigned to keep it clear.

gets0ul:
@maxceem 
Is there any specific step to do to trigger the popup?

maxceem:
Yes, you will need two users, you can find credentials on the challenge forum.

- Use any user (but not `pshah_manager`) to create a new project, copy project URL and close the tab (important)
- Now use user `pshah_manger` from the forum and login in **another** browser 
- Using user `pshah_manger` open the project you've just created using link you copied before, go to the `Project Plan` tab, open any phase, go to the `Discussions` tab of the Phase and post a new comment. (you may need to reload page and make sure post is created, looks like there is some issue with creating posts)
- Copy URL of the current page, it will end with `/plan`
- Now using the first user in the first browser, open TWO tabs using the link which just copied end with `/plan`
- In on of the tabs, make sure you can see notification about a new post, open dropdown with notification in the top right corner, you should see "**Part1 Manag1**  responded to a post"
- In this tab click **Dashboard** tab and wait a few seconds, in the background this notification will be marked as seen when we open **Dashboard**
- Now switch to the second tab where you open link which ends with `/plan`. Here switch to the **Dashboard** tab too. It will try to mark the same notification as `seen` and background and you will see aleat popup with error

See the demo https://monosnap.com/file/xEH4yeZ2g3ExS9eQFN7EUfuv45sZem

[...]

PR #2774, opened 2019-01-23
Fix for issue 2623

gets0ul:
#2623 
Remove showing this message in Connect App, show only in browser console


gets0ul:
Done. Commit added for lint error fix.

PR #2796, opened 2019-01-31
Add organization special page

phoenix303:
Add organization special page support for absolute or related URL.


maxceem:
@phoenix303 just an idea from the implementation point of view. To avoid interfering with other existent features, we can navigate to the special page instead of `/new-project` when we are trying to navigate to the `/new-project` page but without any subroutes.

maxceem:
> Also, something to note when we load https://connect.topcoder-dev.com/new-project/generic-design directly and click back it goes to /new-project, only when we land on generic-page navigating from all-websites and click Back we end up on the page https://connect.topcoder-dev.com/new-project/all-websites.

Good catch didn't notice that before. So this behavier can stay. Only if navigated to https://connect.topcoder-dev.com/new-project/all-websites from the special page and after go to https://connect.topcoder-dev.com/new-project/generic-design when cliking `Back` we go to https://connect.topcoder-dev.com/new-project/all-websites. 
But if directly naivgated to https://connect.topcoder-dev.com/new-project/generic-design from the spec page we can go back to the special page as it's done now.

maxceem:
3. One more small thing, could you please resolve confilct to the `dev` branch in file `src/projects/list/components/Projects/ProjectsGridView.jsx`.

maxceem:
Btw, did you have a chance to check if this change is required or no https://github.com/appirio-tech/connect-app/pull/2795/files#diff-99a91bfabeced5de0d595f019dd0bc9bR36? In case it's not required can we keep it as it was before `return loadUserSuccess(dispatch, token)`?

phoenix303:
Oh yes, it is not required. We can keep it as before.

phoenix303:
>
> Just one thing. If from this page http://local.topcoder-dev.com:3000/organization/new-project/ I'm clicking **Link to all websites** then end up on `404` page. While if I navigate to the same page using an absolute link. It works good. See demo https://monosnap.com/file/ihbN9be3wbIJiijJUNJH8TQ5fwvfAg
> 

I looked into it, the problem is we are navigating to /all-websites from /organization/new-project and <ProjectWizard> loads project metadata template on the first load, which in this case is "/all-websites"
and unlike "/app" and "/generic-design", all-websites is not a template and therefore replaces URL with /404 not found.

One easy fix is to include "all-website" in template.
But all-website is not a template so we can add new 'subType' as a parameter which has all-website value.

Please let me know your thoughts.

vikasrohit:
Sounds good. How do you configure relative vs absolute URLs or are we detecting it in code by parsing the URL value?

maxceem:
Right, detected by parsing the URL value.

vikasrohit:
I think I found that logic https://github.com/appirio-tech/connect-app/pull/2796/files#diff-948050426bae50b1ed6c5fdb15c8697bR314. It seems URLs starting with https are treated as absolute URL, I think we need to make it support `http` as well. By the way we didn't need support for relative or absolute URL as of now as whenever, we are specifying specific project catalog URL for an org, it supposed to be an external URL. But I think it won't hurt us.

maxceem:
Ok, I was thinking the page could be also a part of Connect. It supports both `http` and `https`, as there is a question mark in the regex `https?` which makes `s` optional.

vikasrohit:
Oh, got it. I miss interpreted the regex.

Issue #2775, opened 2019-01-23
Only allow users with role `Connect Copilot Manager` to assign copilot to Connect project


RishiRajSahu:
We have created a new role for connect user named as `Connect Copilot Manager` , only users assigned with this role are allowed to assign copilot to the project going forward.

changes needs to be done to have the above said functionality.

fyi @vikasrohit @acshields

RishiRajSahu:
To handle this, for non `Connect Copilot Manager` users -
1.We won't be showing the copilot option in dropdown
2.We won't be allowing changing role of added member to copilot.

fyi - @maxceem

maxceem:
@RishiRajSahu do I understand it right, that if current user doesn't have role `Connect Copilot Manager`
we should remove these two options from the **Topcoder Team** dialog:

![image](https://user-images.githubusercontent.com/146016/51598669-fde84c80-1f38-11e9-9b10-d38012ff1257.png)

acshields:
Hi @maxceem @RishiRajSahu -  

In my ideal workflow, I would still like for a Copilot to see the option to invite another to a project - However, if they did so, then a message could appear saying "Your request has been sent to the Connect Copilot Manager for review. If the Connect Copilot Manager approves the assignment Copilot X will be invited to the project." and then had an email/notification triggered to the Connect Copilot  Manager(s) saying that Copilot X has requested that Copilot Y be added to the project. If the Connect Copilot Manager approves the assignment, then an invitation to the project should be sent to the new Copilot. Further, if a Connect Copilot Manager invites a Copilot to a project directly then there should be no further approval process - the invitation should be sent immediately.

Additionally, we would not want this new role (Connect Copilot Manager) to be displayed as an option to invite a team member in the "Invite Team Members" section of Connect. Since this role is intended for existing admins, think of it as an admin role with the special capability of assigning Copilots to a project.

RishiRajSahu:
@acshields 
>Connect Copilot Manager invites a Copilot to a project directly then there should be no further approval process

Only this is achievable as of now,

Regarding the other flow you mentioned is  tricky and a new feature which needs quite some time we can discuss more about this in our next call.

Also, `Connect Copilot Manager` - Yes this role won't displayed as an option as this is admin kind of role with special capabilities.

maxceem:
@vikasrohit I guess this is already in DEV branch.

vikasrohit:
Yes, it is.

PR #2798, opened 2019-01-31
allow only to Connect Copilot Managers invite users as copilot or change role to copilots

maxceem:
fix for issue https://github.com/appirio-tech/connect-app/issues/2775

Disable copilot role in dropdown and switcher if user doesn't have role `Connect Copilot Manager`:

image

Even administrators cannot assign copilot role now if they don't have role `Connect Copilot Manager`.

Note, I've added a tooltip with an explanation why items are disabled, see video https://monosnap.com/file/m1SC9BiiYEtNNuaBSeAuPFQBjw1sDU

There is a visual issue when the tooltip is shown some space is being added and the window "jumps". This issue can only be reproduced in Chrome and not in FireFox. I've spent some time trying to resolve it. But there is no real reason for such issue, looks like some issue of the browser.

I can either remove tooltip or run F2F challenge to find out a way to get rid of this issue. Or we can leave it for some time as we are going to change this behaviour by adding approval process.

fyi @vikasrohit


maxceem:
> Thats good. But it seems the video link is not correct.

Fixed the link to the demo video of the tooltip and the issue in Chrome https://monosnap.com/file/m1SC9BiiYEtNNuaBSeAuPFQBjw1sDU

> Could we quickly allow administrators and Connect Admin to be able to add copilots as well? 

Done.

PR #2806, opened 2019-02-08
Feature/attachment permissions

gondzo:


vikasrohit:
LGTM... Merging to test it out in dev. We can issues fixed as tasks, if found any.

Issue #2805, opened 2019-02-07
Don't allow to activate project if there is no 'planned` phase


maxceem:
From `Connect - UI/UX Enhancements` ducument.

#### Overview
Project plan must have at least one phase in planned status prior to project becoming active
In order to increase adoption of project plans, a project should not be able to be moved into active status until there is at least one phase in planned status.

#### Implementation
- When have any project which is no in `Active` state now.
- When we use a user which has rights to switch project to the active state like `pshah_manger`
- We open dropdown to change state

    ![image](https://user-images.githubusercontent.com/146016/52392270-049bc580-2adc-11e9-99dd-ba40ff372add.png)

- If there is no phase in `Planned` state

   ![image](https://user-images.githubusercontent.com/146016/52392298-21d09400-2adc-11e9-8c48-54afac798bd9.png)

- The option `Active` in the propdown should be disabled. And when we put mouse on it there should be tooltip shown: `To activate project there should be at least one phase in "Planned" status. Please, check "Project Plan" tab.`.

   The hidden option and tooltip should work same as for phase status dropdown:
   
    ![image](https://user-images.githubusercontent.com/146016/52392439-c9e65d00-2adc-11e9-9001-40d7e24ac807.png)

RishiRajSahu:
Done.

PR #2809, opened 2019-02-10
Feature/project activation improvement

maxceem:
winning submission for the challenge 30083315 - Topcoder Connect - Project activating validation, fixes issue #2805

+ also, fixed to also allow activate project if it has active phases, which we missed in the issue specification, but I think it should be also allowed like that


vikasrohit:
LGTM

Issue #2735, opened 2019-01-07
[Team Management 2.0] Do we need auto complete for adding existing topcoder users


vikasrohit:
Currently after new team management section, we are missing the auto complete feature for searching existing topcoder user while adding a user to the team. It might be an intentional feature, however, creating the issue to track the discussion.

fyi @acshields @vic-topcoder @RishiRajSahu

vikasrohit:
@acshields as per our last discussion, we need this feature again so we are including this our bug bash.

vikasrohit:
@maxceem one important thing to keep in mind for this feature is that we should be able to invite the member by handle even if there is no result returned for that by the api.

@acshields @vic-topcoder could you please confirm that this is an expected behaviour?

maxceem:
To sum up:
- this featuer should work in both dialogs **Project Team** and **Topcoder Team**. We can see them by using user `pshah_manager` and opeining this project https://connect.topcoder-dev.com/projects/7839 by clicking the **Manage** link on the bottom of the left sidebar.
- we previously had autocomplete working in the previous implementation of team management feature. To check it out can use commit `a9b60bc8f2fd845d3eff82b860033f192e824ba3`. See demo https://monosnap.com/file/zmx4r3Gp4ENHgrkmG59SJmqJOXZlDh
- now we should start typing user handle with `@` to make autocomplete work. As we also can type email which don't start from `@` and they shouldn't have autocomlete
- even if there are no results returned by autocomplete we should be able to type handle ourself and invite user (unlike in previous implementention) 
- we should support navigating by keyboard (didn't have before):
   - press `down`/`up` should select next/previous suggested handle
   - pressing `Enter` should complete the typing handle with the currently selected handle
- note, that we can type several handles in the field and in **Project team** dialog we can type both handles and email.

vikasrohit:
> * now we should start typing user handle with `@` to make autocomplete work. As we also can type email which don't start from `@` and they shouldn't have autocomlete

@maxceem we have a new requirement from the team that `@` is causing confusion to the product users, so we need to get rid of it. So, could we also get rid of `@` now along with this task? I know the side effect would be that it would make server calls even when user is typing emails, but I guess that does not harm, user might find existing user for that email in this case.

maxceem:
@vikasrohit got it. Should we let users enter handles both ways with or without `@`? `@handle` and `handle`?

vikasrohit:
[...]

PR #2829, opened 2019-02-13
Feature/team management autocomplete

maxceem:
This is a winning submission from challenge 30083313 - Topcoder Connect - Team Management - Autocomplete

This fixes issue https://github.com/appirio-tech/connect-app/issues/2735

Note, in the implementation I've asked to reuse the library we already use `react-select`, which is also used for autocomplete in attachment permissions dialog.

We still need to test it, and probably think about some UX improvements.
Also, for now we created a separate component from suggestion for attachments permissions.
Maybe we can/should merge them and use the same CSS style for them.

We also use a very old version of `react-select@0.9.1` while the current version is `2.3.0`. This includes the fact that this old version causes some warnings in the browser console. So we consider updating it.


src/components/TeamManagement/AutocompleteInput.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const asyncOptions = (input) => {
+    const value =  typeof input === 'string' ? input : ''
+    const createOption = {
+      handle: value,
+      // Decide if it's email
+      isEmail: (/(.+)@(.+){2,}\.(.+){2,}/).test(value),
+    }
+    if (value.length >= AUTOCOMPLETE_TRIGGER_LENGTH) {
+      return loadMemberSuggestions(value).then(r => {
```
RishiRajSahu: @maxceem are we directly making api call here instead of going through action-reducer path ?

maxceem: @RishiRajSahu Yeah, I've noticed it. IMHO as we don't need this data in redux as it's very temporary and calling it directly without going through the whole flow keeps the speed fast for this place. As going through the action-redux flow may slightly slow it down.

But I think we should add the comment to the code that this place is an exception due to the reason above.

Or if you think it would be better to keep in uniform everywhere, we can do it with actions/reducers instead.

Though I think, first we have to confirm if the implemented workflow is good for us, and make any enhancements we need.

RishiRajSahu: @maxceem we need to action/reducer pattern intact and I don't think it will be lot slower that way. Also shouldn't we be using debounce/throttle here ?

maxceem: @RishiRajSahu  Regarding using action/reducer I've created a separate issue, as we have to update the `react-select` version first, before we can move to action/reducer approach https://github.com/appirio-tech/connect-app/issues/2846

> Also shouldn't we be using debounce/throttle here ?

debouncing is implemented

RishiRajSahu: LGTM

PR #2830, opened 2019-02-13
Feature/account settings improvements

maxceem:
PR with enabled changing email


vikasrohit:
@RishiRajSahu we might need to check with product team if they want the email editing enabled yet or not.

RishiRajSahu:
Sure, we can revert this change if not agreed by product team.

Issue #2815, opened 2019-02-12
Optimize the number of calls being made for loading a Connect project dashboard


vikasrohit:
As of now, we are making too many calls to fetch the data of the project on the dashboard and the reason is that we need to show some portion of every possible data attached with a project, on the dashboard. Along the feature development across releases, we introduced new API calls to fulfil certain requirements. However, it seems that we are now making some extra calls to the APIs which we might be able to reduce still keeping all the features working.
Purpose of this tasks is to identify all such places where we can improve the page load performance, either by directly optimizing the code of the front end or identify the API changes which can significantly improve the performance.

fyi @maxceem @RishiRajSahu

vikasrohit:
@maxceem One improvement we can do here is to don't make separate calls to loadMembers action for every phase topic load, instead of that we can make a combined call to `loadMembers` after all phase topics are loaded successfully. Do you think that would create any issue in terms of rendering of the phase topics?

maxceem:
@vikasrohit 
I guess it should be ok. We will need to make sure that until members are loaded we can still display topics/posts in the phases with some placeholder info for users.

vikasrohit:
@maxceem I tried something here https://github.com/appirio-tech/connect-app/commit/a0d137b9dd2eaf7d7d94d383843f91de273b50e9, in case you want to take a look tomorrow morning.

vikasrohit:
@maxceem Another possible optimization would be to get phase topics in single call by filtering at `referenceId`. I think we did have a rationale for fetching them individually but I am not able to recall. So, it might be a good time to refresh that.

maxceem:
@vikasrohit I guess we can do this optimization together with implementing a new related feature: https://github.com/appirio-tech/connect-app/issues/2835

vikasrohit:
Yes, absolutely.

maxceem:
@vikasrohit  At the moment posts are not shown at the Phases discussion tabs on DEV, see https://connect.topcoder-dev.com/projects/7489/plan:

![image](https://user-images.githubusercontent.com/146016/54655205-0190ed80-4afc-11e9-9de6-d14e1422e113.png)

Could it be connected with this PR? https://github.com/appirio-tech/connect-app/pull/2840

vikasrohit:
Fully probable. Good catch. Do you think you would be able to look at it?

vikasrohit:
[...]

PR #2840, opened 2019-02-15
Feature/reduce number of member search calls

vikasrohit:
One efforts towards resolving #2815 

fyi @maxceem @RishiRajSahu, originally I thought of keeping it as hotfix because I was seeing very slow response on the production, however, it turned out that caching was not working for cloudfront on production and these extra calls for _search were not introduced in our last release 2.4.9. So, merging the changes as regular feature, so that we can test it out more in dev env.
In case, we see the need of this on production before our next release, we will cherry pick the changes to the master hotfix.


vikasrohit:
@maxceem @RishiRajSahu merging this PR now so that we can test if the optimization has introduced any issue.

PR #2843, opened 2019-02-19
Account manager changes

gondzo:


src/config/constants.js
```diff
@@ -540,6 +540,7 @@ export const PROJECTS_LIST_VIEW = {
  */
 export const PROJECT_ROLE_COPILOT = 'copilot'
 export const PROJECT_ROLE_MANAGER = 'manager'
+export const PROJECT_ACCOUNT_MANAGER = 'account_manager'
```
RishiRajSahu: @gondzo Could we rename it to `PROJECT_ROLE_ACCOUNT_MANAGER` instead ?

gondzo:
yes, these permissions here are just a demo for the centralized permissions management feature. we can now expand this to all permission checks in the app

src/config/constants.js
gondzo: done

PR #2861, opened 2019-02-26
Gauravgautam

gautam1168:
Added test cases for the operators that were not tested already.

There are 32 failing test cases and 81 passing ones. The failing test cases should pass if the evaluate function is supposed to work like js engine in chrome.

The lint shows 12 errors but those are because of the `Throw` function of `chai`. I cannot change that.


gautam1168:
Ah I see. I was not aware of the lower case throw.

PR #2866, opened 2019-02-26
issue 2836 fix

Gauravseta:


Gauravseta:
@maxceem can you please create a new discussion and check inside it? till then i confirm you on already created discussions.

maxceem:
@Gauravseta you are right. It works great when creating a new discussion.

So we need to make it also work when editing discussions on the dashboard page http://local.topcoder-dev.com:3000/projects/7885

image

Or inside phase http://local.topcoder-dev.com:3000/projects/7885/plan

image

PR #2871, opened 2019-02-28
fixes isssue 2828

sumitdaga:


sumitdaga:
@maxceem  
if i remove this, i can see console errors ...
 <pre>if (!props.projectTemplate) {
   return null
}</pre>

strange you don't see it ...not sure!

PR #2888, opened 2019-03-08
Adding Function: Replace prepared conditions inside expressions.

kotamanegi:
Hello. My topcoder handle name is "kotamanegi".
I'm here for development challenge at https://www.topcoder.com/challenges/30085656
I've added function,and fixed test.
The reason why I have changed the test is that the testcase says inserting Unnecessary "()" (not written in the input) to the result expression is necessary.
if you have any questions,please feel free to contact me.

### DIFF
--- a/src/helpers/dependentQuestionsHelper.js
+++ b/src/helpers/dependentQuestionsHelper.js
@@ -274,6 +274,33 @@ export function getFieldNamesFromExpression(expression) {
  * @returns {String} expression
  */
 export function populatePreparedConditions(expression, preparedConditions) {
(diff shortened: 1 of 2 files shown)


maxceem:
Hi @kotamanegi.

It's ok to fix unit tests as you did, the brackets are not needed there.
I would recommend you remove this PR and remove the branch from the repository online. As now everybody can see your solution before the challenge is finished.

Instead, you have to submit a git patch using button Submit on the challenge page where now you submitted a text file.

kotamanegi:
OK,Thank you very much for your advice!

Issue #2881, opened 2019-03-07
[Dependant questions] - support preparedConditions


maxceem:
From @vikasrohit:

If you notice the code for project estimation, https://github.com/appirio-tech/connect-app/blob/dev/src/config/projectWizard/index.js#L519, you would see that we are using a new config called `preparedConditions` which are kind of variables for various conditions and we are using these variables in actual complex conditions. So, I would like to use the same variables when we are evaluating the conditions for show/hide of sections/subsections/questions etc. This would have two benefits:
1. It would evaluate all conditions only once per re-render
2. and more importantly, it would keep all conditions in single place just like we do it in our code

maxceem:
@vikasrohit here is how the current algorithm works:

![conditions dependency](https://user-images.githubusercontent.com/146016/53925930-a226ec80-40bc-11e9-80a2-1403aea1c765.png)

### Current alrogithm:
When any field in the form is changed (means form `data` is changed).

- 1\. Calculate condition `1`. If the value is changed then we update `data`.
- 2\. Calcualte condition `2` using updated `data`. If the value is changed then we update `data.`
- ...
- 10\. Calculate condition `10` using updated `data`.  If the value is changed then we update `data`.

#### Notes
- Condtions may only depend on the condtitions `above` inside the tree.

### New alrogithm:
If we introduce prepared conditions, the algrithm would work like this.
When any field in the form is changed (means form `data` is changed).

- 1\. Check condition `1`
   - if it doesn't have `preparedConditions` inside expression, calculate it. 
   - if it has `preparedConditions` inside expression, **REcalculate** `preparedConditions` first and after we calculate the condition.

   after, if the value is changed we update `data`.
- 2\. Check condition `2`, now it uses updated `data`.
   - if it doesn't have `preparedConditions` inside expression, calculate it using updated `data`.
   - if it has `preparedConditions` inside expression, **REcalculate** `preparedConditions` first using updated `data` and after we calculate the condition using updated `data`.

   after, if the value is changed we update `data`.
- ...
- 10\. ...
- 11\. After conditions for all nodes are calculated we should recalculate all `preparedConditions` with the updated `data` as not all `preparedConditions` was calculated during steps above. As an alternative we can recalculate only `preparedConditions` which hasn't been recalculated during steps above.

#### Notes
- Condtions may only depend on the condtitions `above` inside the tree. `preparedConditions` shouldn't add dependencies which doesn't follow this rule.
[...]

PR #2891, opened 2019-03-09
Feature/form redesign prepared conditions

maxceem:
fix issue #2881

### DIFF
--- a/src/helpers/dependentQuestionsHelper.js
+++ b/src/helpers/dependentQuestionsHelper.js
@@ -254,3 +254,26 @@ export function getFieldNamesFromExpression(expression) {
 
   return fieldNames
 }
+
+/**
+ * Replace prepared conditions inside expression
+ *
+ * @example
+ *   const expression = 'preparedCondition1 == 1'
+ *   const preparedConditions = {
+ *     preparedCondition1: '2 - 1'
+ *   }
+ *
+ *   const output = populatePreparedConditions(expression, preparedConditions)
+ *
+ *   // output => '(2 - 1) == 1'
+ *
+ * @param {String} expression         expression
+ * @param {Object} preparedConditions prepared conditions
+ *
+ * @returns {String} expression
+ */
+export function populatePreparedConditions(expression, preparedConditions) {
+  !preparedConditions // this is to avoid lint error - have to write implementation instead
+  return expression
+}
--- a/src/helpers/dependentQuestionsHelper.js
+++ b/src/helpers/dependentQuestionsHelper.js
@@ -274,6 +274,11 @@ export function getFieldNamesFromExpression(expression) {
  * @returns {String} expression
  */
 export function populatePreparedConditions(expression, preparedConditions) {
-  !preparedConditions // this is to avoid lint error - have to write implementation instead
+  // we assume here that the preparedConsition could only have spaces around it, only then it can be replaced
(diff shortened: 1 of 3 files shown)


src/helpers/dependentQuestionsHelper.js
```diff
@@ -254,3 +256,35 @@ export function getFieldNamesFromExpression(expression) {
...
+ *
+ * @returns {String} expression
+ */
+export function populatePreparedConditions(expression, preparedConditions) {
+  // in the Regexp we describe situations when preparedCondition can be replaced,
+  // instead of defining situations when it cannot be replaced
+  const allowedBefore = ['^', '\\s', '\\(', '!'].join('|')
+  const allowedAfter = ['$', '\\s', '\\)'].join('|')
+
+  preparedConditions && _.forEach(preparedConditions, (value, key) => {
```
vikasrohit: @maxceem Just curious, what issues we can have if we do a simple replace of variables like https://github.com/appirio-tech/connect-app/blob/dev/src/config/projectWizard/index.js#L504 instead of handling it like the proposed way? I know it would be better, but just wanted to list down what are the benefits. One of the negative of my approach in https://github.com/appirio-tech/connect-app/blob/dev/src/config/projectWizard/index.js#L504 is that it fails to replace variables correctly when two variables are in way where one variable is a substring of another. Does this new approach would overcome that?

vikasrohit:
LGTM @maxceem except one comment about the way we are replacing the conditions. Also, unit tests are also failing, so we need to fix the unit tests before merging the PR.

src/helpers/dependentQuestionsHelper.js
maxceem: @vikasrohit 
> Does this new approach would overcome that?

yes, this approach overcomes that and few other edge cases.

Let's say we have prepared condition variable: `PREPARED_CONDITION_1` then in all next cases we should NOT replace it:
- `nestedObject.PREPARED_CONDITION_1` - part of object 1st level
- `nestedObject.propertyWithObject.PREPARED_CONDITION_1` - part of object, deeper levels
- `SUFFIX_PREPARED_CONDITION_2` and  `PREPARED_CONDITION_2_POSTFIX` - as part of some other longer variables
- `someArray contains 'PREPARED_CONDITION_1'` - as a string literal

maxceem:
@vikasrohit unit tests are fixed.

src/helpers/dependentQuestionsHelper.js
vikasrohit: Okay, however, I think having field names of the template object to be same as of variables is very very rare event. But still, the benefit of avoiding replacement of partial string in longer variable, is good enough to adopt this approach. Now, I am curious if we should use the same logic at the place I shared [above](https://github.com/appirio-tech/connect-app/blob/dev/src/config/projectWizard/index.js#L504).

maxceem: I think it makes sense to use the same approach there. Also, I think we can replace the prepared conditions once as it's done for the templates. As it looks currently this preprocessing is done each time we recalculate the price.

vikasrohit: The difference is that [here](https://github.com/appirio-tech/connect-app/blob/dev/src/config/projectWizard/index.js#L504) we are not just replacing the variables with conditions, we are evaluating them as well upfront. So, we have evaluate them every time there is a change in the data.

Issue #2933, opened 2019-03-27
[$90] [Team Management] - Use semicolon and space to select the email/handle


vikasrohit:
### Expected behavior
As soon as user enters `;` or hits SPACE key after entering email, while inviting a member, it should select the entered email as email to be invited.
We should apply the same behaviour when user is inviting via handle as well.

### Actual behavior
As of now, it only selects the entered email when user clicks on the `Create` link/button in the auto suggest dropdown.

### Steps to reproduce the problem
- Open any project
- Try to invite any member via email
- After typing the email, click outside the auto suggest dropdown and notice that it empties the text box where you entered the email. This behaviour is okay for now, however, if you enter `;` or hit SPACE before clicking outside, it still emptying the text box which is not the desired behaviour.

--

#### Environment
- OS: Any
- Browser (w/version): Any
- User role (client, copilot or manager): Any
- Account used: N/A

maxceem:
Sum up:
- To reproduce the issue can use a user `pshah_manager` and for example this project https://connect.topcoder-dev.com/projects/7959
- To open two possible dialog to invite user click link `Manage` on the left sidebar:
   
    ![image](https://user-images.githubusercontent.com/146016/55321984-978a2800-54ad-11e9-9907-eab220a0a00b.png)

gurmeetb:
@maxceem plz assign me not able to self assign

maxceem:
done

gurmeetb:
@maxceem I don't see any projects - am I missing something? I followed `Run locally for development` in README.

maxceem:
There is some issue with Project Service on DEV. Working to fix it.

Will updated as soon as it’s back, see https://apps.topcoder.com/forums/?module=Thread&threadID=934455&mc=2#2330163

gurmeetb:
@maxceem Do we also need similar behaviour for `Topcoder Team Management`?

gurmeetb:
@maxceem It’s done and ready for review.

RishiRajSahu:
Working fine in dev. Closing.

PR #2943, opened 2019-04-11
Fix for issue #2933

gurmeetb:


gurmeetb:
@maxceem fixed and also added support so users cannot enter first value as space or semi-colon.

gurmeetb:
@maxceem my bad, missed to run lint. Fixed it now.

PR #2944, opened 2019-04-11
issue 2935 fix

rashmi73:


maxceem:
@rashmi73 found one more thing.

Currently, filters don't make any effect for the customer if we click them. To make them work we have to remove this code:
https://github.com/appirio-tech/connect-app/blob/dev/src/projects/actions/loadProjects.js#L23-L42

Make sure to update this line https://github.com/appirio-tech/connect-app/blob/dev/src/projects/actions/loadProjects.js#L46 to use `criteria` instead of `requestCriteria` after code removal.

Also, have fix lint errors if there are any.

Thank you.

PR #2953, opened 2019-04-13
issue 2916 fix

rashmi73:


rashmi73:
@maxceem ok implementing it

maxceem:
@rashmi73 just in case would like to remind about this PR as the Bug Bash challenge in review phase now, which will be finished in 1.5 days.

Issue #2934, opened 2019-04-02
[$70] Newest message displays at the top of thread


ajefts:
Objective: Inverse the order that messages are displayed in a thread, so that the newest message is displayed at the top of the thread and then subsequent messages are displayed in descending order from newest to oldest. The "load earlier messages" button should be moved to the bottom of the thread and will be displayed after the five most-recent messages.

ajefts:
Link to feature: https://tcprod.productboard.com/feature-board/planning/features/1744834

vikasrohit:
@maxceem let me know if you have any query. I would like to include it in bug bash.

maxceem:
@vikasrohit 

1. where should we keep the field to enter a new post?
   
    ![image](https://user-images.githubusercontent.com/146016/55699787-0fe96f80-59ff-11e9-8122-3034b017b473.png)
    If we keep it at the bottom like it's done now it could be a little bit not comfortable as it would be far from the last message we are responding to.
   Not sure if it would also feel good if we move it up and would response to the latest post which would be under it.

2. Should we make the same change for Phases?
    
     ![image](https://user-images.githubusercontent.com/146016/55699891-6191fa00-59ff-11e9-919c-952afdb33281.png)

3. If we inverse the order, it looks like we would need to move this sticky date to the bottom:

    ![image](https://user-images.githubusercontent.com/146016/55699931-a74ec280-59ff-11e9-851d-324c6dbfde81.png)

4. This also should be changed in the full screen mode
    
     ![image](https://user-images.githubusercontent.com/146016/55699947-c4839100-59ff-11e9-9e95-175d6501c0c1.png)

5. Also, should be changed in mobile resolution: 
    Dashboard posts:
    ![image](https://user-images.githubusercontent.com/146016/55699984-fb59a700-59ff-11e9-8288-58a9f465b2cb.png)

    Dashboard posts (fullscreen):
    ![image](https://user-images.githubusercontent.com/146016/55700013-2fcd6300-5a00-11e9-8348-c4597dec88cc.png)

    Phase posts:
    ![image](https://user-images.githubusercontent.com/146016/55699998-10363a80-5a00-11e9-801b-5875ce4eb912.png)

RishiRajSahu:
@maxceem 
1. Let's keep the text area as it is for now. Still if you could quickly show a mock using browser tool it would be helpful
2. Yes for phases also
[...]

PR #2956, opened 2019-04-15
fix for issue #2934

gurmeetb:


gurmeetb:
@maxceem are we sure we want this change, I revisited the details and it was expected to leave it that way, Refer the snapshot attached in final change details from you -

maxceem:
@gurmeetb yes, we should make this change. This screenshot just shows the button which we have to take care about as we don't have a new design for this.

This button should work taking into account the new order and should be shown on the place where we "hide" old comments. In reverse order it's bottom.

gurmeetb:
@maxceem no worries, updated MR with the change requested.

maxceem:
Tested and works good.

PR #2957, opened 2019-04-16
fix for issue 2180

sumitdaga:


maxceem:
@sumitdaga just in case would like to remind about this PR as the Bug Bash challenge in review phase now, which will be finished in 1.5 days.

sumitdaga:
@maxceem  
should the orange color be present when editing title as well ?

maxceem:
@sumitdaga

> should the orange color be present when editing title as well ? (i think it should not)

It would be great if not complicated but is not required.

maxceem:
Ah, during editing it should be orange. During creating would be also good, but not required.

sumitdaga:
@maxceem  also should the lock icon be present during editing ?

maxceem:
@sumitdaga no, I think better to keep space for the fields. The color would be enough.

sumitdaga:
> During creating would be also good, but not required.  

you mean make it orange when the private switch is toggled to on ...it may look odd!

sumitdaga:
@maxceem  please take a look now!

maxceem:
> you mean make it orange when the private switch is toggled to on ...it may look odd!

yeah, it feels that it would make it more easy to track if it's private or not and reduce mistakes. But as it wasn't required initially, it can be skipped.

sumitdaga:
@maxceem  
i have updated background when creating post as well!

src/projects/detail/containers/FeedContainer.js
```diff
@@ -526,16 +535,24 @@ class FeedContainer extends React.Component {
...
 
 const mapStateToProps = ({ projectTopics, members, loadUser, notifications, projectState }) => {
   const project = projectState.project
   const projectMembers = _.filter(members.members, m => _.some(project.members, pm => pm.userId === m.userId))
+  // all feeds includes primary as well as private topics if user has access to private topics
+  let allFeed = projectTopics.feeds[PROJECT_FEED_TYPE_PRIMARY].topics
+  if (checkPermission(PERMISSIONS.ACCESS_PRIVATE_POST)) {
+    allFeed = [...allFeed, ...projectTopics.feeds[PROJECT_FEED_TYPE_MESSAGES].topics]
+  }
+  const allFeedCount = projectTopics.feeds[PROJECT_FEED_TYPE_PRIMARY].totalCount + projectTopics.feeds[PROJECT_FEED_TYPE_MESSAGES].totalCount
```
sumitdaga: @maxceem  
allFeedCount needs to be updated as well (checkPermission) ...although i don't think feedTotalCount is used anywhere!

maxceem: Thank you @sumitdaga, fixed it https://github.com/appirio-tech/connect-app/commit/0277743a8c47fb2843b599118c775047a44598c2

Issue #2938, opened 2019-04-08
[$40] Editing existing file title changes another file title


maxceem:
### Expected behavior
When we edit the title of the second file we should change the title of that file.

### Actual behavior
When we edit the title of the second file we change the title of the first file instead.

### Steps to reproduce the problem
- Use `pshah_manger` user create a new `App` project https://connect.topcoder-dev.com/new-project/app
- Upload file 1
- Upload file 2
- Change the title of file 1
- you would see that instead of changing the file 1 it would change the title of file 2.

### Screenshot/screencast
https://monosnap.com/file/tPHWX6OkvCipCGABeKXOiKwheQJJVK

This issues may have the same root as this one https://github.com/appirio-tech/connect-app/issues/2878

mfikria:
@maxceem is this ticket open for public?

maxceem:
Yes, it's open for everyone to pick up as a part of this Bug Bash challenge https://www.topcoder.com/challenges/30088391

Please, check the rules if you'd like to part.

mfikria:
@maxceem okay i take this one

maxceem:
@mfikria it's on you since you commented as per our Bug Bash rules. To keep it clear I can assign you to this issue in the UI. For this, please accept the invitation to this repository which you should get to your email.

mfikria:
@maxceem resolved in https://github.com/appirio-tech/connect-app/pull/2942. The problem is when editing the attachment it refers to the index instead to the attachment id

mfikria:
Note that i try this with local api-backend because i've got error when use https://api.topcoder-dev.com/v4/projects/ https://monosnap.com/file/vmcmjMnSXeDG7FRJ8UtnbQrchap0CO seems something's wrong with dev elasticsearch

maxceem:
@mfikria great! thank you for using local deployment. We are aware of the issue with Project Service and trying to get it back as soon as possible.

RishiRajSahu:
Working fine in dev. Closing.

PR #2959, opened 2019-04-16
Issue 2896 fix

PrakashDurlabhji:


maxceem:
@PrakashDurlabhji just in case would like to remind about this PR as the Bug Bash challenge in review phase now, which will be finished in 1.5 days.

PrakashDurlabhji:
@maxceem which api  needs to be used?
removeProjectAttachment?

or

updateProjectAttachement with proper payload?

maxceem:
@PrakashDurlabhji `removeProjectAttachment`, I think the issue is because you call with `idx` but I guess it should be called with attachmentId

maxceem:
@PrakashDurlabhji there are also some conflicts with other changes has been done during this Bug Bash, please kindly resolve them before I can accept this PR.

PrakashDurlabhji:
@maxceem is edit attachement working proper already? 

I dont think edit attachement is working as even it uses idx, 

Also there is no attachementId property in filelinks object, hence will implement same logic as links delete which splices the element at idx

maxceem:
> is edit attachement working proper already?

it was broken and it has been fixed in another issue https://github.com/appirio-tech/connect-app/issues/2938

> I dont think edit attachement is working as even it uses idx,

right, it didn't work with idx

> Also there is no attachementId property in filelinks object, hence will implement same logic as links delete which splices the element at idx

plese, get the latest commits in the `cf16` branch, and check how the issue https://github.com/appirio-tech/connect-app/issues/2938 has been fixed in this PR https://github.com/appirio-tech/connect-app/pull/2942

I think in that issue attachementId has been added so you can reuse it in your submission.

PrakashDurlabhji:
@maxceem MR updated with all points

Issue #2177, opened 2018-07-09
[$100] New post notification doesn't update the interface to the appropriate post


vic-topcoder:
### Expected behavior
When I click on a new post notification, the system would (re)load the page to show the post, and scroll to the post into view.

### Actual behavior
The interface doesn't load the appropriate post.

Aha! Link: https://topcoder.aha.io/features/TCCONNECT-465

Gauravseta:
@vic-topcoder can i take this? or is it not for open for pickup hence not a part of bug bash?

gondzo:
It's not part of the bug bash @Gauravseta

gondzo:
This can be implemented after #2306 is merged

maxceem:
@vikasrohit for now it doesn't work as expected in this issue.

Here is a demo where `customer` made one more comment but `manager` doesn't see it after clicking on the notification https://monosnap.com/file/PZzZ0p1vOv6QWQwUNxe194xGQGMof9

Note, before the fix for the issue https://github.com/appirio-tech/connect-app/issues/2465 I think this could work automatically but not anymore.

Before fix for https://github.com/appirio-tech/connect-app/issues/2465 we reloaded all the posts on the dashboard tab every time we navigate to it from any other place. But for https://github.com/appirio-tech/connect-app/issues/2465 I have to disable it, see explanation for this change here https://github.com/appirio-tech/connect-app/issues/2465#issuecomment-423019557.

So it looks like to keep smooth behaviour for transactions between tabs when using sidebar and still update posts when we click notifications, we have to add some additional callback for clicks on notifications which will force posts reloading. What do you think? Is there any better solution?

vikasrohit:
@maxceem I guess you linked wrong demo video in the comment. However, I do understand the issue here. Can we reload the topic in the FeedContainer when we see the `#` in the URL? I mean if the URL has `#comment-2345`, can just refresh the topic of the comment? I think doing that in only `componentWillMount` would work as ideally if the user is already on the same page where the notification leads to, that page should automatically ask user to refresh the page because of the availability of the new content.

vikasrohit:
> So I think we have to check if topic/comment is already loaded.

How would we do that? And what would be the cost in terms of time?

IMO, it should not be bad experience to always, without checking if it exists or not, load the specific topic and its posts only when the container is mounted and the URL contains the anchored link.
[...]

PR #2961, opened 2019-04-17
[WIP] Updates the interface to the appropriate post on notification

sagarbaver:
## Description
This PR attempts to address https://github.com/appirio-tech/connect-app/issues/2177.

I've raised this PR to gather some thoughts on the page reload strategy.

## Work done so far
- As suggested by @maxceem (https://github.com/appirio-tech/connect-app/issues/2177#issuecomment-478836122), I've added an event to listen to changes in the URL as a result of clicking on a new notification.
- I've added checks to determine if the notification is for a new feed, a new feed post a new phase or a new phase post.
- I've further gone on to determine if the content for which the user has been notified is already loaded in the view or not.

## RFC

- In the preliminary approach, once it is determined that the notified content is not yet loaded, we could potentially load it by one of the following ways:
  1. Asynchronously request for new feeds or phases. While loading new content asynchronously is ideal, I'm not sure if the experience of making the user wait longer having already clicked on the notification is frictionless (by displaying a loader to wait longer).
  2. Redirect to the newly posted content by refreshing the set URL (`history.pushState`). This although not ideal, appears as a faster response to taking the user to the desired notification.


maxceem:
Hi @sagarbaver 

Thank you for the update. Generally, the logic looks good, though it has some issues currently.

1. I prefer to go with the approach of asynchronously reload posts/topics using actions `loadDashboardFeeds`, `laodProjectMessages` and `loadPhaseFeed`.
   When we reload feed on the dashboard, we should call both `loadDashboardFeeds` (load public topics) and `loadProjectMessages` (load private topics) as we don't know if the new topic if private or public.

   For posts on the dashboard, we can detect if the post is inside private or public topic. If it's private, call `loadProjectMessages` if public, call `loadDashboardFeeds`. (To detect private/public topic we should check `topic.tag`. If `topic.tag === PROJECT_FEED_TYPE_MESSAGES` then it's a private topic, otherwise - public).

   For phases, we use `loadPhaseFeed`.

2. Looks like for phases logic would be a little bit more complicated. In case another user creates a new phase and after adds a new post to it. If we load posts for such case it wouldn't work. As we don't have ready actions to load only one phase, to keep is simple we can reload all the phases and their posts in such case (I think it would rare).
    Currently, phases are being loaded here: https://github.com/appirio-tech/connect-app/blob/cf16/src/projects/actions/projectDashboard.js#L47-L69. First we load all the phases, after we load all the posts for phases and after we load users for the posts if necessary.
    We can reuse the same code to reload all the phases in case if the phase with posts which we want to show is not loaded.

3. We should make sure that we show a loader when we load topics on the dashboard and posts for phases. I think for dashboard it should work automatically during topics loading. But for phases, we should add LoaderIndicator during phase posts loading for one phase.
    Also, we should show a loading indicator when we reload all the phases.

PS. I've pushed some update to the `cf16` branch, please get them, to avoid merging conflicts.

sagarbaver:
Thanks, @maxceem! Your comment is very helpful. I'll work on your suggestions as soon as I can.

maxceem:
Hi @sagarbaver 

When do you think you would have time for this? We are kind of out of the Bug Bash timeline rules, as an issue has to be fixed for 12 hours with a possible extension but this one already takes 3 days.

Thanks.

sagarbaver:
@maxceem I'm back on it.

maxceem:
@sagarbaver in the previous review you may keep the issue 1 optional:
image
So you may skip it and we will fix it separately.

When do you think you can complete the issues 2, 3 and 4?

sagarbaver:
@maxceem I'll get on with it today. Thanks for your review.

maxceem:
Hi @sagarbaver 

Do you have any progress with this issue? It is moving quite slow I think if we do not complete it asap, I would have to assign you from this issue and it would be included in the next Bug Bash.

Please, let me know if you have time to complete in near time.

sagarbaver:
Hi @maxceem. I'm currently on a planned trip with low connectivity. I'll be back to my city early tomorrow. I should be able to resume work on this only once I'm back. 
I fully understand if you want to move this to the next BugBash or mark it as up for grabs.

vikasrohit:
@maxceem are you planning to include this in next bug bash?

maxceem:
@vikasrohit yes, will put it to the next Bug Bash

PR #2965, opened 2019-04-19
Add-ons flag

maxceem:
Feature/add ons flag

This PR requires merging Project Service supporting PR first https://github.com/topcoder-platform/tc-project-service/pull/293

### DIFF
--- a/src/api/templates.js
+++ b/src/api/templates.js
@@ -3,7 +3,7 @@
  */
 import _ from 'lodash'
 import { axiosInstance as axios } from './requestInterceptor'
-import { TC_API_URL } from '../config/constants'
+import { PROJECTS_API_URL } from '../config/constants'
 
 const projectTemplates = require('./templates-json/project-templates.json')
 const productTemplates = require('./templates-json/product-templates.json')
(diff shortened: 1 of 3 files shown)


src/api/templates.js
```diff
@@ -3,7 +3,7 @@
  */
 import _ from 'lodash'
 import { axiosInstance as axios } from './requestInterceptor'
-import { TC_API_URL } from '../config/constants'
+import { PROJECTS_API_URL } from '../config/constants'
```
RishiRajSahu: @maxceem did we have this variable already present in our constants file, also was TC_API_URL also had the same value earlier ?

maxceem: Yes, both variables we have in our constant file.
`TC_API_URL` we use for many services, not only project service
while `PROJECTS_API_URL` we only use for project service.

During local testing, we want to point Connect App to the Project Service by changing `PROJECTS_API_URL`. But we should leave `TC_API_URL` as it is so other services work.

As metadata is loaded from project service we can use here `PROJECTS_API_URL`.

PR #2971, opened 2019-04-22
skip empty addons

maxceem:
> can we by pass the specific deliverable addon section when there is no addon for that particular deliverable e.g. as of now Deployment does not have any add on, we want to skip directly to the final summary screen when user selects only the Deployment as deliverable and answers all required question i.e. we want to skip the intermediate summary screen and the the empty screen which asks for selecting addons.

This is done. At the moment the solution uses section IDs to detect intermediate section and add-ons section:
```
section.id === 'summary-intermediate'
section.id === 'add-ons'
```
This binds the logic to the particular template. But anyway it looks like there is no other way to describe this logic or we should introduce some additional properties to define these sections. 

I think we can use this solution until we get some additional requirements or other templates examples. We should use these IDs for now for these two sections.

### DIFF
--- a/src/helpers/wizardHelper.js
+++ b/src/helpers/wizardHelper.js
@@ -315,11 +315,12 @@ export const forEachNode = (template, iteratee, iterateSublevelCondition) => {
  *
  * @param {Object}  template            raw template
  * @param {Object}  project             project data (non-flat)
+ * @param {Array}   productTemplates    product templates
  * @param {Object}  incompleteWizard    incomplete wizard props
  *
  * @returns {Object} initialized template
  */
-export const initWizard = (template, project, incompleteWizard) => {
+export const initWizard = (template, project, productTemplates, incompleteWizard) => {
   let wizardTemplate = _.cloneDeep(template)
   const isWizardMode = _.get(template, 'wizard.enabled')
   // try to get the step where we left the wizard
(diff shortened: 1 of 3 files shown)


src/helpers/wizardHelper.js
```diff
@@ -938,6 +940,34 @@ export const updateNodesByConditions = (template, project) => {
...
+
+      if (!_.get(q, '__wizard.hiddenByCondition')){
+        hasActiveAddOnSection = true
+      }
+    })
+  })
+
+  // skip intermediate summary step if no add-ons step found
+  if (!hasActiveAddOnSection) {
+    summaryIntermediateSection.__wizard.hiddenByCondition = true
```
vikasrohit: Do we need to hide the addonSection explicitly as well? I think it won't be shown any way when there is no visible question underneath, however, just to be more readable, could we explicitly mark the `addOnSection` hidden as well?

maxceem: @vikasrohit we can do that, though it would make the logic even more specific for the particular template. At the moment we can theoretically have other questions in the add-ons section. So the section still could be shown if there are any other questions to show.

But as we already have some other logic bound to the particular template structure, we can make it, to make sure the current logic is more reliable.

vikasrohit: Okay, makes sense. Lets put an inline comment mentioning the same. And leave the code as is.

maxceem:
@vikasrohit I've added an inline comment to the code as per discussion. And fixed template object mutation which has been there.

vikasrohit:
Thanks @maxceem

PR #2997, opened 2019-04-25
Enabling layout property for radio group component so that we can have vertical layout for radio group

vikasrohit:
Need to merge only if we need vertical layout for new kubik forms.


vikasrohit:
Merged PR as we are making another hotfix for #3008, taking the opportunity to push these changes as well to production.

PR #3025, opened 2019-05-11
issue 2969 fix

PrakashDurlabhji:


src/projects/detail/components/timeline/Form/Form.jsx
```diff
@@ -89,7 +89,7 @@ class Form extends React.Component {
               )
             case 'number':
               return (
-                <FormFieldNumber {...field} key={field.name} />
+                <FormFieldNumber disabled={field.disabled} {...field} key={field.name} />
```
maxceem: @PrakashDurlabhji I think we don't have to add `disabled={field.disabled}` separately, as `{...field}` already will pass all the properties from `fields` to the `<FormFieldNumber>` component. So this change looks unnecessary. Can we revert it?

PrakashDurlabhji: Yes sure will update it immediately and check if it works

PrakashDurlabhji:
@maxceem PR update

PR #3027, opened 2019-05-11
issue 2991 fix

PrakashDurlabhji:


PrakashDurlabhji:
@maxceem 

1. To display error message, it was needed to handle error in different reducers for each, also considering scope of issue, currently error has been handled at GET_PROJECTS_FAILURE.

2. on receving error props in coderbot component, on debugging and finding its properties, it came to know that, though stack trace can be seen at console, but actually on converting error props to string only message available is "Error: demo error", that is what ever data passed in error constructor is recevied, stack trace is not available in error object. also on finding keys and values of error object nothing was found.

3. button is near to text, reducing space.

4. encode uri component is used.

PR updated

PrakashDurlabhji:
@maxceem updating immediately

PrakashDurlabhji:
@maxceem all changes are implemented and ready

PR #3028, opened 2019-05-11
issue-3011 mention in private message

applehit16:


src/projects/detail/containers/FeedContainer.js
```diff
@@ -560,7 +560,13 @@ FeedContainer.PropTypes = {
...
 const mapStateToProps = ({ projectTopics, members, loadUser, notifications, projectState }) => {
   const project = projectState.project
-  const projectMembers = _.filter(members.members, m => _.some(project.members, pm => pm.userId === m.userId))
+  const projectMembersMap = _.keyBy(project.members, 'userId')
+  const projectMembers = Object.values(members.members) 
+    .filter(m => projectMembersMap.hasOwnProperty(m.userId))
+    .map(m => ({
+      ...m,
+      role:projectMembersMap[m.userId].role
+    }))
```
maxceem: @applehit16 I see you've updated the filtering code here. As I understand your version should work faster as it uses object keys instead of an array. But it took some time from me to understand how it works :-). 

I think speed is not that much improtant here, while it would be good to keep code easier to understand for any person.  Can we keep the previous version of this code and just add 
```js
.map(m => ({
      ...m,
      role:projectMembersMap[m.userId].role
    }))
```
to populate `relo` for the members?

applehit16: @maxceem one reason is for the speed, the time complexity is O(1) instead of O(n^2). The other reason is currently the project doesn't support lodash chaining. so in order to add `map` method, I need to do something like
```js
_.map(_.filter(members.members, m => _.some(project.members, pm => pm.userId === m.userId)), (m) => ({
  ...m,
  role: project.members.filter(pm => pm.userId === m.userId)[0].role
}))
```
Since we don't generate the hashMap, so that we need to traverse the array again for `role` which is another O(N^2) case. And the code is actually harder to read I guess?
Just my 2 cents, it's up to you

maxceem: @applehit16 Thank you for the detailed explanation, makes sense.

We can chain regular `.map` function as lodash returns a regular array, like this:
```js
_.filter(members.members, m => _.some(project.members, pm => pm.userId === m.userId).map((m) => ({
  ...m,
  role: project.members.filter(pm => pm.userId === m.userId)[0].role
}))
```

But let's keep your code for now. I've never seen such an approach so it was harder for me to understand what is going on. But it brings some optimizations, will see how it works for other people.

Thank you.

Issue #2937, opened 2019-04-04
Show links from the posts in the sidebar


maxceem:
From @vikasrohit 

We want to identify that if user has pasted a link to a file or a general hyperlink in a post/thread. And if we identify while user is writing the message, we would create a proper object of link or file in our system so that we can render that link or file in the left bar of the project. We can treat the title of the thread as title of the link/file and the actual URL typed/pasted by the user as URL for the link/file.

maxceem:
As I understand we have to choose a way how we would determine if the link we have pasted should be treated as a link or as a file.

Most likely we would have to download the linked resource and check its mime type or something like that.

maxceem:
@vikasrohit 

1. If we detect that link leads to a file it looks like we would need to download it and upload to S3 bucket which may take some time. As an alternative we can change our attachment implementation in project service to support third-party links to files. It would make this feature almost instant, but such links to files could get broken if files are removed in these third party locations. What do you think?

2. If we edit post and change link, should we update respectively link or file?

3. if we edit post and delete link, should we delete according link or file?

4. > We can treat the title of the thread as title of the link/file

    One topic may have many posts inside and every post may have several links inside. So it looks like there would be multiple links/files with the same name. We can try to use several ways of determining more link/file name using the next mehtods if you like them:
    
    1. If we have a "downloadable" link to a file with defined `Content-Disposition` it may have a filename defined inside this header: `Content-Disposition: form-data; name="fieldName"; filename="filename.jpg"` so we can use it as file name.
    2. if we don't have filename inside `Content-Disposition`, we can check if the URL is ended with something which looks like a filename with an extension: `<filename>.<ext>`
    3. if URL doesn't have it we can still try to use the ending of URL for the filename, but it may looks very bed like `https://tpc.googlesyndication.com/simgad/2536338633819292849`. So probably in such case we should use the topic name. But we can still try to guess the file exntesion by the MIME type from `Content-Type` header, for example `image/png` - > `.jpg`.
    4. if have to use topic name for link/file name and we already have the same name we can add a number in the parenthesis to the name, like:
       ![image](https://user-images.githubusercontent.com/146016/55706215-36190a80-5a13-11e9-9bab-b32fa54034eb.png)
    5. Another way would be to let user define file/link title using interface of the Drat.js post editor. This may require additional thinking about how to do it form UX point of view.

   Let me know if you would like to have any of these ways to handle file/links titles.

RishiRajSahu:
[...]

PR #3029, opened 2019-05-12
New DraftJs Link Editor functionality…

maxceem:
New DraftJs Link Editor functionality (like in Google Docs) which would also help to fulfill the requirements of #2937

In particular, it automatically shows a small popup as soon as link is created, to encourage user to add a title to the link.


RishiRajSahu:
@maxceem thanks will check it out.

package.json
```diff
@@ -84,10 +84,10 @@
     "brace": "^0.11.1",
     "classnames": "^2.2.3",
     "coffeescript": "^1.12.7",
+    "decorate-component-with-props": "^1.1.0",
```
vikasrohit: It seems like this repo is not very much popular and have potential of getting unmaintained very soon. @maxceem Do you know why exactly we need this library and can we avoid using this?

package.json
```diff
@@ -125,6 +125,7 @@
     "react-infinite-scroller": "^1.1.1",
     "react-layout-pane": "^0.1.16",
     "react-modal": "^1.9.7",
+    "react-popper": "^1.3.3",
```
vikasrohit: I am okay with other new library `react-popper` as it seems very popular and maintained. However, we already have `react-modal` library included for showing modals, can we use that instead of introducing a new one?

maxceem: We already use `react-popper` in `react-component` library https://github.com/appirio-tech/react-components/blob/feature/connectv2/package.json#L82. The purpose of this library is quite different from `react-modal` it has very good logic for positioning popups and we use widely for implementing dropdowns, selects, tooltips and so on.

`react-modal` is very simple and basically can only show modal windows at the center of the screen. The main purpose of this lib, is to be a portal, so we can include modal HTML in any children react component, while it would be mounted at the root of the app, to avoid style collisions.

package.json
maxceem: Good catch. Usage of that lib was redundant in the current code so I've removed it and pushed to this branch.

maxceem:
@vic-topcoder 
> Also, can we have the plugin separate out (I am okay for having separate task for that) in react-components library for sharing the newly developed DraftJS plugin?

Yes, we can do that and during the challenge it was one of the requirements to keep the plugin code isolated so it could be potentially reused.

Just want to note, that at the moment we don't have DraftJs editor component inside `react-components` library. It's fully implemented inside Connect App. Would you like to move DraftJs editor component from Connect App to `react-components` and reuse it from there?

package.json
vikasrohit: Thanks

package.json
vikasrohit: Okay, makes sense. Last thing, is it possible to use `react-popper` in place of `react-modal`, just in case?

vikasrohit:
> Just want to note, that at the moment we don't have DraftJs editor component inside react-components library. It's fully implemented inside Connect App. Would you like to move DraftJs editor component from Connect App to react-components and reuse it from there?

@maxceem What are the other components for related to the DraftJS which we need to move? And I assume once we refactor them to `react-components` repo, it won't have any specific logic for Connect.

package.json
maxceem: In theory, I think it's possible. If we reuse `react-popper` we would need to rewrite some logic/styles which already done for `react-modal`. Also, there could be some issues on the way which could need to be fixed, when we customize the behavior of `react-popper` to implement Modal functionality.
So IMO these libraries play their roles quite nicely for their purposes. And for replacing `react-modal` with `react-popper` we would need to make more detailed research to check how easily it would fit to replace it.

maxceem:
> @maxceem What are the other components for related to the DraftJS which we need to move? And I assume once we refactor them to react-components repo, it won't have any specific logic for Connect.

I think we only have to move DraftJs editor [RichTextArea](https://github.com/appirio-tech/connect-app/blob/feature/draftjs-link-editor/src/components/RichTextArea/RichTextArea.jsx#L514) component as a component which would let us edit Rich Text. At the moment this component in Connect App is tightly bound with other Connect specific features. So the whole component renders next part:

image
 It includes:
- field to edit title
- user avar
- `Attach a file` button
- `Private` switch
- `Cancel` and `Post` buttons

If we move this component to `react-components` we should only move Rich Text editor:
image

- the tricky thing would be to handle panel with buttons, as in our design it's mixed with other Connect specific functionality on the right. So on the one hand this panel should be quite independent from DraftJs component so we can put it where we want. On the other hand, it's tightly bound with DraftJs component, so if we keep this panel as a separate component, we should somehow bind functions of the buttons with DraftJs editor. Preferably, without writing a lot of code when reusing this component.

package.json
vikasrohit: Agreed.

vikasrohit:
Thanks @maxceem for the detailed update. Lets move this discussion to a separate issue which we may take up later.

PR #3030, opened 2019-05-12
Issue 2936 fix

rashmi73:


src/components/NotificationBadge/NotificationBadge.jsx
```diff
@@ -0,0 +1,41 @@
...
+        return (
+            getBadge(count)
+        )
+    }
+}
+
+NotificationBadge.propTypes = {
+   count: PropTypes.number.isRequired,
+   text: PropTypes.string,
+   id: PropTypes.number
```
maxceem: @rashmi73 `NotificationBadge` component looks like doesn't use `id` but have in the property list. Can we remove it?

rashmi73: @maxceem it is removed

rashmi73:
@maxceem all points are covered and updated PR.kindly check

PR #3035, opened 2019-05-14
fixes issue 2177

sumitdaga:


maxceem:
@sumitdaga how is it going? When do you think you would be able to complete this issue?

sumitdaga:
@maxceem.  
Believe it or not , I have spent almost entire today working on it ...It's quite complicated ... Hopefully tomorrow I will be able to finish

maxceem:
@sumitdaga oh, ok. Which one of the 3 items above is the most complicated?

sumitdaga:
@maxceem there are primarily 3 more things that needs to be addressed

1.  when we are on project plan tab and there is a notification for a new comment (which is not yet on redux store) then the new comment is not loaded (even with the my fixes so far) .

2. on project plan tab when there is a hash in url it scrolls to it on every update to the component (something you said is not desirable) . Note that this was an existing issue.

3. Showing loading indicators when only phases are re-loaded in the project plan tab (i don't think this was already implemented) and then loading indicator when only posts for one phase are being loaded ...

and then there are other smaller things to check (and some which we may have missed still) .... and it takes some amount of time to verify a fix!

maxceem:
I see, thanks for the detailed update @sumitdaga.

sumitdaga:
@maxceem  
btw i think the loadPhaseFeed method i am using https://github.com/appirio-tech/connect-app/blob/dev/src/projects/actions/phasesTopics.js#L37 from here  loads the feed along with members  (FYI for your concern about loading members when phase/phasePosts are loaded)

sumitdaga:
@maxceem  i am almost done!  
but as of now i cannot add posts to a new phase ? the discussions tab is just empty ...can you please verify if its working for you ?

maxceem:
@sumitdaga works good to me at the moment https://monosnap.com/file/c23EAxyTylOPUh6q2vPapqFGFFbwZ2

sumitdaga:
@maxceem  not to an existing phase  
if you add a new phase and then open the discussions tab ...and the strange thing is it was working earlier

maxceem:
@sumitdaga yeah, that's strange. Checking...

maxceem:
@sumitdaga works in this project for me https://connect.topcoder-dev.com/projects/8048/plan. Could you please check.

sumitdaga:
@maxceem  i am unable to push changes  
i have enabled two factor auth in my github account ...i guess i will need to be added again to the group ?

maxceem:
@sumitdaga sent you invitation again. Though it's strange, as you push changes to your fork, but just in case.

sumitdaga:
@maxceem   
oh sorry you are right,  so after adding 2 factor authentication i couldn't login using to github on my terminal  with username/password (my origin url was https based) and i changed the origin url to ssh and it worked ...

ok so about the fix:
this is what my general approach looks like - to load a feed/post/phase which is not yet loaded i have put the relevant `handleUrlHash` functions in the component which renders that particular post/phase etc so that scroll happens only once (same for expanding phase, i put the `handleUrlHash` function in `ProjectState` component, so that it gets expanded only once) when it gets loaded for the first time while the actual loading of the phase/post is handled in `ProjectInfoContainer`  
I think i have handled the loadingIndicator issue wherever it was needed  
To check all the possibilites , you need to check if a feed can be scrolled to from home (/projects) page, dashboard tab and other tabs ...similarly for phases/posts 

there are many changes, so goes without saying you will need to check it carefully so that i have not missed something or create some new bug ...

and of course if the fix works and all is ok please consider the prize on this one :)

maxceem:
@sumitdaga thank you for the details. I'll check it out tomorrow, regarding all your points :-)

src/projects/detail/containers/PhaseFeedHOC.jsx
```diff
@@ -43,9 +43,9 @@ const phaseFeedHOC = (Component) => {
     }
 
     componentWillMount() {
-      const { isLoading, loadPhaseFeed, phase, project } = this.props
+      const { isLoading, topic, loadPhaseFeed, phase, project } = this.props
 
-      if (!isLoading) {
+      if (!isLoading && !topic) {
```
sumitdaga: @maxceem about this  
it seems the feed of a phase was being loaded every time for ex when user clicked on discussions tab after expanding phase ..it was not apparent because there was no loading indicator ...so i added this check ...i am not sure what is desirable here though ?

maxceem: Hmm, it feels that it shouldn't be loaded every time this component is mounted as `topic` suppose to come from the redux store. So the topic should be loaded when the component is mounted only if it's not loaded yet to redux store. Let me know if I'm wrong or there is any issues in the logic.

maxceem:
Posting it here just not to forget some case:
1. Dashboard:
   1. new `post` is `already` loaded
   2. new `post` is `not yet` loaded
   3. new `topic` is `already` loaded
   4. new `topic` is `not yet` loaded

2. Phases:
    1. new `post` is `already` loaded for `already` loaded `phase`
    2. new `post` is `not yet` loaded for `already` loaded `phase`
    3. new `post` is `not yet` loaded for `not yet` loaded `phase`

sumitdaga:
and check all the above from these pages `/projects`, `projects/${id}` (dashboard tab), `projects/${id}/plan` (project plan tab) ...since the behevaiour depends on which component is already mounted or not mounted

maxceem:
Here is my testing:

1. Dashboard (from Project Plan):
   1. new `post` is `already` loaded - works https://monosnap.com/file/nKf1IG1gx2BtvVWDqkVylF90DyhTYi
   2. new `post` is `not yet` loaded - almost works - ❌ `DIDN'T SCROLL` https://monosnap.com/file/BRvnyyr7tTrgQEjTAV0UzlJtvBDTQO
   3. new `topic` is `already` loaded - works https://monosnap.com/file/9Byx8kVhICnTbq6OBsN5S7KYE2IeZZ
   4. new `topic` is `not yet` loaded - almost works - ❌ `DIDN'T SCROLL` https://monosnap.com/file/h7NogN9GsE6ecla6As3OiEU4mlw1IZ

2. Phases:
    1. new `post` is `already` loaded for `already` loaded `phase` - almost works - ❌ `RELOAD PHASE with already loaded post` - https://monosnap.com/file/jMDs843lGLdbUzi44NM403n8YwCUhq
    2. new `post` is `not yet` loaded for `already` loaded `phase` - almost works - ❌ `reload posts for ALL pahses instead of only one` - https://monosnap.com/file/T4kj6CUODMgQPxG69HC8FkrnQqAgRG
    3. new `post` is `not yet` loaded for `not yet` loaded `phase` - works - https://monosnap.com/file/2rPsrr9lMoCUQshhvB8VurWPB6kHSz

#### Issue
When I open the specification tab on any other phase it automatically scrolls to the phase I've clicked in the notification before https://monosnap.com/file/fchpLZhvGbIDXRlN7wc2NoVt0tZvxa This is an existent issue, I know you attempted to fix it, could you please check?

src/projects/detail/components/PhaseFeed/PhaseFeed.jsx
```diff
@@ -5,26 +5,54 @@
...
+  constructor(props) {
+    super(props)
+  }
+
+  componentWillMount() {
+    !_.isEmpty(this.props.location.hash) && this.handleUrlHash(this.props)
+  }
+
+  componentWillReceiveProps(props) {
+    if (!_.isEmpty(location.hash) && props.location.hash !== this.props.location.hash) {
```
maxceem: @sumitdaga here you use `location.hash` which is a native browser object, while in other places you use `props.location.hash` which comes from `withRouter` HOC. Can we use here also from HOC for consistency?

src/projects/detail/components/ProjectStage.jsx
```diff
@@ -130,23 +130,26 @@ class ProjectStage extends React.Component{
...
 
-  componentDidUpdate() {
-    const { phaseState } = this.props
-    if (_.get(phaseState, 'isExpanded')) {
-      const scrollTo = window.location.hash ? window.location.hash.substring(1) : null
-      if (scrollTo) {
-        scrollToHash(scrollTo)
-      }
+  componentWillReceiveProps(props) {
+    if (!_.isEmpty(location.hash) && props.location.hash !== this.props.location.hash) {
```
maxceem: @sumitdaga I'm not sure, but wouldn't it be better to use methods `componentDidMount` and `componentDidUpdate` so we run scrolling `after` component is rendered/rerendered? Or at least closer to that time.

src/projects/detail/components/PhaseFeed/PhaseFeed.jsx
sumitdaga: @maxceem  this is a mistake ..should be `props.location.hash`

src/projects/detail/components/ProjectStage.jsx
sumitdaga: @maxceem  i think its better to use the `did` methods rather than the `will` methods ...i guess i used the `will` methods because at most places i saw these in the connect app ...i will try to use the `did` methods and see how it goes

sumitdaga:
@maxceem  
i have fixed all other issues except  
> When I open the specification tab on any other phase it automatically scrolls to the phase I've clicked in the notification before  

the above issue is happening because the `specification tab` renders the `SpecSection` (https://github.com/appirio-tech/connect-app/blob/dev/src/projects/detail/components/SpecSection.jsx#L430) component which is wrapped in `ScrollToAnchors` method ...which causes scrolling if there is a URL hash   

i am not sure how to fix this as specSection is used at multiple other places ...do you have a suggestion ?

sumitdaga:
@maxceem  i have pushed rest of the fixes except the last fix ...i will see tomorrow if i can figure an efficient fix for this

maxceem:
Thanks @sumitdaga I'll check it out later today.

> i am not sure how to fix this as specSection is used at multiple other places ...do you have a suggestion ?

I think the quickest fix would be to add `disableAutoScrolling` flag to the `specSection` component which would remove or disable this scrolling HOC. And on the `Project Plan` tab we can set this flag to disable this functionality as it's only used on the `Spec` tab. So we don't use it here and can disable to avoid conflicts.

Let me know if see any drawbacks.

sumitdaga:
> I think the quickest fix would be to add disableAutoScrolling flag  

@maxceem  oh darn , you are right ...i have fixed all issues i think !  
let me know if you see any more issues!

thanks

maxceem:
Something is with my internet, so I cannot upload video. But I did the same actions as before.

1. Dashboard (from Project Plan):
   1. new `post` is `already` loaded -  ❌ I've been immediately scrolled to the post - `smooth scrolling doesn't work` 
   2. new `post` is `not yet` loaded - ❌ I haven't been scrolled to the post - `scrolling doesn't work` 
   3. new `topic` is `already` loaded -  ❌ I've been immediately scrolled to the post - `smooth scrolling doesn't work` 
   4. new `topic` is `not yet` loaded - ❌ I haven't been scrolled to the topic - `scrolling doesn't work` 

2. Phases (from Dashboard):
    1. new `post` is `already` loaded for `already` loaded `phase` - ✅ no unnecessary loading now -  ❌ I've been immediately scrolled to the post `smooth scrolling doesn't work`  
    2. new `post` is `not yet` loaded for `already` loaded `phase` -  ✅ works perfectly - loads only posts for one phase including members and scrolls smoothly
    3. new `post` is `not yet` loaded for `not yet` loaded `phase` - looks good, ❌ but it doesn't load `Timeline` tab for a newly added phase https://monosnap.com/file/ktOT0e1QcB8dRVjcW19NbFehwyboz0. After reading Timeline is there, also it's shown after creating in another browser. Also, it scrolls to the correct place, though scrolling is almost immediate, not sure if it's just because the CPU is busy with something at the same moment.

sumitdaga:
@maxceem  
let me know if everything is fine now?

maxceem:
@sumitdaga Implementation for loading phases looks like working good!

#### Test results

1. Dashboard (form Project Plan):
   1. new `post` is `already` loaded - ✅ works - https://monosnap.com/file/yi6o18DLq8Tdxd6D4LEHFhnMlXCtxu
   2. new `post` is `not yet` loaded - ❌ doesn't scroll - https://monosnap.com/file/UpQ5sTRriG3LNNvDXRAe5dpUSpcPhd
   3. new `topic` is `already` loaded -  ✅ works - https://monosnap.com/file/33MNu7JmRSiScq8TOgkm3J3H9M4mar
   4. new `topic` is `not yet` loaded - ❌ doesn't scroll - https://monosnap.com/file/5ON8Qv74T4BIFSProWfmD94qj6ZuPd

2. Phases (form Dashboard):
    1. new `post` is `already` loaded for `already` loaded `phase` - ✅ works
    2. new `post` is `not yet` loaded for `already` loaded `phase` - ✅ works
    3. new `post` is `not yet` loaded for `not yet` loaded `phase` - ✅ works https://monosnap.com/file/PStGBeOfSZwSBEPV2NabVtIeX44AAF

3. Phases (form Project Plan):
    1. new `post` is `already` loaded for `already` loaded `phase` - ❌ doesn't do anything https://monosnap.com/file/fiqIw0dUr5fctyFNAwmaqjKy15DGRB
    2. new `post` is `not yet` loaded for `already` loaded `phase` - ❌ doesn't do anything https://monosnap.com/file/otoO9CWZkRCJi4rFLowSPYAjFWBC04
    3. new `post` is `not yet` loaded for `not yet` loaded `phase` - ❌ doesn't do anything https://monosnap.com/file/8yJ3QL60l1Gu5kiUhnIQL4HYkBmLIo

#### Issues
1. If I have some URL to the phase post and open `Timeline` with that post, it automatically scrolls to the `Timeline` tab https://monosnap.com/file/YdL8fjQmUABCO01NrwUCalSDn2gp1r. Expected behavior: it shouldn't scroll to the Timeline tab when we open it.

sumitdaga:
@maxceem  
this should do it ...i am not sure of the exact reason but my best guess is that using withRouter does not always trigger changes when location prop changes so i have added location as a prop from the top down so that at least these changes are triggered ...  
let me know if its working now ..or you think it can be better handled!

vikasrohit:
@maxceem @sumitdaga we have similar usage of `location` prop at multiple places to fix similar problems few examples are https://github.com/appirio-tech/connect-app/blob/dev/src/components/TopBar/ProjectsToolBar.js#L199, https://github.com/appirio-tech/connect-app/blob/dev/src/routes/metadata/components/MetaDataLayout.jsx#L16 and https://github.com/appirio-tech/connect-app/blob/dev/src/components/TopBar/ProjectToolBar.js#L79.  However, I think we don't need to pass it through to the bottom of the hierarchy. Its only the component, which is blocking the re-rendering on location change event, that is need to be passed `location` as prop. Could you please check if passing `location` only to the super parent which is blocking the re-rendering, works?

sumitdaga:
@vikasrohit  @maxceem  
i have passed the location prop from and to the minimum possible places for the issue to get fixed (to be best how much i could)
i checked the entire heirarchy and found where all it was blocked

sumitdaga:
@vikasrohit @maxceem  
just a suggestion , it may be worth its while to check if a custom withRouterHOC fixes the issue because for anyone who is not aware this becomes very frustrating (i was yesterday :))

maxceem:
#### Test results

1. Dashboard (form Project Plan):
   1. new `post` is `already` loaded - ✅ works -https://monosnap.com/file/9QRhi0CJWeILERrk87sboseqHVLqpa
   2. new `post` is `not yet` loaded - ❌ doesn't scroll - https://monosnap.com/file/8AJbDNYaYKXufZdkOhJjkQmkLRwMSr
   3. new `topic` is `already` loaded -  ✅ works - https://monosnap.com/file/fYoBBRWPdxYBg2kzBomObREiT2utOk
   4. new `topic` is `not yet` loaded - ❌ doesn't scroll - https://monosnap.com/file/82SaJYGNTbSPJm1z8C4MKedLbyN8fA

2. Phases (form Dashboard):
    1. new `post` is `already` loaded for `already` loaded `phase` - ✅ works - https://monosnap.com/file/eLw5ohr8DbSxNE1j6NX0HFMLuu1xMi
    2. new `post` is `not yet` loaded for `already` loaded `phase` - ✅ works - https://monosnap.com/file/rdAaEWK8xKDriKJiJ7mTyfMrvBfJ3T
    3. new `post` is `not yet` loaded for `not yet` loaded `phase` - `cannot test now because of some server issue`

3. Phases (form Project Plan):
    1. new `post` is `already` loaded for `already` loaded `phase` -  ✅ works - https://monosnap.com/file/iOeZQpqWHH5FrBEjOeKhoauonMYt1a
    2. new `post` is `not yet` loaded for `already` loaded `phase` - ✅ works - https://monosnap.com/file/9Ghzztx6nFf9Da7nPxKOxq72QbDOI1
    3. new `post` is `not yet` loaded for `not yet` loaded `phase` - `cannot test now because of some server issue` 

I was testing with the latest commits:

image

maxceem:
> Could you please check if passing location only to the super parent which is blocking the re-rendering, works?

@vikasrohit it looks like we have to update for every component which has `shouldComponentUpdate` and stops re-renders.

> @vikasrohit @maxceem
> just a suggestion , it may be worth its while to check if a custom withRouterHOC fixes the issue because for anyone who is not aware this becomes very frustrating (i was yesterday :))

Good idea. Though as I see from the `react-router` documentation, the way it's done by @sumitdaga is the recommended way https://github.com/ReactTraining/react-router/blob/v4.1.1/packages/react-router/docs/guides/blocked-updates.md

Also, as per discussion in a related [issue](https://github.com/ReactTraining/react-router/issues/5037) (as not only us come across with it) the other way of implementing `withRouter` HOC would be complicated and prone to various issues, see [comment](https://github.com/ReactTraining/react-router/issues/5037#issuecomment-297470402).

sumitdaga:
@maxceem  
all issues should be fixed now, or at least i hope :)

src/projects/detail/containers/FeedContainer.js
```diff
@@ -79,9 +78,6 @@ class FeedView extends React.Component {
 
   componentDidMount() {
     window.addEventListener('beforeunload', this.onLeave)
-  }
-
-  componentWillMount() {
```
maxceem: @sumitdaga  can wee keep here `componentWillMount`? As we use init to build comments inside Feed, and I guess it's better to keep it as before. So comments are rendered maybe faster.

maxceem:
Thanks @sumitdaga I've tested and everything works good.

#### Test results

1. Dashboard (form Project Plan):
   1. new `post` is `already` loaded - ✅ works - https://monosnap.com/file/Q7GVqlD5u9R5QP58sDbVreIVnmzseW
   2. new `post` is `not yet` loaded - ✅ works - https://monosnap.com/file/dd0BIuk3yxg8a3dauZER7HAXQ8d2wV
   3. new `topic` is `already` loaded - ✅ works - https://monosnap.com/file/KPI915QUwxUXToQfBDM60LEzfMWe3i
   4. new `topic` is `not yet` loaded - ✅ works - https://monosnap.com/file/H9ACmj4P0ny0sw763DexZMP2Zv15DS

2. Phases (form Dashboard):
    1. new `post` is `already` loaded for `already` loaded `phase` - ✅ works - https://monosnap.com/file/YMtbqfetvHU0no0ObFe18PKeSag03G
    2. new `post` is `not yet` loaded for `already` loaded `phase` - ✅ works - https://monosnap.com/file/DNEceMXRsnqstTjD4Gtd2lxCj7rk36
    3. new `post` is `not yet` loaded for `not yet` loaded `phase` -  ✅ works - https://monosnap.com/file/ljM5MZxhzL4Bb10vnijKGwqp2DaeVs (`NOTE` I guess in the future we may consider improving this case and load only new phases, as currently it takes some time to load all the phases before the scrolling works. But as this case should be very rare, we should fine for now).

3. Phases (form Project Plan):
    1. new `post` is `already` loaded for `already` loaded `phase` - ✅ works - https://monosnap.com/file/BMgJXlDGRvZyV52CENc36lAyZrT8TW
    2. new `post` is `not yet` loaded for `already` loaded `phase` - ✅ works - https://monosnap.com/file/1Wv8BhTI5QvJg7rrKkCccABjUCdMmr
    3. new `post` is `not yet` loaded for `not yet` loaded `phase` - ✅ works - https://monosnap.com/file/lwbCseFKpuUMwRkLFxPiaUDLYddJJF (`NOTE` I guess in the future we may consider improving this case and load only new phases, as currently, it takes some time to load all the phases before the scrolling works. But as this case should be very rare, we should fine for now).

### Enhancement
The only thing I see that the way with passing location and calling `withRouter` on different levels - is the way to handle it extremely prone to bugs. So on the second thoughts, I agree with @vikasrohit this, should be fixed some other way, the best with introducing a new HOC like `withLocation` which would always update when the location is changed.

So we can conclude this PR at this moment. But instead of merging it to DEV, we would keep it in a separate branch and try to achieve the same with a custom `withLocation` component.

maxceem:
Note, this PR would be merged to `cf17-notifications-reload-subject` branch for now.

src/projects/detail/containers/FeedContainer.js
sumitdaga: @maxceem  
oh i did not intend to remove it, i was testing something and forgot to add it back ...should i put the init function inside `componentWillMount` (like before) or `componentDidMount` (which is also present)

maxceem: @sumitdaga please, keep it as it was before.

PR #3052, opened 2019-05-21
Update to v5 notifications api

gondzo:


vikasrohit:
I thought, `limit` is the standard parameter for setting the page limit.

gondzo:
v5 standard uses `per_page` and `page`

gondzo:
here is the v5 standard document for reference (I'm not sure if it's hosted somewhere??)

[CORE-v5APIStandards-040518-2128.pdf](https://github.com/appirio-tech/connect-app/files/3201027/CORE-v5APIStandards-040518-2128.pdf)

sachin-maheshwari:
@gondzo how we are using next pages in connect for list-notifications ? changes are required there too ?

gondzo:
We aren't using it yet - we're just fetching a fixed number of notifications. That might even be ok since we always want to fetch all unread notifications, there is no pagination support

vikasrohit:
Merging to dev as now tc-notifications is deployed to production with v5 changes and it is still backward compatible

PR #3053, opened 2019-05-21
Show the post attachments in the left panel

mfikria:


maxceem:
@mfikria

Good start. I've noticed a couple of small issues: https://monosnap.com/file/K1hGoEI98kUP8Pdh9IPDhGih3OWgYC.

1. We should list on the sidebar human-readable titles, same like in posts.
2. And we should order them same like posts in reverse order.

maxceem:
Thank you for update @mfikria. When do you think you would be able to completely delete file functionality?

maxceem:
Just remembered one note: we should only be able to delete files from the posts we are the author. If it's an attachment not from our post we should see a delete button.

maxceem:
Fixes for order, name, and link address of post attachments work good.

mfikria:
@maxceem when deleting topic attachment. Do we only need to call DELETE /v5/topics/attachments/:attachmentId/delete or also need to call edit post endpoint to remove it from attachmentIds?

mfikria:
> Thank you for update @mfikria. When do you think you would be able to completely delete file functionality?

working on it, today

maxceem:
@mfikria Good question. As so far we don't have to call `/v5/topics/attachments/:attachmentId/delete` when we delete attachments from the posts, let's not call it too here. So only call edit post endpoint.

mfikria:
@maxceem the delete action should be working now. Note the attachment could only be deleted by its creator

src/projects/detail/containers/ProjectInfoContainer.js
```diff
@@ -257,6 +261,69 @@ class ProjectInfoContainer extends React.Component {
...
+        const attachments = post.attachments
+          .filter(attachment => attachment.id !== attachmentId)
+        const attachmentIds = attachments.map(attachment => attachment.id)
+        return saveFeedComment(topicId, feed.tag, {
+          id: postId,
+          content: post.rawContent,
+          attachmentIds
+        }).then(() => {
+          post.attachments = attachments
+        })
```
maxceem: There are few issues related to this code with using `then`:
1. `post.attachments = attachments` is most likely mutates object in the redux store or component, but we never should mutate objects in redux store or component state, see https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns
2. When we use `then` here, we kind of skip out action-> reducer workflow. Most likely we don't need this `then` here. `saveFeedComment` method calls action [SAVE_PROJECT_FEED_COMMENT](https://github.com/appirio-tech/connect-app/blob/dev/src/projects/actions/projectTopics.js#L202) which updates post in the redux store [here](https://github.com/appirio-tech/connect-app/blob/dev/src/projects/reducers/projectTopics.js#L465). So after `saveFeedComment` is done, the post object should be automatically updated in redux store and come from the props updated.

src/components/LinksMenu/LinksMenuAccordion.jsx
```diff
@@ -20,9 +26,31 @@ class LinksMenuAccordion extends React.Component {
     })
   }
 
+  onDeleteConfirm() {
+    const link = this.props.link.children[this.state.linkToDelete]
+    if (link) {
+      this.props.onDeletePostAttachment({ topicId: link.topicId, postId: link.postId, attachmentId: link.attachmentId, topicTag: link.topicTag })
+        .then(() => {
+          this.onDeleteCancel()
+        })
```
maxceem: Using `then` here kind of overcomes action->reducer workflow here. We shouldn't be using `then` after acitons at least in `components`. Can it be done some other way following Redux best practices?

mfikria:
@maxceem updated, please check it again. Thanks!

Issue #3031, opened 2019-05-14
[$100] Attached file is not shown immediately when creating a post


maxceem:
When I create a post and attach a file the file is not shown after I click `Post` and post is created. I have to reload the page and after I can see the file I've attached to the post. See demo video: https://monosnap.com/file/9B2cbyBzlLc7oQYE0rmDHZcofUdIaU

@vikasrohit should we include it to the CF?

vikasrohit:
Yes, please and thanks for catching this.

maxceem:
This issue is now open for pickup.

PrakashDurlabhji:
@maxceem pls give to me thanks

maxceem:
sure

PrakashDurlabhji:
@maxceem I investigated a lot  with efforts and actually it is not a frontend issue but a backend issue as attachements array is coming blank in reducer 
 
CREATE_PROJECT_FEED_COMMENT_SUCCESS

hence it is not displayed first time.

and when we reload the page, then in LOAD_PROJECT_FEEDS_SUCCESS has that same object of post filled with attachements property.hence we are able to see attachements.

2. coming back to CREATE_PROJECT_FEED_COMMENT, it makes a post call, response of which shall contain attachements property filled, but it is blank hence that is cause of it.

Need some time and extra efforts to solve this and PR need to be raised on backend.

maxceem:
Thank you @PrakashDurlabhji for putting efforts into it, great finding.

@vikasrohit There is an issue with the message service:
- when we create a post with attachments:

   ![image](https://user-images.githubusercontent.com/146016/57741249-8ceac180-76ee-11e9-9dd9-7b9a26432351.png)

   It returns a response of a newly created post without attachments:

   ![image](https://user-images.githubusercontent.com/146016/57741265-a68c0900-76ee-11e9-9c07-18edd2ad4081.png)

- technically we can still fix the issue client-side as we know the ids of attachments we send to the server, se we can use them, even server doesn't return them back. But it's a hacky way and prone to inconsistency between client and server.

- Just would like to confirm, that fixing this issue on the server is the desired way to go.

vikasrohit:
Yes, go with the server side fix.

maxceem:
@PrakashDurlabhji as we need to get it fixed on the backend I'm going to increase the prize for this issue to $100 to cover the increased efforts. Let me know you are intereseted to fix this issue server-side.

maxceem:
[...]

PR #3061, opened 2019-05-27
CF17

maxceem:
PR with all the issues fixed during CF17.

Only two issues are not included yet and would be merged separately:
https://github.com/appirio-tech/connect-app/issues/3031
https://github.com/appirio-tech/connect-app/issues/2177

Though, we can already test the rest of stuff in DEV.


src/projects/detail/components/SpecQuestionList/SpecQuestionList.jsx
```diff
@@ -62,8 +64,9 @@ const SpecQuestionListItem = ({
           {!!titleAside && <div className="spec-section-title-aside">{titleAside}</div>}
           { help && (<HelpModal {...help} />) }
         </h5>}
+        {!!introduction && <div className="introduction markdown" dangerouslySetInnerHTML={{__html: markdownToHTML(introduction)}} />}
         {children && <div className="child-component">{children}</div>}
-        {!hideDescription && <p className={cn({bigger: !icon})}>{description}</p>}
+        {!hideDescription && description && <div className={cn('description', 'markdown', {bigger: !icon})} dangerouslySetInnerHTML={{__html: markdownToHTML(description)}} />}
```
vikasrohit: @maxceem I was trying to avoid using `dangerouslySetInnerHTML` for setting the rich content and that was the reason, I thought of using markdown in assumption that it won't require this. However, it seems it is not possible for us to have the rich content without `dangerouslySetInnerHTML` attribute. So, can we verify if using this attribute, have potential of script/HTML injection?

src/projects/detail/containers/ProjectInfoContainer.js
```diff
@@ -366,6 +431,14 @@ class ProjectInfoContainer extends React.Component {
     }
     links = links.concat(phaseLinks)
 
+    // extract attachment from posts
+    attachments = [
+      ...attachments,
+      ...this.extractAttachmentLinksFromPosts(feeds),
+      ...this.extractAttachmentLinksFromPosts(phaseFeeds)
+    ]
+    console.dir(attachments)
```
vikasrohit: can we get rid of this @maxceem ?

maxceem: removed in commit https://github.com/appirio-tech/connect-app/pull/3061/commits/678b94096724f9580838d604968da17364b2e6b8

src/projects/detail/components/SpecQuestionList/SpecQuestionList.jsx
maxceem: We use https://www.npmjs.com/package/remarkable package to convert `markdown` to `html` which looks robust. Though, maybe we can write some unit tests for our `markdownToHTML` method to verify, that any HTML or JS injections are not possible.

vikasrohit: Okay, please keep that in your todo list.

Issue #3065, opened 2019-05-28
Attached files stay in the create post UI


maxceem:
After we create a post with attachments and open a dialog to create another new post, the previously attached files are still listed there.

![image](https://user-images.githubusercontent.com/146016/58451960-2a5ee000-8148-11e9-8215-e1cd66bd77e1.png)

See video https://monosnap.com/file/gEntEZedkKPGW7ZqobJzdXss9M8efJ

Follow up from https://github.com/appirio-tech/connect-app/issues/3031

vikasrohit:
Working fine. Closing.

PR #3073, opened 2019-05-30
fixes issue 3065

sumitdaga:


maxceem:
@vikasrohit this is a fix for issue https://github.com/appirio-tech/connect-app/issues/3065.

In particular, it follows the same logic as for posts content:
- if say i attch a file and then click on `cancel` button the attachment should also be gone. Same like content inside post
- if there is content (meaning user has typed something in the input boxes) and we click outside ...the editor does not close. Same if we already attached files.

This PR looks good to me and can be merged to DEV.

Issue #3068, opened 2019-05-30
Attach files during project creation


maxceem:
We should be able to attach files during project creation. We already have ability to show file uploader dialog in project creation forms, but there are two things should be improved.

1. Section type `files` should be moved to question type `files` and support read optimized mode: collapsing to the accordion as other questions types. We should keep section type `files` for backward compatibility though.
2. So far we only can attach files to existent projects as we are required to set `projectId` for attachments.
     Maybe what we can do here, is to upload files to S3 during filling project creation form. 
     And when we send request to the server to create a project, we can send additional property `attachments` with list of files uploaded to S3. So create project endpoint would have to internally add file as if we added them by permission dialog. See the requests which are happen during file uploading currently to illustrate the idea https://monosnap.com/file/b3WJE4oLUsXBczxLBpr1GmrNAYi8cw

vikasrohit:
@maxceem Thanks for logging it. We can take a look at message service as well to compare the approach you suggested as we recently added support for adding attachments to the posts and they have similar behaviour where we have to attach a file before post creation. It would be great if you can evaluate both approaches and let us know your thoughts.
Overall, I am inclined towards the approach you suggested because that does not creates junk data in our database, however, at the same time it creates junk in S3 which does not have any link to track and delete if a user does not create project after uploading the file. On second thoughts, if we only upload files to temporary S3 bucket which we use for FileStack to dump the files, then we should be good as we can always empty that S3 bucket.

vikasrohit:
we need to start on this as well @maxceem

vikasrohit:
Initial commit in 84fc58943d10a21966b329808a23f88d916bf2f0

vikasrohit:
@maxceem @RishiRajSahu added more commits to clean up few console errors that were coming. Also, updated the template 104, to have the file section at the final summary screen so that we have to less worry about keeping files state between the wizard steps.
Only thing, that is pending theoretically is to have the ability of retaining the uploaded files for the incomplete projects i.e. if user closes or refreshes the browser tab on the final summary screen, they would loose the uploaded files for now.

maxceem:
@vikasrohit  I've tested it today and it works good.

The couple of things I've noticed, which I would check more precisely:
- we can drag&drop files directly to the question in which case the window to file uploading doesn't popup. I'm not sure if there are any drawbacks, but looks like there are two logics
[...]

PR #3121, opened 2019-06-18
Feature/attachments during project creation

vikasrohit:
Adding support for attaching the files during the project creation. #3068 

fyi @RishiRajSahu @maxceem @acshields


vikasrohit:
Thanks @maxceem for the review. Will try to wrap them today.

vikasrohit:
@maxceem I have just pushed and deployed changes which should fix the issues you observed. I still need to verify if my changes has broke any of the existing functionality though.

vikasrohit:
Thanks @maxceem for the review. It gives me confidence in moving ahead.

vikasrohit:
> Currently, another branch is deployed on the project service. So there was an error when I've tried to create a new project with files. Though the request looks good for me. And as it worked before I think this part also should work, when the project service has updates deployed.

Noted

Issue #3108, opened 2019-06-11
[Metadata Management] Support form, priceConfig and planConfig versioning


maxceem:
### Overview

[Recently](https://www.topcoder.com/challenges/30086308/?type=develop&noncache=true&tab=details) we've extracted `form`, `priceConfig` and `planConfig` from `ProjectTamplates` and `ProductTemplates` into separate models and implementing versioning support for them server-side.

We have to support new functionality in the **Metadata Management** tools.

Each of the new objects `Form`, `PriceConfig` and `PlanConfig` support two type of versioning.

- `version` - is something for people. When we want to make some new version of some form, we could manually click a button to create a new version. We will be aware of what exact version we are working at the moment and what version is being used inside templates.

- `revision` - is kind of history of changes or “Ctrl + Z” of changes we saved to the DB. Sometimes we accidentally save broken data into DB and finding issues could be quite hard in big JSONs. So every time we save a particular object instead of updating it we would create a copy with updated data and increased revision number. This would happen without user notice and we don’t need to know what is the number of revision we are seeing at the monent. Though a user could be able to press some button to come to the previous revision.

### Requirements

- Editing "old" `ProjectTemplates` and `ProductTemplates` which don't use separate models should work like before without changes.

   ![image](https://user-images.githubusercontent.com/146016/59256523-b9024f80-8c66-11e9-8edf-09afba019a86.png)

- If some fields like `form`, `priceConfig` and `planConfig` is defined for `ProjectTemplates` or `ProductTemplates` for such fields we would improve editing functionality and additionally show `Version` and `Revision`:

    ![image](https://user-images.githubusercontent.com/146016/59258268-38455280-8c6a-11e9-84ac-fd18f6d6a8d3.png)

### Question

If we choose some `Version` we would load the respective JSON from that version. When we save the ProjectTemplate we would update the value of the version to the one which we choose. So for example, initially 
   ```js
    projectTemplate.form = { key: 'dev', version:5 }
   ``` 
   after we choose version `3` and save it would become 
   ```js
   projectTemplate.form = { key: 'dev', version: 3 }
   ``` 
   This part is clear.

We should agree on how we would like to work with `revision`.

For example, the current revision is `14` and the form object looks like this:
```js
    projectTemplate.form = { key: 'dev', version:5 }
``` 
so `14` is the last revision.
[...]

PR #3133, opened 2019-07-02
Feature/metadata forms price plan configs

maxceem:
Implementation for issue https://github.com/appirio-tech/connect-app/issues/3108

Has been accomplished by challenge 30093918 - Topcoder Connect - Metadata - Form, PlanConfig and PriceConfig and small private task for final fixes.

Tested locally with a local deployment of Project Service and works good for me.


src/components/SelectDropdown/SelectDropdown.jsx
```diff
@@ -25,7 +25,7 @@ class SelectDropdown extends Component {
   }
 
   componentWillMount() {
-    this.updateSelectedOptionValue(this.props.value)
+    this.updateSelectedOptionValue(_.toString(this.props.value))
```
vikasrohit: @maxceem Is it really needed for this PR? I might be wrong here, but I don't want to miss the functionality of having non string as value of the `SelectDropdown` component, if it is already there.

maxceem: You were right @vikasrohit. We cannot drop support for numeric values in `SelectDropdown` as we use it in some other places. So the support for numeric values was restored and metadata form in this PR was adjusted instead. 

ref commit https://github.com/appirio-tech/connect-app/commit/1d7b3cc3a6e4d74efc5ac57e4a432d84253cd377

Issue #3128, opened 2019-06-26
[$150] Make text fields expandable in Scope form


vikasrohit:
After recent changes for new smart scope forms, we lost the ability to expand a text fields for better readability. We need to bring that back i.e. allow users to expand the text areas and by default render them with height of 3 rows.

fyi @acshields @maxceem

maxceem:
@vikasrohit hmm, the fields with type `textbox` are still expanding for me, and while `textinput` fields cannot be expanded, which I guess is expected:

![image](https://user-images.githubusercontent.com/146016/60158723-4b3c5300-9824-11e9-8a52-a86fb6fe09dd.png)

- the only thing it doesn't have 3 rows height by default, I'm not sure if they have such height by default before

vikasrohit:
I meant expanding by dragging the mouse pointer from the bottom right corner of the text box. I am not able to see that cursor any more neither when directly editing the text box during project creation nor when editing an existing project. Also, I think we did some improvement very long back where we used to show the text boxes with multiple lines to occupy the full height which fits the content by default when user is editing such fields. As of now, it just shows single line to me when I try to edit the content of the text box after expanding the accordion.

maxceem:
> I meant expanding by dragging the mouse pointer from the bottom right corner of the text box. I am not able to see that cursor any more neither when directly editing the text box during project creation nor when editing an existing project.

currently, we are using textbox component which extends the height dynamically when we are typing. Should we disable this feature and let users extend the textboxes manually instead?

> As of now, it just shows single line to me when I try to edit the content of the text box after expanding the accordion.

I guess there could be some issues, when we expand accordion and textbox doesn't update its height dynamically. Should we fix this issues instead of changing the behavior to manual?

By any chance do you have an example of project creation form where you experience such issue? To make sure we are talking about the same cases.

vikasrohit:
> currently, we are using textbox component which extends the height dynamically when we are typing. Should we disable this feature and let users extend the textboxes manually instead?

No, we should keep that behaviour too

> I guess there could be some issues, when we expand accordion and textbox doesn't update its height dynamically. Should we fix this issues instead of changing the behavior to manual?

Yes, we need to fix this issue as well i.e. the text box should occupy the full height to accommodate the text in it when user expands the accordion.

[...]

PR #3139, opened 2019-07-04
Make text fields expandable in Scope form

mfikria:


mfikria:
@maxceem here is the demo https://monosnap.com/file/AcNcnEQFdABqDnrBme6FV9oj8q6B9x
Do you have used https://github.com/appirio-tech/react-components/pull/321 ?

maxceem:
@mfikria yes, I used a new version from `react-components`, though your video looks good, I'll try to test one more time tomorrow.

maxceem:
@mfikria I've tested one more time and made sure that I use the updated code (I've added a comment to your new code and I can see it browser console), but it works with issues for me:

1. When I load page there is an error in the browser console:

    image

2. When I press `Enter` there is another error in the console:

    image

3. and after 2-3 times press `Enter` the text is truncated again, see video https://monosnap.com/file/1YL4YJgXWPaKtBI0fC0DN593zr7C3m

OS: macOS 10.14.5
Browser: Google Chrome Version 75.0.3770.100 (Official Build) (64-bit)
Also tested in: Firefox 67.0.4

mfikria:
@maxceem for the `Invariant Violation` you could try to
- remove node_modules: `rm -rf node_modules`
- revert any change to `package-lock.json`
- `npm i`

Reference: https://gist.github.com/jimfb/4faa6cbfb1ef476bd105. It works for me

mfikria:
@maxceem i 've updated the PR

mfikria:
@maxceem about the `Invariant Violation` error i think it is because the missmatch version of react/react-dom package between `connect-app` and `appirio-tech-react-components`. See explanation inhttps://gist.github.com/jimfb/4faa6cbfb1ef476bd105
Note that this error happened in local because we use `npm link`
So to fix the issue follow these steps:
1. in `connect-app` folder project: remove `node_modules`, revert any change in `package.json` and `package-lock.json`
2. in `appirio-tech/react-components` folder project: run `npm link`, remove `node_modules`, revert any change in `package.json` and `package-lock.json`
3.  in `connect-app` folder project: run `npm link appirio-tech-react-components`, run `npm i`
4.  run `connect-app` via `npm start`

mfikria:
1. `handleAccordionOpen` is called when accordion is opened. Since I did not know a way to call textarea.focus() on that function. The changes from this pr can be excluded except change to scala `src/projects/detail/components/SpecSection.scss` which made the textarea can be resized manually.

2. Since we depend on https://www.npmjs.com/package/react-textarea-autosize. There is a prop called `minRows` and it is different from the native `rows` https://www.w3schools.com/tags/att_textarea_rows.asp

maxceem:
@mfikria 

1. ok, maybe we remove it, after deciding what to do with https://github.com/appirio-tech/connect-app/issues/3128#issuecomment-509097319

2. > Since we depend on https://www.npmjs.com/package/react-textarea-autosize. There is a prop called minRows and it is different from the native rows https://www.w3schools.com/tags/att_textarea_rows.asp

Got it. I guess, then as `AutoresizeTextarea` doesn't support the `rows` prop, we can use it instead of adding new `minRows`. Like this https://monosnap.com/file/6pOClH0divJQqja7DwO1DJVf6JSO6Z. What do you think?

maxceem:
1. Regarding 1. as we are ok with the current solution as per https://github.com/appirio-tech/connect-app/issues/3128#issuecomment-509102428 I guess the extra code may be removed from the PR.

mfikria:
@maxceem seems good to me. I've updated the PRs. Thanks.

mfikria:
@maxceem well, I'll try to update the library

mfikria:
@maxceem i've updated the library. Hope this can resolve the issue

maxceem:
@mfikria, unfortunately, it didn't help. Especially it can be easily reproduced in FireFox, see demo video https://monosnap.com/file/iOeebmreBl870v73wVPYAKPZVPMKQq

The text is truncated when I quickly add lines using `Enter` or remove lines using `Backspace`, can you reproduce it?

mfikria:
@maxceem ok noted will try it in firefox

mfikria:
@maxceem how about change the library to https://www.npmjs.com/package/react-autosize-textarea ?

maxceem:
@mfikria let's give it a try. Could you please, create a separate PR with that library, just in case if the current library works better, so we can keep it.

maxceem:
Hey, @mfikria any luck with another library?

mfikria:
@maxceem no, https://www.npmjs.com/package/react-autosize-textarea and current library use the same core library which is https://www.npmjs.com/package/autosize. The alternative is rewrite the library from scratch but it need too much effort

maxceem:
Got it @mfikria. Thank you for looking into it, I think we would be fine with the way it works currently as it looks like further fixes would take too much efforts.

vikasrohit:
@maxceem could you please log this summary in the issue as well so that we can close that issue with a final note.

maxceem:
@vikasrohit summaries on the issue page and created a follow-up issue for the edge-case.

PR #3150, opened 2019-07-05
issue 3149 fix

rashmi73:


rashmi73:
@maxceem updating in sometime

rashmi73:
@maxceem this conditon is covered after a lot of analyzing

PR updated.

Also, can you please help me in understanding 
1.what preparedconditons and conditions property inside subsection array does exactly?

2. though i am aware it is to check validity of question based on conditons.

because wizardhelper.js has lot of utils, which are required to understand the flow, how properties are manipulated.

but any such document which is internally shared with your team ? which can help understand object tree well?as then it would be lot easier to understand code.

rashmi73:
@maxceem this is updated long  back. you can test it

rashmi73:
@maxceem 
Thanks for detailed explanation, i got much idea about it now.

Yes I will take this, as I feel it can be solved by me, but the modification you mentioned for https://github.com/appirio-tech/connect-app/pull/3150/files#diff-a52fca6a10cebc946297b5eb2dca6533R648

can I go with different approach and do modification in "_.mergeWith" callback function, as this is util which forms "updatedProject" object, so better i apply general check here itself and make perfectly "updatedProject", so that isEqualCheck needs no modification.

maxceem:
@rashmi73

> Yes I will take this, as I feel it can be solved by me

Sure, but let's set a deadline for this task for 48 hours as it already took too long. And this issue is quite serious and should be fixed asap. So if there is no **final fix** for 48 hours, I'll have to assign you and solve it some other way.

> can I go with different approach and do modification in "_.mergeWith" callback function, as this is util which forms "updatedProject" object, so better i apply general check here itself and make perfectly "updatedProject", so that isEqualCheck needs no modification.

I'm not sure where you want to use `mergeWith`. But I'm fine with any fix if it's **general** and works for **any** form data not only for this particular form.

I just want to note, if you want to use `mergeWith` inside `wizardHelpers` methods, you can fix the current form data which we edit. But you still have to fix data which comes from the server, as on the server we already have forms with any kind of data. Also, note, that methods inside `wizardHelpers` are kind of sensetive and easy to broke, as they are also used for project [creation forms](https://connect.topcoder-dev.com/new-project/) and logic here is quite complicated.

So IMHO its easier and safer to fix comparison at this line https://github.com/appirio-tech/connect-app/pull/3150/files#diff-a52fca6a10cebc946297b5eb2dca6533R648. But if you feel confident about your approach I'm also fine with it as long as it works for all forms data and doesn't break existent functionality.

rashmi73:
@maxceem can i get something for my efforts till now in finding root cause?

also many things are already given as hint by me(though it is hardcoded), but now i am in doubt whether i can solve it completely because even if i go with your approach 
1.how to determine whether the property needs to be omitted?
2. by looking at prepared conditions ? or conditions? 

Hence I wish to opt out from here as I do not wish to waste time if I could not solve it.

Thanks

rashmi73:
@maxceem i guess i have a solution, working on it

maxceem:
@rashmi73 as there is no solution yet and we don't have time anymore to give it a try so I have to un-assigning you.

> @maxceem can i get something for my efforts till now in finding root cause?
> also many things are already given as hint by me(though it is hardcoded)

Generally, we only accept and grant payments for solved issues. If the issue is not solved it cannot be paid.
As your finding are still valuable, as an exception I may process half payment for the initial issue `$30`. I understand that it may be not covers your efforts, so I would recommend to give up the issue next time if you feel that you cannot find a solution in a reasonable time to not spend your efforts without a reward.

Thanks.

PR #3153, opened 2019-07-06
issue 3051 fix

PrakashDurlabhji:


PrakashDurlabhji:
@maxceem 
1.can you please confirm if "isOpen" was ever working in NotificationDropDownContainer?
because suddenly there is an issue.
2. also, if route "notification/seen" exists?

maxceem:
@PrakashDurlabhji

> 1.can you please confirm if "isOpen" was ever working in NotificationDropDownContainer?
because suddenly there is an issue.

You may check the current version on https://connect.topcoder-dev.com/ - I don't see any issues with `isOpen`.

> 2. also, if route "notification/seen" exists?

Do you mean the API URL to mark notifications `seen`? The url is `https://api.topcoder-dev.com/v5/notifications/${id}/seen`, where `${id}` - is id of the notification. We already have a service method for it https://github.com/appirio-tech/connect-app/blob/dev/src/routes/notifications/services/notifications.js#L20

src/components/NotificationsDropdown/NotificationsDropdownContainer.jsx
```diff
@@ -122,6 +131,7 @@ const NotificationsDropdownContainerView = (props) => {
               onToggle={(isOpen) => {
```
PrakashDurlabhji: @maxceem ideally this callback shall receive isOpen as a boolean, but it recevies as object, may be an event object kind of, in which there is no info about the open/close state of dropdown.

in current functionality, isOpen here is always an object, but by default in parent container it is made false i guess and then toggle happens in parent when we click on bell, due to which functionality works properly, though container opens/close, but it is not due to the boolean here.

which is hence not allowing me to call "markAllNotificationsSeen"

src/routes/notifications/services/notifications.js
```diff
@@ -16,9 +16,13 @@ const markNotificationsRead = (id) => {
   }
 }
 
-// the id can be either: notification id; or '-' separated ids, e.g. '123-456-789'
+// the id can be either: null/undefined (mark all); notification id; or '-' separated ids, e.g. '123-456-789'
 const markNotificationsSeen = (id) => {
-  return axios.put(`${TC_NOTIFICATION_URL}/${id}/seen`)
+  if (id) {
```
PrakashDurlabhji: @maxceem also the implementation done is here is similar to read flow,
1. i guess on read flow, there two routes exists, "/read" and "/${id}/read"
2. whereas at backend in seen flow, there is only one route, "/${id}/seen".

3. hence on making a request to mark all seen ("/seen") on opening notification dropdown, it gives 404 that no such route exists.

I will check in tc-project-service now, how read flow is implemented.

maxceem: Got your question. Yeah, for the `seen` we don't have a route for marking all notifications as we have for read, see https://github.com/topcoder-platform/tc-notifications/blob/dev/src/routes.js

Though, instead of one id in `"/${id}/seen"` we can pass a list of ids like `/123-456-789/seen` - this would mark notifications with ids `123`, `456` and `789` as `seen`.

src/components/NotificationsDropdown/NotificationsDropdownContainer.jsx
maxceem: @PrakashDurlabhji I see, do you think there is anything that can be done here?

src/routes/notifications/services/notifications.js
PrakashDurlabhji: @maxceem shall i raise PR at tc-notifications? as it would be easier to put a method  in Notificationcontroller, else as you wish i can even added - separated notification ids.

src/components/NotificationsDropdown/NotificationsDropdownContainer.jsx
PrakashDurlabhji: @maxceem yesterday I did try to put some logic in parent and try it but failed to find any such flow.
thinking if notificationbell component can be converted to stateful component and make a callback via props.

src/routes/notifications/services/notifications.js
maxceem: @PrakashDurlabhji I would prefer to use the current API, if you don't see some serious drawbacks there.

src/components/NotificationsDropdown/NotificationsDropdownContainer.jsx
maxceem: I guess you can make component `NotificationsDropdown` stateful, so it keeps the `isOpen` status.

Note, there are TWO ways which call `onToggle` method:
1. From `EnhancedDropdown` https://github.com/appirio-tech/connect-app/blob/dev/src/components/NotificationsDropdown/NotificationsDropdown.jsx#L15 in this way we get the correct value for `isOpen`, see https://github.com/appirio-tech/connect-app/blob/dev/src/components/NotificationsDropdown/EnhancedDropdown.jsx#L33
2. When we click the bell https://github.com/appirio-tech/connect-app/blob/dev/src/components/NotificationsDropdown/NotificationsDropdown.jsx#L20
     There we call `onClick={onToggle}` so `onToggle` will have `event` for the first argument instead of `isOpen`.

Let me know what you think.

PrakashDurlabhji:
@maxceem update done

PrakashDurlabhji:
@maxceem i making a commit, well tested and red cue disappears perfectly on opening notifications by clicking on bell.
just 10 mins more

maxceem:
@PrakashDurlabhji good, I'll review as soon as you post an update.

PrakashDurlabhji:
@maxceem commit updated. kindly verify thank you

src/routes/notifications/services/notifications.js
```diff
@@ -16,9 +16,13 @@ const markNotificationsRead = (id) => {
...
+    return axios.put(`${TC_NOTIFICATION_URL}/${id}/seen`)
+  } else {
+    return axios.put(`${TC_NOTIFICATION_URL}/seen`)
```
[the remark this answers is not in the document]
PrakashDurlabhji: done

src/components/NotificationsDropdown/NotificationsDropdownContainer.jsx
```diff
@@ -70,7 +73,7 @@ const NotificationsDropdownContainerView = (props) => {
       }, 0)
     }
   }
-  const hasNew = hasUnread && lastVisited < _.maxBy(_.map(notifications, n => new Date(n.date)))
+  const hasNew =  hasUnseen
```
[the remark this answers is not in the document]
PrakashDurlabhji: done

src/components/NotificationsDropdown/NotificationsDropdown.jsx
```diff
@@ -9,25 +9,47 @@ import EnhancedDropdown from './EnhancedDropdown'
...
+
+    this.toggle = this.toggle.bind(this)
+  }
+
+  toggle(isOpen) {
+    if (typeof isOpen === 'object') {
+      this.props.onToggle(!this.state.isOpen)
+      this.setState({ isOpen: !this.state.isOpen})
+    } else {
+      this.props.onToggle(isOpen)
```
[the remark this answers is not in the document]
PrakashDurlabhji: done

PrakashDurlabhji:
@maxceem all updated

maxceem:
@PrakashDurlabhji one lint error is left, also `VISIT_NOTIFICATIONS` is not removed completely.

image

PR #3172, opened 2019-07-13
Cf18 part 2

maxceem:
Fix for the last issues from CF18


maxceem:
Merged this PR by myself, as there is only one issue and it's related to the Metadata admin interface.

Issue #3149, opened 2019-07-05
"Save Changes" stays enabled if we don't have any change in scope form


maxceem:
### Expected behavior
"Save Changes" should get disabled again if we don't have any changes in the scope form. And we shouldn't get an alert popup when we try to leave the scope tab.

### Actual behavior
"Save Changes" stays enabled if we don't have any change in scope form. And we get an alert popup when we try to leave the scope tab.

### Steps to reproduce the problem
- Open scope tab https://connect.topcoder-dev.com/projects/8197/scope
- Change some field to make `Save Changes` button enabled.
- Revert the change - I mean change back as for has been initially.
- Button supposed to get disabled as there are no changes in the form but it stays enabled.

### Screenshot/screencast
https://monosnap.com/file/mka4az6YPGuZsoL5zgpgjETe6oGmC8

### Additional notes

FYI for some form it works as expected, example: https://connect.topcoder-dev.com/projects/8178/scope, see [demo video](https://monosnap.com/file/oBreGfR6TTtAHpY02Eb1bJHrxoHLjV)

--

#### Environment
- OS: Max OS
- Browser (w/version): Google Chrome
- User role (client, copilot or manager): all
- Account used: pshah_manager

rashmi73:
give me thanks

maxceem:
This issue is now open for pick up.

maxceem:
Ok @rashmi73 assigned to you.

maxceem:
Updated link to the demo project with the issue https://connect.topcoder-dev.com/projects/8197/scope

rashmi73:
@maxceem PR ready https://github.com/appirio-tech/connect-app/pull/3150

maxceem:
@rashmi73 please, have a look on the review here https://github.com/appirio-tech/connect-app/pull/3150#pullrequestreview-258586832

maxceem:
Hey, @rashmi73 do you think you can finish it during next 24 hours?

rashmi73:
@maxceem actually backend response is not having consistent model properties.

I mean even if is blank or null, still object/key-value pair shall be received,
I analyzed that response was not proper in one case whereas it was proper for other hence i added fields, but if response becomes consistent then it would be more helpful?

maxceem:
> I mean even if is blank or null, still object/key-value pair shall be received,
I analyzed that response was not proper in one case whereas it was proper for other hence i added fields, but if response becomes consistent then it would be more helpful?

[...]

PR #3175, opened 2019-07-14
Disable "Save Changes" button again if there is no changes.

gets0ul:
#3149

### DIFF
--- a/src/projects/detail/components/EditProjectForm/EditProjectForm.jsx
+++ b/src/projects/detail/components/EditProjectForm/EditProjectForm.jsx
@@ -23,6 +23,7 @@ import {
   STEP_VISIBILITY,
   STEP_STATE,
 } from '../../../../helpers/wizardHelper'
+import { clean } from '../../../../helpers/utils'
 
 import './EditProjectForm.scss'
 
@@ -151,6 +152,17 @@ class EditProjectForm extends Component {
           template: updatedTemplate,
           project: hidedSomeNodes ? nextProps.project : this.state.project,
         })
+
+        // re-check again if any hidden values when an option is deselected
+        const updatedProject = clean(removeValuesOfHiddenNodes(updatedTemplate, nextProps.project))
+        const skipProperties = ['members', 'invites']
+        const clearUpdatedProject = clean(_.omit(updatedProject, [...skipProperties, 'isDirty']))
+        const clearUpdatedNonDirtyProject = clean(_.omit(nextProps.projectNonDirty, skipProperties))
+        const isDirty = !_.isEqual(clearUpdatedProject, clearUpdatedNonDirtyProject)
+        // update the state, always use this flag to check if changed
+        this.setState({
+          isProjectDirty: isDirty,
+        })
       }
     }
   }
@@ -206,7 +218,7 @@ class EditProjectForm extends Component {
   }
 
   isChanged() {
-    return !!this.props.project.isDirty
+    return !!this.state.isProjectDirty
   }
 
   enableButton() {
--- a/src/projects/reducers/project.js
+++ b/src/projects/reducers/project.js
@@ -669,6 +669,20 @@ export const projectState = function (state=initialState, action) {
           }
           return srcValue
         }
+
+        // reset the values when deselected
+        if (key === 'deploymentTargets') {
+          if (!action.payload.deliverables) {
+            srcValue = []
+          }
+          return srcValue
+        }
+        if (key === 'progressiveResponsive') {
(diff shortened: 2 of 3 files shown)


gets0ul:
@maxceem 
Thanks for the feedback.
I'll let you know if I can comeback with a general fix.

gets0ul:
@maxceem 
We can remove out the all changes with specified fields, and leave only general part from my PR.
It will work.
The only problem is the default initial value of an option when they are selected and deselected.

From https://github.com/appirio-tech/connect-app/pull/3150#pullrequestreview-259442296
> The main reason for the issue is because we can show/hide them. So when we hide some filed, we don't have it's value at all in the project details - it's undefined. But when we show it, the field would be shown, but if we don't choose any value, then value would be **empty string, empty object or an empty array**. So the difference is if field is visible at the moment or no, but still it doesn't have value.

It's easy if the value were **empty string, empty object or an empty array**. It's already taken care by general fix of my code.

The problem here if the value is none of above.
For example, https://connect.topcoder-dev.com/projects/8186/scope
The type of budget. The current value is "Design".
If we add "Development", it will set the Development Units field `development.qty = 1` in `budgetDetails`
"Save Changes" button is enabled.
Now if we deselect the "Development", it won't disabled the "Save Changes" button again as the "development.qty = 1" remains there in `budgetDetails`. Need to manually check if `budgetDetails.deliverables` array no longer contain 'dev'.

Same as https://connect.topcoder-dev.com/projects/8197/scope
If we changes the "Devices" to "Web Browser" we need to choose the "Web App Type" from radiobutton groups  which will set the `progressiveResponse` field in the dirty project.
If we change our mind and revert back the "Devices" to "Mobile", the field `progressiveResponse` is still there so it stays dirty.

For the first example, it's easier to solve as it requires the discipline of the app user.
Project 8186. we can clear the Development Units field and deselect Development.
The "Save Changes" button will be disabled again.

But it is not that easy for the 2nd example as it came from radio button value and there is no way to reset/deselect it.

Having said that, I will try to look if a general fix that covering all cases is possible.

gets0ul:
@maxceem 
PR updated without using hardcoded fields.  
Thanks.

src/projects/detail/components/EditProjectForm/EditProjectForm.jsx
```diff
@@ -206,7 +218,7 @@ class EditProjectForm extends Component {
   }
 
   isChanged() {
-    return !!this.props.project.isDirty
+    return !!this.state.isProjectDirty
```
maxceem: I guess when we are making this change, we don't use `isDirty` which we update inside Redus store, so I have a couple of questions:
1. Should we still have and update `isDirty` in the Redux store?
2. Do you think it's safe to use `isProjectDirty` instead of `isDirty` from the Redus store?

gets0ul:
@maxceem 

That's the correct behavior.
Selecting `Deployment` requires you to choose the `Deployment Targets`
The same case as `Web Browser` with the 'Web App Type`

The save button won't be enabled until we enter the value,

src/projects/detail/components/EditProjectForm/EditProjectForm.jsx
gets0ul: `isDirty` from Redux is still used.
That is where the initial value of `isProjectDirty` come from.

Both will always have same value most of the time.
The only time they can be different is for the cases I already mention here https://github.com/appirio-tech/connect-app/pull/3175#issuecomment-511349937

When a node is updated and the hidden value is still there.
When an option is selected and the default value is set.

`isDirty` will stays `true` when that option is deselected because of that hidden default value.
That's the only time, the value of `isProjectDirty` will be different.

I think it safe as we have already been using the state `isProjectDirty` when creating `SpecSection` component to render the questions.

maxceem:
Sorry for the delay with review @gets0ul. Today we are releasing the new version of Connect app. Will review this PR tomorrow.

src/projects/reducers/project.js
```diff
@@ -631,13 +632,18 @@ export const projectState = function (state=initialState, action) {
         if (key === 'screens' || key === 'features' || key === 'capabilities') {
           return srcValue// srcValue contains the changed values from action payload
         }
+
+        // project's name might contain ampersand
+        if (key === 'name') {
+          return _.escape(srcValue)
+        }
```
maxceem: @gets0ul I'm not sure what is the reason for adding this code, is there any situation when it helps?

gets0ul: @maxceem 

It is for the case like this https://connect.topcoder-dev.com/projects/7977/scope
when a project contains ampersand on it's name.

PR #3185, opened 2019-07-17
[HOTFIX] [PROD] post release 2.4.13

maxceem:


src/components/AssetsLibrary/SubFolder.jsx
```diff
@@ -56,6 +56,7 @@ class SubFolder extends React.Component {
   render() {
     const { link, renderLink, goBack, formatModifyDate } = this.props
     const { linkToDelete } = this.state
+    const isLinkSubFolder = !this.props.onDeletePostAttachment
```
vikasrohit: @maxceem Didn't understand the need of this change.

src/components/UserSummary/UserSummary.jsx
```diff
@@ -19,9 +12,6 @@ const UserSummary = ({user}) => {
     drafts: 7,
     delivered: 5
   } */
-  const powerUserRoles = [ROLE_CONNECT_COPILOT, ROLE_CONNECT_MANAGER, ROLE_ADMINISTRATOR, ROLE_CONNECT_ADMIN]
-  const isCustomer = _.intersection(user.roles, powerUserRoles).length === 0
-  const role = isCustomer ? 'Customer' : 'Member'
```
vikasrohit: @maxceem just a note, it was showing me as `Customer` even when I have power user roles.

src/components/AssetsLibrary/SubFolder.jsx
maxceem: @vikasrohit we have to know inside subfolder if this is subfolder to render links or files, to properly set icons. I've update this code to not use the hacky way with `onDeletePostAttachment` and added additional prop.

src/components/UserSummary/UserSummary.jsx
maxceem: @vikasrohit I guess it was an issue with logic. If I'm not mistaken we didn't have direct requirements from the product team on what text should be shown here. So I've simplified logic to always show `User since` for now.

PR #3198, opened 2019-07-19
Hotfix/post release 2.4.13.3

vikasrohit:


maxceem:
@vikasrohit do you have any project example where you faced this issue?

maxceem:
I've checked the [package-lock.json history](https://github.com/appirio-tech/connect-app/commits/dev/package-lock.json). The `draft-js` version has been always `0.10.5` there (since the long time ago). So I think this issue is not because of the `draft-js` versions update. I guess there was some edge-case which we didn't come across before. So any steps to reproduce the issue would be helpful.

image

vikasrohit:
@maxceem are you not seeing the issue for any of the projects in dev env?

maxceem:
@vikasrohit no. I've checked out the commit before your fixes https://github.com/appirio-tech/connect-app/commit/f7bfa86fc9e54c167a947107efad44c88dfa99ec locally.

And tried to create topics and create/edit/delete posts in a couple of projects. But all works without issues:
http://local.topcoder-dev.com:3000/projects/8219/messages/
http://local.topcoder-dev.com:3000/projects/7442/messages/

Also, I cannot reproduce it on DEV (after the fixes):
https://connect.topcoder-dev.com/projects/8219/messages/
https://connect.topcoder-dev.com/projects/7442/messages/

vikasrohit:
@maxceem I have create a post http://connect.topcoder-dev.com/projects/7442/messages/32745 which ideally should be able to reproduce the issue because I copied pasted the content (obfuscated the text leaving the formatting as is) of the message from production which probably caused the issue.

maxceem:
Thanks, @vikasrohit. I've checked it out before your fix is applied, but the message works to me without issues in the browser console.

image

Maybe you can share with me this message in the server response to make sure that we keep the exact code of the message as during copy/pasting something may be changed. So I could create an exact copy of the message but submitting a direct request to the server to create the same post. It's ok to obfuscate the text.

vikasrohit:
@maxceem check this http://local.topcoder-dev.com:3000/projects/7442/messages/32753 Now it exhibits the same error the message which caused this is `>`. And it works if I put `block.text = ''` at https://github.com/appirio-tech/connect-app/blob/dev/src/helpers/markdownToState.js#L290 and remove the previous patch https://github.com/appirio-tech/connect-app/blob/dev/src/helpers/markdownToState.js#L298 (filtering the blocks)

maxceem:
Thanks, @vikasrohit that helped a lot. Your solution with setting `block.text = ''` appears to be the right way to go here. I've created a new HOTFIX https://github.com/appirio-tech/connect-app/pull/3217 where I also reversed the previous hotfix changes, as we don't need them anymore after a new fix.

PR #3217, opened 2019-07-25
Hotfix/post release 2.4.13.7

maxceem:
## fix rendering DraftJS message when block doesn't have inline content

Fix for the issue with rendering `>` in DraftJS.

The root of the issue is that in code we create a block first, and after when we find `inline` content inside that block we set some properties of the block like `text`. But if there is no inline content, then such properties of the block stay unutilized which causes an error later as a block is not fully defined.

So the solution is set the default values of properties for the block as it has empty inline content. And if there is some inline content, such properties would be updated.

### DIFF
--- a/src/helpers/markdownToState.js
+++ b/src/helpers/markdownToState.js
@@ -1,5 +1,4 @@
 import {convertFromRaw} from 'draft-js'
-import _ from 'lodash'
 const Remarkable = require('remarkable')
 
 // Block level items, key is Remarkable's key for them, value returned is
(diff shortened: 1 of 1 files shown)


vikasrohit:
@maxceem the strange thing is that why this issue started to appear only after our recent release. The message which has this problem was created on Jun 24, so it should have appeared before our release as well.

vikasrohit:
Also, could you please log a ticket and update your investigation results there.

maxceem:
> @maxceem the strange thing is that why this issue started to appear only after our recent release. The message which has this problem was created on Jun 24, so it should have appeared before our release as well.

That is strange, but so far I don't see any reason how it could happen only after the release but not before.

> Also, could you please log a ticket and update your investigation results there.

Created https://github.com/appirio-tech/connect-app/issues/3219

PR #3232, opened 2019-07-30
Issue 3231 fix

Gauravseta:


src/helpers/templates.js
```diff
@@ -155,9 +155,8 @@ export function getProjectProductTemplates(productTemplates, projectTemplates, p
   }
 
   const projectTemplate = getProjectTemplateById(projectTemplates, project.templateId)
-  const phases = _.get(projectTemplate, 'phases', {}) // TODO we can log error details here for missing project template
+  const phases = projectTemplate.phases ? projectTemplate.phases : {} // TODO we can log error details here for missing project template
```
maxceem: @Gauravseta After this change, it may create an error if `projectTemplate` is not defined.  Your new code looks like it does the same as before. Are you sure we need this change?

src/projects/detail/ProjectDetail.jsx
```diff
@@ -67,8 +67,7 @@ const ProjectDetailView = (props) => {
   if (!currentMemberRole && props.currentUserRoles && props.currentUserRoles.length > 0) {
     currentMemberRole = props.currentUserRoles[0]
   }
-
-  const template = _.get(props.projectTemplate, 'scope', {})
+  const template = buildTemplate(props.projectTemplate, props.forms, props.planConfigs, props.priceConfigs).scope
```
maxceem: @Gauravseta to be safe here, can we keep the code before using `_.get`:
```js
const builtTemplate = buildTemplate(props.projectTemplate, props.forms, props.planConfigs, props.priceConfigs)
_.get(builtTemplate, 'scope', {})
```

src/projects/detail/containers/DashboardContainer.jsx
```diff
@@ -144,12 +144,14 @@ class DashboardContainer extends React.Component {
...
+      priceConfigs,
+      forms,
+      planConfigs
     } = this.props
     const projectTemplate = project && project.templateId && projectTemplates ? (getProjectTemplateById(projectTemplates, project.templateId)) : null
-
     let template
     if (project.version === 'v3') {
-      template = _.get(projectTemplate, 'scope')
+      template = buildTemplate(projectTemplate, forms, planConfigs, priceConfigs).scope
```
maxceem: @Gauravseta same here, let keep it safe and use get:
```js
const builtTeamplate = buildTemplate(projectTemplate, forms, planConfigs, priceConfigs)
template = _.get(builtTeamplate, 'scope')
```

src/helpers/templates.js
Gauravseta: @maxceem it is purposely made like this, because:

1.  phases is coming null, i checked it, so what happens with previous code is "_.get" of lodash is not returning "{}" Object, but it is making phases as null, so in below lines phases is going across for loop and it breaks.

2. so i added check if .phases present then take phases else return object.

conclusion: _.get of lodash not working well here dont know why, I guess "null" is considered a valid value by _.get , hence phases is null. 
so if phases is undefined then _.get returns default object.

src/projects/detail/ProjectDetail.jsx
Gauravseta: sure will do it

src/projects/detail/containers/DashboardContainer.jsx
Gauravseta: sure will do it

src/helpers/templates.js
maxceem: I see. Then I guess we can do it like this:
```js
const phases = projectTemplate && projectTemplate.phases ? projectTemplate.phases : {}
```

So we also have to check that `projectTemplate` is defined. Otherwise when we check `projectTemplate.phases` and `projectTemplate` is not defined, such code will produce error.

maxceem: Thanks for update @Gauravseta. Only this one left, and I'll do the final check.

Gauravseta: done

maxceem:
@Gauravseta there are two lint errors:

image

Issue #3228, opened 2019-07-30
Need to show option value's label/title instead of their values


vikasrohit:
### Expected behavior
It should show the text of the question's option (e.g. title or label) when rendering the change log in `ScopeChangeRequest` component.

### Actual behavior
It is showing the `value` of the options

### Steps to reproduce the problem
- Open any project in `reviewed` or above status
- Go to the scope tab
- Change couple of fields like (Screens and/or Devices)
- Submit the change request
- Observe the change log rendered by the component

### Screenshot/screencast
<img width="804" alt="Screen Shot 2019-07-30 at 12 02 23 PM" src="https://user-images.githubusercontent.com/2417632/62106092-ed3ad980-b2c1-11e9-988d-750fe8c882b0.png">

--

#### Environment
- OS: Any
- Browser (w/version): Any
- User role (client, copilot or manager): Admin to test the flow with single login
- Account used: N/A

vikasrohit:
Working fine @vigneshTheDev. Thanks.

PR #3233, opened 2019-07-30
fix issue #3228

vigneshTheDev:
show value labels instead of values in scope change request.


vigneshTheDev:
@vikasrohit Found that this branch is a bit outdated. Please ignore this pull request. Can you please create a new branch from latest dev for working on scope change related issues?

vikasrohit:
@vigneshTheDev Just to confirm, we don't need this action now, right?

vigneshTheDev:
@vikasrohit Yes, we don't need this now.

Issue #3230, opened 2019-07-30
Array fields are not being merged correctly after Approving the scope change request


vikasrohit:
### Expected behavior
When removing some options of a multi select question while creating scope change request, it should remove those options when `Approving` and `Activating` a scope change request.

### Actual behavior
When removing some options of a multi select question while creating scope change request, it not actually removing those options when `Approving` and `Activating` a scope change request.

### Steps to reproduce the problem
- Go to any project which is in `reviewed` or above statuses, please ensure that the `Target Devices` question has more than one devices selected before moving to the `reviewed` status
- Go to scope tab
- Remove at least one device from the `Target Devices` question
- Hit `Submit Change Request` button to create the request
- Hit `Approve` and then hit `Activate` buttons
- Observe the `Target Devices` question, it still contains originally set devices (i.e. devices set before making the scope change request)

### Screenshot/screencast
N/A

--

#### Environment
- OS: Any
- Browser (w/version): Any
- User role (client, copilot or manager): Admin (to test the flow with same login)
- Account used: N/A

vikasrohit:
Working fine. Thanks @vigneshTheDev

vikasrohit:
@vigneshTheDev I just noticed a new issue that appeared after this fix. I have logged it [here](https://github.com/appirio-tech/connect-app/issues/3251) .

Issue #3229, opened 2019-07-30
After request creation, scope form resets but the summary labels of the fields does not


vikasrohit:
### Expected behavior
Summary labels of the fields in the scope form should also reset themselves, after successful creation of a scope change request.

### Actual behavior
While the fields are reseting themselves to previous state (i.e. without the changes of the scope change request), the summary labels of the fields are still showing the updated values.

### Steps to reproduce the problem
- Go to any project which is in `reviewed` or above statuses
- Go to scope tab
- Edit couple of scope affecting fields e.g. Screens and/or Devices
- Hit `Submit Change Request` button to create the request
- Now observe that the fields are reset to original value but summary labels are not

### Screenshot/screencast
<img width="696" alt="Screen Shot 2019-07-30 at 12 22 04 PM" src="https://user-images.githubusercontent.com/2417632/62107183-ac908f80-b2c4-11e9-9759-263c26169f31.png">
<img width="796" alt="Screen Shot 2019-07-30 at 12 22 08 PM" src="https://user-images.githubusercontent.com/2417632/62107190-b0bcad00-b2c4-11e9-9b46-588d5dd127fb.png">

--

#### Environment
- OS: Any
- Browser (w/version): Any
- User role (client, copilot or manager): Any
- Account used: N/A

vikasrohit:
Working fine. Thanks @vigneshTheDev

PR #3242, opened 2019-08-01
Fix bugs related to scope change request

vigneshTheDev:


src/projects/detail/components/ScopeChangeRequest/ScopeChangeRequest.jsx
```diff
@@ -57,6 +57,41 @@ class ScopeChangeRequest extends React.Component {
...
+  /*
+    Gets the readable labels (E.g. "Banking or Financial Services") for the values (E.g. finserv)
+    using the field key (E.g. details.appDefinition.mobilityTestingType)
+  */
+  deduceValueLabel(template, invertedScopeFieldPaths, fieldkey, value) {
+    const key = invertedScopeFieldPaths[`details.${fieldkey}`].replace(/\.fieldName$/, '')
+    const fieldDescriptionObject = _.get(template, key)
+    const fieldType = fieldDescriptionObject.type
+
+    const questionTypesWithOptions = [
```
vikasrohit: @vigneshTheDev is there any way to fix the inconsistent summary labels without referencing the questions types here? I mean it would require us to update this list on every addition or removal or form fields.

@maxceem Do you think you can suggest some way here?

maxceem: @vikasrohit @vigneshTheDev if I understand this correctly, this is the fix for the issue https://github.com/appirio-tech/connect-app/issues/3229 right?

So the issue is about that summary label is not updated when we revert the question value. The summary label is formatted by the Accordion component and it already has logic to format the summary label.

What is the reason for repeating the logic to format the summary label in this place? If we update the value of the question, the summary label supposes to be re-rendered by the Accordion component.

vigneshTheDev:
Hi @maxceem, acutally, this is the fix for #3228. 


This PR has fixes for 3 issues (#3228, #3229, #3230). Though, I have created 3 different commits.

src/projects/detail/components/ScopeChangeRequest/ScopeChangeRequest.jsx
vikasrohit: > If we update the value of the question, the summary label supposes to be re-rendered by the Accordion component.

Yes, that is my concern as well @maxceem @vigneshTheDev

vikasrohit: Got it @vigneshTheDev. This change is for finding the actual label of the option. I guess, we can validate if there exists `options` field for the question and then assign either `option.label` or `option.title` as label for rendering in ScopeChangeRequest.

maxceem: Ok, got it. Then I think there is no other way to build such labels for answers rather than per question type.

We already have this functionality inside Accordion component: 
- https://github.com/appirio-tech/connect-app/blob/feature/workstreams-timeline/src/projects/detail/components/Accordion/Accordion.jsx#L126-L148
- https://github.com/appirio-tech/connect-app/blob/feature/workstreams-timeline/src/projects/detail/components/Accordion/Accordion.jsx#L35-L62 

So I think the best approach here would be to move out these format methods from Accordion component into separate helper file. And reuse them from there to format summary label inside Accordion and in other places like this one https://github.com/appirio-tech/connect-app/issues/3228

So later, when we add new question types, we add formatting rules for them only in one place.

vikasrohit: Thanks @vigneshTheDev for the latest commit where you have removed the array of question types.

@maxceem I think here we can skip the formatting logic of value, which we are using in Accordion, because we are just rendering the change in the answer for a question and I think it would make sense to show the title or label of the field changed as is instead of formatting it like we do in Accordion. So, I guess at present we don't have multiple places where we are formatting the answers of a question, however, it might be good idea to move out that logic to helper function some day.

PR #3260, opened 2019-08-13
Reusable PostsContainer (phases comments regactoring)

maxceem:
Phases comments component refactoring into a reusable PostsContainer component which can be used to show posts for any topic, not only for phases.

The main purpose is to show comments in new Work UI, but I propose to merge it separately earlier, as it doesn't have breaking changes in the current functionality only refactoring. So we can test it more for regression issues on DEV.

This was implemented by the challenge https://www.topcoder.com/challenges/30098185/?type=develop.

The next (optional) step is to refactor Messages page (Drawer with comments) to also use this component instead of having its own separate logic. This would reduce code duplication and would make future support easier.


vikasrohit:
@maxceem I think we can get the optional part done as well via private task, with the winner of this challenge. It would be speed up the changes.

maxceem:
@vikasrohit we don't have to make it now and can leave it after 2.5 release. I'll check after how much work it would require. And we can decide later on if the challenge or private task is better.

In general, I prefer a challenge, as the review is done by three people: 2 reviewers and me. This is to reduce the chance of regression issues. In case of a private task only me is testing, and if there are a lot of changes - there are more chances that I would miss some issues.

vikasrohit:
Got it. Feel free to launch challenge after the release.

PR #3265, opened 2019-08-14
adding temporary condition to avoid old notifications missing

sachin-maheshwari:
…eploying topic name change.

### DIFF
--- a/src/routes/notifications/helpers/notifications.js
+++ b/src/routes/notifications/helpers/notifications.js
@@ -324,6 +324,16 @@ const getNotificationRule = (notification) => {
   const notificationRule = _.find(NOTIFICATION_RULES, (_notificationRule) => {
     let match = _notificationRule.eventType === notification.eventType
 
+    // For supporting old topic name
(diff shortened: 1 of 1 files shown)


vikasrohit:
Thanks @maxceem for the detailed review.

sachin-maheshwari:
@maxceem FYI, the error you shared are coming before merging as well.

maxceem:
Right @sachin-maheshwari. We have warnings to show that some notifications have types which we don't know how to process in Connect. After the notification types have been renamed in configs but not in DB we have many warnings. Because types in configs were updated, but for existent notifications in DB we have old types.

After the fix in this PR, most of warnings disappear because after this fix by new logic we already can find rules for notifications from the DB, but they are not correct.

So I've added new warnings in the screenshot above. They show which notifications rules we associate to which notifications types. So it can be seen, that we associate notification rules with type which don't match notification types.

vikasrohit:
Agreed with @maxceem Reverting the PR.

PR #3272, opened 2019-08-20
fix #3263

### LINKED ISSUE #3263: Connect Registration/Profile Page Improvements
- [x] Mark Required Fields https://github.com/appirio-tech/connect-app/issues/3262
- [x] Split First Name & Last Name into 2 Fields

- On the **registration** page for Connect, there should now be two fields for name: First Name and Last Name.
- On the **profile** page for Connect, there should now be two fields for name: First Name and Last Name.

- [x] Update Business Number in Profile Settings

- Update 'Business phone' on the **profile** page in Connect to match the **registration** page, where the requirement for country code is called out and the dropdown of codes by country is available.

- [x] Country List Alphabetization

- Ensure the **registration** page's country list in the dropdown menu is in descending alphabetical order
- Ensure the **profile** page's country list in the dropdown menu is in descending alphabetical order

- [ ] Country List Responsive Typing

- Enable responsive typing/selection for country selection on the **registration** page.
- Enable responsive typing/selection for country selection on the **profile** page.

vigneshTheDev:


src/routes/settings/routes/profile/components/ProfileSettingsForm.jsx
```diff
@@ -109,9 +142,34 @@ class ProfileSettingsForm extends Component {
             uploadPhoto={this.props.uploadPhoto}
           />
         </div>
-        {this.getField('First and last name', 'firstNLastName', true)}
+        {this.getField('First Name', 'firstName', true)}
```
RishiRajSahu: @vigneshTheDev I think we have to put check in a way so that we would be able to load old profiles without error which had`firstNLastName`  instead of divided fields ?

vigneshTheDev: Hi @RishiRajSahu, Thanks for the review. This is already taken care of in this helper file: https://github.com/appirio-tech/connect-app/pull/3272/files#diff-1774c038214d5fe34f0f59dd4be77130L56-L61. The `data` variable mentioned here is the initial input to the form. 
In old profiles also, we're deriving firstNLastName from firstName & lastName fields. Now, I'm directly taking firstName and lastName instead of combining them into firstNLastName. So, the data object mentioned above will not have firstNLastName field at all (even for old profiles).

RishiRajSahu: Okay, we are reading firstname and lastname from basic_info trait so we are good.

PR #3309, opened 2019-09-19
issue 3215 fix

PrakashDurlabhji:


PrakashDurlabhji:
@maxceem it means I need to then calculate position of tooltip from top dynamically ? so that it is shown perfectly as intend to? let me give a try or  else I will comment immediately to release it in few hours

maxceem:
> @maxceem it means I need to then calculate position of tooltip from top dynamically ? so that it is shown perfectly as intend to?

Kind of yes. Though I think you don't have to write the actual calculation math. We use `react-popper` which can make all the calculations. So probably we only have to configure it some way, not sure. I also found this package https://www.npmjs.com/package/react-popper-tooltip which implements the Tooltip using the same library `react-popper` as we do. And in that implementation Tooltip is placed in the `body` of DOM and it's no cut in any situation.

So there are two ways I see:
1. Use `react-popper-tooltip` to implement our Tooltip component https://github.com/appirio-tech/react-components/tree/feature/connectv2/components/Tooltip.
2. Or we can read how `react-popper-tooltip` is implemented and in a similar way fix our component https://github.com/appirio-tech/react-components/tree/feature/connectv2/components/Tooltip.

In any way we should update our `Tooltip` component so it works same as now. Because we already use in many places, so we shouldn't break anything, just fix this issue with cutting.

maxceem:
As this PR has solution which breaks other functionality, I'm closing it.

Issue #3213, opened 2019-07-23
[$70] Messaging on mobile


gondzo:
Composing a message in mobile is broken (at least on android). Can't use the editor at all - ex type "thank you." and the text displayed is "t ." and there is no way to delete that text either. I'd suggest just using the raw textarea in mobile

vikasrohit:
Thanks @gondzo it seems with new UI changes, it started to render the draft js editor for mobiles as well which was not intended despite the fact that we didn't work on mobile designs for the new UI.

fyi @maxceem

vikasrohit:
@gondzo I can see that it is still rendering the plain text editor instead of draft js editor for me. It works fine when trying to create new thread, however, when creating new post, it seems like the the new post modal is hidden behind the drawer and the drawer itself is having horizontal scroll.

![Screenshot_2019-08-26-11-16-28](https://user-images.githubusercontent.com/2417632/63668052-8abaf780-c7f3-11e9-9e6c-95a0145e3c13.png)

maxceem:
Sum up:

1. Drawer should open with 100% instead of hardcoded value on the screens with width less than [$screen-md](https://github.com/appirio-tech/tc-ui/blob/feature/connectv2/src/styles/_variables.scss).

   - It should also start from the top of screen without space on the top.

2. When we click `Write post` we open the window to write a post. It should open on TOP of the drawer. Now it's opened, but behind the drawer so we cannot write a new post on mobile device.

3. In general we have to make sure that we can create new topics and posts on the `Messages` page, see [example project](https://connect.topcoder-dev.com/projects/8363/messages).

suppermancool:
@maxceem pls assign this to me

maxceem:
Done @suppermancool

suppermancool:
@maxceem done in https://github.com/appirio-tech/connect-app/pull/3313

RishiRajSahu:
@lakshmiathreya could you please test this issue on PROD and confirm ?

vikasrohit:
Verified on production and it seems to working for me. Closing.

PR #3313, opened 2019-09-19
Messaging on mobile #3213

suppermancool:
Messaging on mobile #3213

### DIFF
--- a/src/components/TopicDrawer/TopicDrawer.jsx
+++ b/src/components/TopicDrawer/TopicDrawer.jsx
@@ -4,7 +4,7 @@ import Drawer from 'appirio-tech-react-components/components/Drawer/Drawer'
 
 import FeedContainer from '../../projects/detail/containers/FeedContainer'
 
-require('./TopicDrawer.scss')
+import './TopicDrawer.scss'
 
 /**
  * A drawer that shows the posts under the selected topic
@@ -20,7 +20,7 @@ const TopicDrawer = ({
   return (
     <Drawer
       open={open}
-      className="drawer-container"
+      styleName="drawer-container"
       containerStyle={{
         top: '60px',
         height: 'calc(100% - 60px)',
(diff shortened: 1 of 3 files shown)


src/components/TopicDrawer/TopicDrawer.jsx
```diff
@@ -18,6 +20,7 @@ const TopicDrawer = ({
   return (
     <Drawer
       open={open}
+      className="drawer-container"
```
maxceem: Can we use `CSS Modules` instead of regular CSS to avoid CSS names collisions?

See example https://github.com/appirio-tech/connect-app/blob/dev/src/components/CardListHeader/CardListHeader.jsx#L18

src/components/TopicDrawer/TopicDrawer.scss
```diff
@@ -0,0 +1,13 @@
+@import '~tc-ui/src/styles/tc-includes';
+
+:global {
```
maxceem: As mentioned above, if we use CSS Modules, we don't have to use `:global` to avoid CSS classnames collisions.

src/components/TopicDrawer/TopicDrawer.jsx
```diff
@@ -4,6 +4,8 @@ import Drawer from 'appirio-tech-react-components/components/Drawer/Drawer'
 
 import FeedContainer from '../../projects/detail/containers/FeedContainer'
 
+require('./TopicDrawer.scss')
```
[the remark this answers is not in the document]
suppermancool: @maxceem done in latest commit

src/components/TopicDrawer/TopicDrawer.jsx
suppermancool: @maxceem done in latest commit

src/components/TopicDrawer/TopicDrawer.scss
suppermancool: @maxceem done in latest commit

Issue #3225, opened 2019-07-26
[$45] 403 in suggesting users to the Customers while adding members to the team


vikasrohit:
Member API allows only specific roles to access the `_search` endpoint which essentially means a regular customer (whose roles is not in the list of allowed roles) can never access the `_search` endpoint. So, every time a customer tries to add a member to their project's team, we get `403` error in console.
We should avoid calling this API when customer is trying to add members to the team.

vikasrohit:
Sentry error https://sentry.io/organizations/topcoder/issues/1000936941/events/65073870372/

vikasrohit:
fyi @maxceem

maxceem:
I guess you meant endpoint `_suggest` like `https://api.topcoder-dev.com/v3/members/_suggest/phash_` which we use for getting suggestions during user invitations. It returns error `403` for customers.

When I'm trying to use a privilege user (`pshah_manager`), I'm also getting error from the server on `DEV`:

- Request: `GET https://api.topcoder-dev.com/v3/members/_suggest/pshah`
- Response:

   ![image](https://user-images.githubusercontent.com/146016/64601674-0c02b480-d3c6-11e9-9afe-c787a2dd761e.png)

This still works good on PROD when I use my account as a `copilot`.

> We should avoid calling this API when customer is trying to add members to the team.

Which API should we call when customers are trying to invite members? Or should we disable suggestions for customers?

vikasrohit:
Not sure about why the `suggest` is not working in dev. We have security patch for `_search` endpoint though but that was not restricting any one to call the endpoint rather it was cleaning up the fields based on the roles. Good to know that it is still functional on production though.

For the purpose of this issue, we should just disable the suggestions for the customers as they are actually not getting any suggestion even today. But please remembers, we should restrict the suggestion based on logged in user role and should not just disable the suggestion in the customer team management dialog.

maxceem:
Sum up:
1. When customer user is typing a user handle to invite we should not try to show suggestions as we always get error `403` see a [screenshot](https://monosnap.com/file/wa6RBpugf1aR0dMzX7rXZMSR0gSnHS). Though customers should be still able to type username fully with the option `Create "<userhandle>"` as it's done now.

   For demo you can use project http://local.topcoder-dev.com:3000/projects/7442/ with user `pshah_customer`. Click `Manage` link next to `Team` header on the left sidebar:
   
[...]

PR #3314, opened 2019-09-19
403 in suggesting users to the Customers: Issue #3225

suppermancool:
…team #3225

403 in suggesting users to the Customers while adding members to the team #3225


suppermancool:
@maxceem this make sense, done in the latest commit

Issue #3211, opened 2019-07-23
[$50] Avoid reloading file icons in the Assets Library


maxceem:
### Expected behavior
Icons for files should be loaded only once.

### Actual behavior
Every time when we hide/show some folder on the **Assets Library** page, the icons for files are reloading and blinking.

### Steps to reproduce the problem
- Open any project's **Assets Library** page with some files uploaded, for example, https://connect.topcoder-dev.com/projects/7977/assets
- Observe that icons for files are loaded.
- Open dev tools to watch network calls. 
- Switch to the **Links** tab and switch back to the files tab.
- Observe that icons for files are loaded again in the Network tab of the dev tools. And also visually we can see, that icons are not shown initially when we switch back to the files tab.

### Screenshot/screencast
Demo video: https://monosnap.com/file/MDgmQKiY0VUCUCukXtNqnURA1EZLsc

![image](https://user-images.githubusercontent.com/146016/61688511-cdeafc00-ad57-11e9-97d9-f15077494e8a.png)

--

#### Environment
- OS: any
- Browser (w/version): any
- User role (client, copilot or manager): any
- Account used: `pshah_manager`

vikasrohit:
Good catch @maxceem

suppermancool:
@maxceem pls assign this to me

suppermancool:
@maxceem done in https://github.com/appirio-tech/connect-app/pull/3316

vikasrohit:
Release to production.

PR #3316, opened 2019-09-19
Avoid reloading file icons in the Assets Library #3211

suppermancool:
Avoid reloading file icons in the Assets Library #3211

### DIFF
--- a/src/components/FileIcon.jsx
+++ b/src/components/FileIcon.jsx
@@ -1,7 +1,7 @@
 /**
  * FileIcon component
  *
- * Renders proper project type icon depend on the project type
+ * Renders file icon depend on the file type
  */
 import React from 'react'
 import PT from 'prop-types'
--- a/src/components/FileIcon.jsx
+++ b/src/components/FileIcon.jsx
@@ -1,0 +1,230 @@
+/**
+ * FileIcon component
+ *
+ * Renders proper project type icon depend on the project type
+ */
+import React from 'react'
+import PT from 'prop-types'
+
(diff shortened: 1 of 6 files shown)


suppermancool:
@maxceem actually all images are cached by browser i'm not sure why it didn't work on you pc. I will try implement the way `project type icons` does (it loaded all icon into code, this is not the good way i think).
I think this error will happen for all the image in the app and not just the icon in Assets Library and i don't think we can implement the way of `project type icons` for all the image.

In my site the browser work good and it cached all loaded image:
https://drive.google.com/file/d/11_zByr_hQlLdPa8GL7et1CPbojM60Ebm/view?usp=sharing

maxceem:
> I think this error will happen for all the image in the app and not just the icon in Assets Library and i don't think we can implement the way of project type icons for all the image.

Yeah, agree, though I think it's ok for the images which are "content" to be loaded during page changing. But the images which are the part of UI is better to keep always loaded so the whole UI is shown instantly.

suppermancool:
ah actually this should be the way reactjs does, it alway import the image before using it, i almost forget this 🤦‍♂️

suppermancool:

src/components/FileIcon.jsx
```diff
@@ -0,0 +1,230 @@
+/**
+ * FileIcon component
+ *
+ * Renders proper project type icon depend on the project type
```
maxceem: comment left from another component

suppermancool: @maxceem all done in the latest commit

Issue #3179, opened 2019-07-16
[$75] [Metadata Management] keep entered data if error occurs during entity updating


maxceem:
This is follow up from this comment https://github.com/appirio-tech/connect-app/issues/3132#issuecomment-511789890 from the similar issue which we fixed during object creation https://github.com/appirio-tech/connect-app/issues/3132

This issue is valid for all sections in the metadata: Project Templates, Products Templates, Project Types and so on...

### Expected behaviour
When we edit any object changed data should persist in the form when a server-side error happens during saving.

### Actual behaviour
Updated data is reseted in the form when a server-side error happens during saving.

### Steps to reproduce the problem
- Open any page to edit object in the metadata section, for example editing the next Project Template https://connect.topcoder-dev.com/metadata/projectTemplates/207.
- Update any field, for example change name to `test`
- Click `Save` button. You would see an error message from the server as the user doesn't have permissions to update the object. The same time the name would be reseted to initial value `Subtitle Skills API`.
- Expected behaviour is that we should see our changed value when error happens, so we can repeat our attempt to save changed data and don't loose our changes.

### Screenshot/screencast
Video: https://monosnap.com/file/IQTgWRjBKxv3JFMd8Qn6gkfcXR3XjU

--

#### Environment
- OS: all
- Browser (w/version): all
- User role (client, copilot or manager): all

suppermancool:
@maxceem done in https://github.com/appirio-tech/connect-app/pull/3318 .

vikasrohit:
Working as per requirements.

PR #3318, opened 2019-09-19
Eeep entered data if error occurs during entity updating Issues #3179

suppermancool:
keep entered data if error occurs during entity updating #3179


suppermancool:

PR #3319, opened 2019-09-19
Issue 3155 fix

rashmi73:


maxceem:
@rashmi73 current PR doesn't fix the issue. Would be able to fix it or should I open this issue for pick up?

rashmi73:
@maxceem sorry for delay I would be able to fix it

rashmi73:
@maxceem can we clear everything after successful invites? so it would just endup to making a single line change and not much difficulty

rashmi73:
@maxceem PR update

src/components/TeamManagement/ProjectManagementDialog.js
```diff
@@ -26,6 +26,7 @@ class ProjectManagementDialog extends React.Component {
     if (processingInvites && !nextProps.processingInvites ) {
       const notInvitedSelectedMembers = _.reject(selectedMembers, (selectedMember) => (
         this.isSelectedMemberAlreadyInvited(nextProps.projectTeamInvites, selectedMember)
+        || this.isSelectedMemberAlreadyInvited(nextProps.topcoderTeamInvites, selectedMember)
```
maxceem: @rashmi73 is this change required for the fix?

src/helpers/utils.js
```diff
@@ -53,9 +53,12 @@ export const compareEmail = (email1, email2, options = { UNIQUE_GMAIL_VALIDATION
  *
  * @returns {Boolean} true if two user handles are same
  */
-export const compareHandles = (handle1, handle2) => (
-  (handle1 || '').toLowerCase() === (handle2 || '').toLowerCase()
-)
+export const compareHandles = (handle1, handle2) => {
+  if (handle2.indexOf('@') === 0) {
+    return (handle1 || '').toLowerCase() === (handle2.substring(1) || '').toLowerCase()
```
maxceem: @rashmi73 as before we have logic like this `(handle2 || '')` it looks like sometimes `handle2` can be undefined. To be sure that this wouldn't cause error in runtime it's better to apply `.substring(1)` not to `handle2` directly, but first, we make sure that we have string, like this:
```js
(handle2 || '').substring(1).toLowerCase()
```

src/components/TeamManagement/ProjectManagementDialog.js
rashmi73: @maxceem yes because may be he is not a customer, because i saw parent component where props are declared, invitest filtered by 'customer' so it may fall into another array? which is topcoderTeamInvites.

src/helpers/utils.js
rashmi73: @maxceem sure updating now

rashmi73:
@maxceem updated PR

src/helpers/utils.js
```diff
@@ -53,9 +53,17 @@ export const compareEmail = (email1, email2, options = { UNIQUE_GMAIL_VALIDATION
...
+export const compareHandles = (handle1, handle2) => {
+  let h1 = handle1
+  let h2 = handle2
+  if (handle2.indexOf('@') === 0) {
+    h2 = (handle2 || '').substring(1)
+  }
+  if (handle1.indexOf('@') === 0) {
+    h1 = (handle1 || '').substring(1)
+  }
+  return h1.toLowerCase() === h2.toLowerCase()
```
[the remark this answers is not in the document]
rashmi73: @maxceem updated

src/components/TeamManagement/ProjectManagementDialog.js
rashmi73: @maxceem updated

rashmi73:
@maxceem PR update. Hope it works

rashmi73:
@maxceem done and updated

Issue #3273, opened 2019-08-21
[$150] Show Created By for files and links list in Assets Library


vikasrohit:
We should add `Created By` column for the files and links list in Assets Library

vikasrohit:
fyi @maxceem

maxceem:
We can show it for all files, but at the moment we don't have information who created links for the project. We can still show who created links inside posts, as we know who created a post:

![image](https://user-images.githubusercontent.com/146016/63419698-d5292680-c437-11e9-99c6-80946ab9e05c.png)

Should we update the client-side code to store the information about who created links inside project data? 

Also, if we don't write some additional logic server-side, anybody could update this data in the project data. I don't think it's sensitive information, but if we want to make sure that nobody can fake `createdBy` we should validate project data server-side.

vikasrohit:
Got it. We are storing the links in a JSON field not as a primary object so we don't have the audit fields there. I think, we should start capturing the audit fields (createdBy, createdAt, updatedBy and updatedAt) inside this JSON for every link to show the details about owner of the link.
I didn't get the validation part though. How can one fake `createdBy`? Oh, I think you mean that with `PATCH` request for the project, if we send the bookmarks field as well, it would allow any privileged user to fake the data. Yes, I think we need to do some changes in code for fixing that.

maxceem:
If we we would store `createdBy` for links, should we restrict editing links only to users who has created such link as we do for files? If so, should admin still be able to edit any links no matter who created them?

maxceem:
Sum up:
1. Currently, when we are creating links we store them inside `project.bookmarks` JSON in the next format:
    
    ![image](https://user-images.githubusercontent.com/146016/64605052-fa241000-d3cb-11e9-91d1-46e9fa64286c.png)

   Additionally, we should add the next properties for each link:
   - `createdAt` - creation date in format `2019-09-05T05:52:21.240Z`
   - `createdBy` - id of the user who created the link in format `22688955` (number)
   - `updatedAt` - should be updated when we `create` and `update` links
   - `updatedBy` - should be updated when we `create` and `update` links

[...]

PR #3320, opened 2019-09-19
bug #3273 Show Created By for files and links list in Assets Library

jankyzhang:


maxceem:
@jankyzhang there is some issue with the Member Service. It's not connected with your PR, but I cannot test it properly. So I'm waiting for the member service to come back live.

jankyzhang:
@maxceem 
Sorry for the delay reply.
I miss the E-mail notification.
I check it right now.

jankyzhang:
@maxceem,  
I have fix two issues now.
Please check the code.

About the issue 2:
In the 'AssetsInfoContainer.jsx' . the 'props.projectMembers' is accroding of the state 'members.members' ,which get from the existent action 'loadMembers'.
If we use the action 'loadMembers' again for 'assetsMembers'.
The  'props.projectMembers' may be changed.
So I implement a new action to load information about user members at first.

jankyzhang:
@maxceem ,
Thanks for the reply.
Now the Service is fail.
I will check and fix the issue today later.
screenshot-connect topcoder-dev com-2019 09 29-22_52_06

jankyzhang:
@maxceem 
screenshot-local topcoder-dev com_3000-2019 09 30-01_09_35
Website try to get the user-info about '305384'.
However, just a blank content return. 
So the code request to the server again and again, due to the missUserId '305384'.

How we should do about this ?

maxceem:
> Website try to get the user-info about '305384'.
> However, just a blank content return.
> So the code request to the server again and again, due to the missUserId '305384'.

> How we should do about this ?

We should request users only once. If, like in this situation we didn't get info about some user, we can show `Unknown` as you do now. Note, that you can check how we do on Project Listing page. After loading the list of projects, we request all the users once. If we cannot get some, we don't show them or show `unknown`.
So your approach is good, just have to request users one time only.

jankyzhang:
@maxceem ,
I just update the code.
Please check it.
Thanks.

Issue #3180, opened 2019-07-16
[$45] [Metadata Management] after any error happens, navigating to any other page shows error


maxceem:
This is follow up from this comment https://github.com/appirio-tech/connect-app/issues/3132#issuecomment-511789890

If there is an error in any of screen, going back to the listing page by navigating the navigation menu, it still shows the error page i.e. it does not clears up the error from Redux state.

Have to create steps to reproduce it.

maxceem:
### Steps to reproduce
1. Use user `pshah_manager`
2. Open any object for editing, for example this Project Template https://connect.topcoder-dev.com/metadata/projectTemplates/219
3. Click **Save** button.
4. As this user doesn't have permissions to update objects the error would happen on the server as expected and it would be shown in the bottom left corner.
5. Now try to navigate to any section in the left menu. Instead of the list of object we would see error page:

    ![image](https://user-images.githubusercontent.com/146016/64675148-1f238c00-d47b-11e9-897e-13a4cecda69f.png)

### Requirements
If some error happens and after we navigate to another page we shouldn't see error page. This should be handled in a general way for any kind of error.

vikasrohit:
Good catch. I was trying to log this issue since the day we implemented the metadata UI, but was not able to get time to log it.

nahidshahin:
@maxceem can I try please?

maxceem:
Hey @nahidshahin as per our Bug Bash rules, you don't have to wait to be assigned and can work on the issue as soon as you comment, see the rules https://www.topcoder.com/challenges/30102919.

So since your comment the issue was assigned to you. As during 12 hours since start you didn't provide a Pull Request and didn't requested the extension, I have to unassign this issue from you and open it for everyone.

This issue now open for pick up.

gets0ul:
Please assign me.

maxceem:
Done @gets0ul

gets0ul:
@maxceem PR is done and ready for review.

vikasrohit:
Working fine.

PR #3321, opened 2019-09-20
Fix for issue #3180

gets0ul:
#3180  

Error message has been already shown by alert popup.
Error page will be only shown for error code 500 (as hard-coded for the CoderBot).
https://github.com/appirio-tech/connect-app/blob/b2022930027fa7e0e97cd1a42e0373ed0b70cdb3/src/routes/metadata/containers/ProjectTemplatesContainer.jsx#L117

### DIFF
--- a/src/reducers/templates.js
+++ b/src/reducers/templates.js
@@ -165,14 +165,14 @@ export default function(state = initialState, action) {
     return {
       ...state,
       isLoading: false,
-      error: action.payload.response.data.result.content.message
+      error: action.payload.response.data.result.status >= 500
     }
   case UPDATE_PROJECTS_METADATA_FAILURE:
     Alert.error(`PROJECT METADATA UPDATE FAILED: ${action.payload.response.data.result.content.message}`)
     return {
       ...state,
       isLoading: false,
-      error: action.payload.response.data.result.content.message
+      error: action.payload.response.data.result.status >= 500
     }
   case REMOVE_PROJECTS_METADATA_FAILURE:
   case REMOVE_PRODUCT_CATEGORY_FAILURE:
@@ -183,7 +183,7 @@ export default function(state = initialState, action) {
     return {
       ...state,
       isRemoving: false,
-      error: action.payload.response.data.result.content.message
+      error: action.payload.response.data.result.status >= 500
     }
   case ADD_PROJECTS_METADATA_SUCCESS:
   case CREATE_PROJECT_TEMPLATE_SUCCESS:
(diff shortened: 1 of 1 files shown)


gets0ul:
Yes the error is is not needed and can be removed from the store as it is not used anywhere.   
The GridView component which is this error ends up to, ignores it completely.   

>  I cannot test it properly as I cannot trigger error with code 500 though it looks for me that the main issue is not resolved.

You don't need to trigger the 500. I mean, main issue is whenever error occurs when dealing with metadata management, we can't go back to the any page as error page will be shown (because of that error in the redux)

And my fix already solves that. Using the example project, with phsah_manager account, error is shown by popup, and we are now able to switch to other page as required by the issue spec.

> If during create/update/delete we would get error with code 500 and after we switch to another page, we would still see the error page as was described in the issue #3180

Yes that's why I don't remove the error from the store for 500 error:
` error: action.payload.response.data.result.status >= 500`

maxceem:
> And my fix already solves that. Using the example project, with phsah_manager account, error is shown by popup, and we are now able to switch to other page as required by the issue spec.

Right, it work because now error which we get has code 403. But imagine if we update some object and there is error on server with code 500. In such case if we switch the page after, we would see error page. But we shouldn't see error page after we switch page.

For example, for testing I can update your code to catch errors from code 400:
```js
error: action.payload.response.data.result.status >= 400
```
Now if I try to save any object I will get an error, and if I switch the page, I would still see error page. But I shouldn't see error page, because this error happens before and shouldn't break other pages.

gets0ul:
> we should reset it when we change page.

So we will be doing it instead. Right?

maxceem:
> So we will be doing it instead. Right?

We should either reset it when we change page.
Or if we don't use this error anywhere, which happens during create/update/delete, we just can remove saving it to the Redux Store in these actions. But we have to make sure, that we don't use it. So if we don't set it at all, we don't have to reset it after.

gets0ul:
It is not needed.

Lets take Project Template metadata management as example.
https://github.com/appirio-tech/connect-app/blob/23ad36e35de1e8ff389dc5e83f83ebfda434153d/src/routes/metadata/containers/ProjectTemplatesContainer.jsx#L83

The `error` is used when creating the `ProjectTemplatesGridView`, which will delegate it again to `GridView` component:
https://github.com/appirio-tech/connect-app/blob/23ad36e35de1e8ff389dc5e83f83ebfda434153d/src/routes/metadata/components/ProjectTemplatesGridView.jsx#L156

The `GridView` ignores this passed `error` prop completely. So in the end, it is not needed at all.

gets0ul:
Yes, though we would want it to be set to false as template grid view marks the prop as required.

https://github.com/appirio-tech/connect-app/blob/23ad36e35de1e8ff389dc5e83f83ebfda434153d/src/routes/metadata/components/ProjectTemplatesGridView.jsx#L166

I will submit new commit.
Thanks.

gets0ul:
Yeah, I checked again. We do not need to as it has default `false` value in initial state. I will update the PR.  
Thanks.

Issue #3215, opened 2019-07-24
[$100] Tooltip is cut inside dropdown select


maxceem:
### Expected behavior
Tooltip should be fully visible and go beyond the dropdown select.

### Actual behavior
Tooltip cannot go beyond the dropdown select and is cut.

### Steps to reproduce the problem
- Open any project with a phase which is not started yet, for example https://connect.topcoder-dev.com/projects/8219
- Click `edit` icon for the phase.
- Open the select for the **Status** and hover mouse to the disabled **Delivered** option.
- Observe that tooltip is shown, but it cannot go beyond the select dropdown list and it's cut.

### Screenshot/screencast
![image](https://user-images.githubusercontent.com/146016/61780613-70c57800-ae35-11e9-904f-c95373ab0186.png)

--

#### Environment
- OS: any
- Browser (w/version): any
- User role (client, copilot or manager): manager
- Account used: `pshah_manager`

maxceem:
Sum up:
- this issue has to be fixed for ALL instances of the select component which is used in this form, not only for this particular form

PrakashDurlabhji:
@maxceem I am starting on this. thanks

PrakashDurlabhji:
@maxceem unable to solve it, please unassign me

PrakashDurlabhji:
@maxceem dont unassign me it is done raising PR

PrakashDurlabhji:
@maxceem please give it back to me it is done now

maxceem:
Thanks for letting me know @PrakashDurlabhji.

This issue is open for pickup.

maxceem:
Ok @PrakashDurlabhji assigned you back!

PrakashDurlabhji:
@maxceem PR ready https://github.com/appirio-tech/connect-app/pull/3309

PrakashDurlabhji:
@maxceem releasing this, it is open now

nahidshahin:
@maxceem want to work on it, please.

maxceem:
Thanks for update @PrakashDurlabhji 

@nahidshahin it's on you. Please, check the comments here about the possible implementation for this issue https://github.com/appirio-tech/connect-app/pull/3309

maxceem:
As there is no update during more than 12 hours, this issue is now open for pick up.

maxceem:
This issue may be harder than it looked initially, please, check the comments here about the possible implementation for this issue #3309

So the prize for this issue is raised to `$100`.

[...]

PR #3332, opened 2019-09-23
Fix for issue #3215

gets0ul:
#3215 

Add usePortal=true flag to the tooltip in dropdown's option to indicate that the tooltip body will be appended to the body of DOM instead of the target.

react-portal is added in https://github.com/appirio-tech/react-components/pull/335


gets0ul:
Done.

Issue #3311, opened 2019-09-19
[$120] P3: Update Invited Registered User Workflow


ajefts:
**Objective:** Update email templates sent to registered platform users when they are invited to a project to include 'Join Project' or 'Decline' buttons, which when clicked automatically manage their acceptance or decline of invitation in the platform. This will improve the efficiency of the workflow that exists today, wherein: An invited registered user receives an invitation email, they click the invitation link in the email, they are brought into Connect and see a pop-up for Join/Decline, and have to click on their choice in the platform.

<p></p>

**Requirements** - Workflow: 

[Imagining that users has already completed the steps of sending invitation/receiving invitation]

- Registered user receives project invitation email

- Registered user sees button options of 'Join Project' or 'Decline' in the email body

  - **Option A:** Registered user clicks 'Join Project.' User should be automatically joined to the project.

  - **Option B:** Registered user clicks 'Decline.' User should not be joined to the project and should not see the project listed on their 'all projects' dashboard on next login.

<p></p>

ajefts:
Link to feature: https://tcprod.productboard.com/feature-board/planning/features/2761674

maxceem:
It looks like there are two parts in this issue to be done:
1. In the Connect App create links which users can follow to automatically accept and decline invitation. I'm thinking about creating the next links:
    ```
    // link to automatically accept invitation
    http://local.topcoder-dev.com:3000/projects/8369?invitation=accept

    // link to automatically decline invitation
    http://local.topcoder-dev.com:3000/projects/8369?invitation=decline
    ```
    
    So if user follows such a link we should show the loading indicator during we are sending request to the server to accept/decline the invitation.
     - If invitation is accepted we start loading the project (continue showing loader).
     - if invitation is declined we can redirect to the project listing page
     - if user is already a member of the project no matter which link we use, we just show the project
     - if user is not a member, but there is an error during invitation accept or decline we should show the error page (with error from the server)
    
    @RishiRajSahu @vikasrohit what do you think about security risks with such links?

2. After we implement the first part, we can update email template by adding these two links and button to accept and decline invitations.

RishiRajSahu:
@maxceem I don't think there should be any security risk by creating such URLs to accept/decline the invitation, because as per my understanding we update the PENDING invites present for that user for a particular project, so even if those params are changed manually and there is no PENDING invite present for that project/member combination then there will be an error.

maxceem:
[...]

PR #3335, opened 2019-09-24
Update Invited Registered User Workflow #3311

suppermancool:
Update Invited Registered User Workflow #3311


src/projects/detail/ProjectDetail.jsx
```diff
@@ -112,18 +114,26 @@ const EnhancedProjectDetailView = spinner(errorHandler(ProjectDetailView))
...
+    this.shouldForceCallAcceptRefuseRequest = this.shouldForceCallAcceptRefuseRequest.bind(this)
   }
 
   componentWillMount() {
     const projectId = this.props.match.params.projectId
     this.props.loadProjectDashboard(projectId)
+
+    setTimeout(() => {
+      this.props.acceptOrRefuseInviteFail({code:403, message: 'messageGenerator'})
+    }, 10000)
```
maxceem: @suppermancool can we avoid using timeout? It feels like such a solution culd fail sometimes if we need to wait more than `10` seconds.

src/projects/actions/projectMember.js
```diff
@@ -167,6 +168,23 @@ export function acceptOrRefuseInvite(projectId, item) {
...
+ */
+export function acceptOrRefuseInviteFail(error) {
+  const errorPromise = new Promise(() => {
+    throw new Error(error)
+  })
+
+  return (dispatch) => {
+    return dispatch({
+      type: ACCEPT_OR_REFUSE_INVITE_FAIL,
+      payload: errorPromise
```
maxceem: @suppermancool as we don't make any asynchronous action here, we shouldn't use Promise here and we can have only one action `ACCEPT_OR_REFUSE_INVITE_FAIL`. We don't need other actions with suffixes `_PENDING `, `_FAILURE` and `_SUCCESS` as this action has only one result. So we can just use `ACCEPT_OR_REFUSE_INVITE_FAIL` directly everywhere.

Let me know if there is some need for the current implementation of this action.

suppermancool: @maxceem done in the latest commit

src/projects/detail/ProjectDetail.jsx
suppermancool: @maxceem done in the latest commit

suppermancool:
@maxceem all done in the latest commit

src/projects/actions/projectMember.js
```diff
@@ -167,6 +168,16 @@ export function acceptOrRefuseInvite(projectId, item) {
   }
 }
 
+/**
+ * Accept or refuse invite request fail
+ * @param {Object} error error object
+ */
+export function acceptOrRefuseInviteFail(error) {
+  return (dispatch) => {
+    return dispatch({ type: ACCEPT_OR_REFUSE_INVITE_FAILURE, payload: error })
```
maxceem: @suppermancool I see here you removed the promise and also you renamed the event name from `ACCEPT_OR_REFUSE_INVITE_FAIL` to `ACCEPT_OR_REFUSE_INVITE_FAILURE`. Just would like to note, that this action `ACCEPT_OR_REFUSE_INVITE_FAILURE` is already used for  `acceptOrRefuseInvite` method. I'm not sure if it's intentional to use the same action name or it could lead to some issues.

This is just FYI.

suppermancool: @maxceem done in the latest commit

suppermancool:

Issue #3310, opened 2019-09-19
P3: Update Invited Unregistered 'Join Project' Workflow


ajefts:
**Objective:** Remove secondary step of pop-up window with 'Join Project/Cancel' that appears after the user has already clicked 'Join Project' in 'all projects' dashboard. 

<p></p>

**Requirements** - Workflow: 

[Imagining that users has already completed the steps of sending invitation/receiving invitation/registering (if necessary), and that they are now in Connect on the 'all projects' dashboard]

- User sees the project they have been invited to on their 'all projects' dashboard.

- User sees two action options: 'Join Project' as a primary button, or 'decline' listed in subscript underneath the 'Join Project' button.

  - **Option A:** User clicks 'Join Project' and they are immediately added to the project. 

  - **Option B:** User clicks 'Decline' and the project is removed from their 'all projects' dashboard.

<p></p>

In the workflow described above, the secondary reconfirmation of 'Join Project' pop-up window should not appear.

ajefts:
Link to feature: https://tcprod.productboard.com/feature-board/planning/features/2761599

RishiRajSahu:
@maxceem if this could be added to bug bash ?

maxceem:
For testing the next project can be used http://local.topcoder-dev.com:3000/projects/8369. See [demo video](https://monosnap.com/file/cJFLHSZIX21Z8xESSYxuots1PHtFcD) on how I use user `pshah_manger` to invite user `pshah_customer` and after `pshah_customer` can see `Join project` button on the project listing page.

So as per task description above we should do the next things:
1. Add `decline` blue link under the `Join project` button on the project listing page (in both `grid` and `list` view).
2. When we click `Join Project` button we should redirected to the project details page, show loader indicator while we are sending request to the server to accept invitation. As soon as invitation is accepted we can start loading project. See how it's done when we click `Join Project` button in the dialog on the project details page.
3. When we click `Decline` we should hide the project in the project list, while sending request to the server to decline invitation. If error happens during invitation declining we should show red error alert in the bottom right corner (reuse Alert) with error message and show the project back in the project listing.

On the second thoughts I think we should implement this issue after https://github.com/appirio-tech/connect-app/issues/3311

RishiRajSahu:
@maxceem all correct except - 
> When we click Join Project button we should redirected to the project details page,
we don't redirect user to project details page instead hide the Join/Decline button/link once the action is done.

[...]

PR #3337, opened 2019-09-24
Issue 3310 fix

rashmi73:


src/projects/components/projectsCard/ProjectCard.jsx
```diff
@@ -40,13 +41,31 @@ function ProjectCard({ project, duration, disabled, currentUser, history, onChan
...
+              </button>
+              <div 
+                onClick={(e) => {
+                  e.stopPropagation()
+                  onUserInviteAction(true, project.id)
+                }}
+                className="join-btn" style={{margin: '5px'}}
+              >
                 Join project
               </div>
```
maxceem: @rashmi73 it's better to use tag `button` instead of the `div` for the `Join project` button, same like you do for `Decline`.

src/projects/list/components/Projects/Projects.jsx
```diff
@@ -75,6 +78,18 @@ class Projects extends Component {
...
+  onUserInviteAction(isJoining, projectId) {
+    console.log(isJoining)
+    console.log(projectId)
+    this.props.acceptOrRefuseInvite(projectId, {
+      userId: this.props.currentUser.userId,
+      email: this.props.currentUser.email,
+      status: isJoining ? PROJECT_MEMBER_INVITE_STATUS_ACCEPTED : PROJECT_MEMBER_INVITE_STATUS_REFUSED
+    }).then(() => {
+      
+    })
```
maxceem: @rashmi73 looks like we don't use `then()` here and can remove it.

rashmi73:
@maxceem 

**After we declined the project, it should disappear from the list**

I shall get status as "rejected" when receiving from backend. but i get status as "pending" from backend. hence buttons are always seen.

same applies when we joing project.

**When we click Join project button and request is successful, we have to hide join/decline buttons after. But now they are not hided after joining,**

because status still remains "pending" when json object from backend is recevied. status should change either to "accepted"/"rejected" whichever applicable but it is not changing hence the issue

rashmi73:
@maxceem 
**When we click Decline we should only show loader indicator for the current project where we click the button, not for all projects**

will it be ok to convert stateless component to stateful? so that I can make property true/false and accordingly show loader for respective component.

src/projects/list/components/Projects/Projects.jsx
rashmi73: @maxceem its usage may be require now or I will remove it if not in use.

src/projects/components/projectsCard/ProjectCard.jsx
rashmi73: @maxceem It will be converted to button

maxceem:
> will it be ok to convert stateless component to stateful? so that I can make property true/false and accordingly show loader for respective component.

If we don't have enough information in Redux Store, I think it's ok convert component to stateful.

rashmi73:
@maxceem update PR but still work in progress. waiting for your comment

rashmi73:
@maxceem sorry for delay updating immediately

rashmi73:
@maxceem @RishiRajSahu everything is addressed now except loading icon in grid view.

1.it will show and hide join button perfectly based on clicks thereitself even before refresh.
2.Decline button can be seen in grid view and functioning well.

3.but genuinely dont know how to apply logic to show loader in 
src/projects/list/components/Projects/ProjectsGridView.jsx
because it is a single file which renders for every row. so making it stateful wont help much.
or else if we can make a new stateful component consisting only of join project/decline button?
 
Hope so my solution wont be rejected only for loader missing on grid view.

src/projects/list/components/Projects/ProjectsGridView.jsx
```diff
@@ -186,19 +186,33 @@ const ProjectsGridView = props => {
...
         return (
-          <Link to={url} className="spacing">
-            <div className="join-btn" style={{margin: '5px'}}>
+          <div>
+            <button 
+              onClick={(e) => {
+                e.stopPropagation()
+                onUserInviteAction(true, item.id)
+              }}
+              className="join-btn tc-btn tc-btn-primary tc-btn-md" style={{margin: '5px'}}
```
maxceem: Avoid adding inline styles `style={{margin: '5px'}}` and use SCSS instead. I know it's been done such way before, but for the new code, we should avoid it.

src/projects/list/components/Projects/ProjectsGridView.jsx
```diff
@@ -186,19 +186,33 @@ const ProjectsGridView = props => {
...
               Join project
-            </div>
-          </Link>
+            </button>
+            <button 
+              onClick={(e) => {
+                e.stopPropagation()
+                onUserInviteAction(false, item.id)
+              }}
+              className="join-btn" style={{margin: '5px'}}
```
maxceem: Avoid adding inline styles `style={{margin: '5px'}}` and use SCSS instead. I know it's been done such way before, but for the new code, we should avoid it.

src/projects/reducers/projectSearch.js
```diff
@@ -94,6 +97,29 @@ export default function(state = initialState, action) {
...
+      const projects = state.projects.filter(project => project.id !== action.payload.projectId)
+      return Object.assign({}, state, {
+        projects
+      })
+    } else {
+      // take out the project, change the status and merge it back along with remaining
+      const project = state.projects.filter(project => project.id === action.payload.projectId)[0]
+      project.invites.forEach(element => {
+        if (element.userId === action.payload.userId) {
+          element.status = PROJECT_MEMBER_INVITE_STATUS_ACCEPTED
```
maxceem: In this line, you are mutating the `project` object as you are updating `element` inside `project.invites` array. This may lead to issues which hard to debug, see [Immutable Update Patterns](https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns).

rashmi73:
@maxceem 

**If I invite a manger user pshah_manger using customer user pshah_customer and manger user clicks Decline the project would disappear from the list. After I reload the list the project is shown again. As was clarified in this comment #3310 (comment) for the manager or admin roles after decline invitation we just remove the buttons but we keep the project in the list because such users can see the projects they are not invited to**

**I click Join project on the project which is not first on the list, after accepting the invitation the project becomes first. But I reload the page, it again on the position it was before. So when we click Join poroject button the project should stay on its place, we only have to hide buttons**

Can I just reload the whole page again on click of Join/decline button? so that I dont have to manage anything in reducer. will it be acceptable?

rashmi73:
@maxceem 
**The best way to achieve it would be storing the project id for which we process the invitation in the processingProjectMemberInvite but rename it to something more obvious like processingInviteForProjectId**

Ok doing the changes

maxceem:
> Can I just reload the whole page again on click of Join/decline button? so that I dont have to manage anything in reducer. will it be acceptable?

Reloading the whole page will take a long time. As we have all the information, we should just update the project we are declining/accepting the invitation.

rashmi73:
@maxceem ok understood that I need to use update utitlity from 'react-addons-update' in reducer for immutability.
I will complete it before 18 hrs

rashmi73:
@maxceem working

maxceem:
@rashmi73 As there is still no solution I have to un-assign you from this issue. The desired time for solving issues in our Bug Bash is 12 hours, but this issue took too long, and unfortunately, we cannot wait any longer as this issue is already delayed for release.

PR #3357, opened 2019-10-09
Cf19

vikasrohit:
@maxceem I create this PR just for double check as I saw it showing one pending change to be merged into dev from cf19. As far as I remember, the change shown in this PR is the hotfix we made for handling `@` when inviting users to the project, right? It should be ideally in dev already. but this PR is still showing it as pending change. Do you have idea about it?


maxceem:
@vikasrohit thanks for raising it up. We are safe here though. 

I've merged `cf19` into `hotfix/post-release-2.4.15.2` instead of `dev` as we want to release it as a hotfix, so all the latests changes from `cf19` are in the hotfix https://monosnap.com/file/tPHqdJSx72kgkxzkmiKgPTuSmSgtAU

Now I see it was better to merge the issue solution directly to hotfix branch to avoid false alarm.

vikasrohit:
> Now I see it was better to merge the issue solution directly to hotfix branch to avoid false alarm.

Yep.


@maxceem the PR has conflict.

vikasrohit:
And do we still have the option of merging the feature branches directly to the hotfix now?

vikasrohit:
@maxceem closing the PR.

PR #3358, opened 2019-10-10
Hotfix/post release 2.4.15.2 (small fix)

maxceem:
@vikasrohit this is a small fix I've done to the patch after the testing. The loading indicator wasn't shown after pressing Join/Decline buttons in the Card view.

### DIFF
--- a/src/projects/reducers/projectSearch.js
+++ b/src/projects/reducers/projectSearch.js
@@ -104,6 +104,7 @@ export default function(state = initialState, action) {
         ])
         member.role = action.payload.role
         member.projectId = action.meta.projectId
+        member.deletedAt = null // explicitly set `null` as non-deleted members have this value
 
         // add new member to member list
         let updatedProject = update(projects[projectIndex], {


src/projects/reducers/projectSearch.js
```diff
@@ -104,6 +104,7 @@ export default function(state = initialState, action) {
         ])
         member.role = action.payload.role
         member.projectId = action.meta.projectId
+        member.deletedAt = null // explicitly set `null` as non-deleted members have this value
```
vikasrohit: @maxceem why non-deleted members have `deletedAt` set?

maxceem: @vikasrohit this value comes from the server:

image

So when we add a new member "manually", we have to reconstruct the member object as close as possible to the object which comes from the server. In particular `delatedAt` is required because in another place we check if a member is non-deleted by `deletedAt  === null`, see https://github.com/appirio-tech/connect-app/blob/dev/src/projects/components/projectsCard/ProjectCard.jsx#L17

PR #3360, opened 2019-10-11
upgrade project api to v5

suppermancool:
upgrade project api to v5

### DIFF
--- a/src/api/projects.js
+++ b/src/api/projects.js
@@ -155,7 +155,7 @@ export function createProject(projectProps) {
   // have the discussions tab enabled
   projectProps.details.hideDiscussions = true
 
-  return axios.post(`${PROJECTS_API_URL}/v5/projects/`, { ...projectProps })
+  return axios.post(`${PROJECTS_API_URL}/v5/projects/`, projectProps)
     .then( resp => resp.data)
 }
 
@@ -182,7 +182,7 @@ export function createProjectPhase(projectId, phase) {
  * @return {Promise} resolves to new product
  */
 export function createPhaseProduct(projectId, phaseId, product) {
-  return axios.post(`${PROJECTS_API_URL}/v5/projects/${projectId}/phases/${phaseId}/products`, { ...product })
+  return axios.post(`${PROJECTS_API_URL}/v5/projects/${projectId}/phases/${phaseId}/products`, product)
     .then( resp => resp.data)
 }
 
@@ -192,7 +192,7 @@ export function createProjectWithStatus(projectProps, status) {
   // have the discussions tab enabled
   projectProps.details.hideDiscussions = true
 
-  return axios.post(`${PROJECTS_API_URL}/v5/projects/`, { ...projectProps })
+  return axios.post(`${PROJECTS_API_URL}/v5/projects/`, projectProps)
     .then( resp => resp.data)
     .then(project => {
       const updatedProps = { status }
--- a/src/api/projects.js
+++ b/src/api/projects.js
@@ -18,8 +18,6 @@ export function getProjects(criteria, pageNum) {
     // support for multiple comma separated types
     if (filter.type && filter.type.indexOf(',') > -1) {
       params.type = `in(${filter.type})`
-    } else if (filter.type) {
-      params.type = filter.type
     }
   }
 
--- a/src/api/projects.js
+++ b/src/api/projects.js
@@ -50,9 +50,9 @@ export function getProjectSuggestions() {
  */
 export function getProjectById(projectId) {
   projectId = parseInt(projectId)
-  return axios.get(`${PROJECTS_API_URL}/v4/projects/${projectId}/`)
+  return axios.get(`${PROJECTS_API_URL}/v5/projects/${projectId}/`)
     .then(resp => {
-      const res = _.get(resp.data, 'result.content', {})
+      const res = resp.data
       _.forEach(res.attachments, a => {
         a.downloadUrl = `/projects/${projectId}/attachments/${a.id}`
       })
@@ -72,8 +72,8 @@ export function getProjectById(projectId) {
 export function getProjectPhases(projectId, query = {}) {
   const params = _.mapValues(query, (param) => encodeURIComponent(param))
 
-  return axios.get(`${PROJECTS_API_URL}/v4/projects/${projectId}/phases`, { params })
-    .then(resp => _.get(resp.data, 'result.content', {}))
+  return axios.get(`${PROJECTS_API_URL}/v5/projects/${projectId}/phases`, { ...params })
+    .then(resp => resp.data)
 }
 
 /**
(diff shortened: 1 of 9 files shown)


suppermancool:
@maxceem 
i upgraded all project api except:
1. https://api.topcoder-dev.com/v5/projects/metadata/orgConfig?filter=orgId=in(20000010,20000013,20000015): 
the error: validation error: “orgId” is require
i’m not sure how to send array of orgId (edited) 

2. https://api.topcoder-dev.com/v5/projects/:
 it use metadata in response, i try to get x-total from header but the request doesn’t have Access-Control-Expose-Headers in response so axios can’t get it (more detail https://stackoverflow.com/questions/37897523/axios-get-access-to-response-header-fields)

I also upgraded all project attachments api but it throw some random error

maxceem:
Thanks @suppermancool

1. Checking the issues with attachments now...

2. The headers are fixed now. So we can access pagination metadata in the headers:

     image

suppermancool:
@maxceem we don't have `segment` field in https://api.topcoder-dev.com/v5/projects/, But the code send it :

https://gyazo.com/1c9fe967ebb817ee3dec354377972f9c

suppermancool:
@maxceem also don't have some other fields like: 'members', 'actualPrice', 'estimatedPrice', 'createdAt', 'updatedAt', 'createdBy', 'updatedBy', 'details', 'lastActivityAt', 'lastActivityUserId', 'version', 'templateId', 'description'

Screen Shot 2019-10-14 at 12 08 49

maxceem:
@suppermancool I'm deploying the fix for your two comments above. I will update shortly (10 - 20 mins).

suppermancool:
@maxceem ok done in the latest commit, i pass all fields to request

 the `keyword` filter doesn't work, i try to search project by name with this field but it doesn't work:
https://gyazo.com/599f3636e9623f34f08028c7a46aaa01

src/api/projects.js
```diff
@@ -4,36 +4,31 @@ import { TC_API_URL, PROJECTS_API_URL, PROJECTS_LIST_PER_PAGE } from '../config/
...
 export function getProjects(criteria, pageNum) {
   // add default params
-  const includeFields = ['id', 'name', 'description', 'members', 'status', 'type', 'actualPrice', 'estimatedPrice', 'createdAt', 'updatedAt', 'createdBy', 'updatedBy', 'details', 'lastActivityAt', 'lastActivityUserId', 'version', 'templateId']
   const params = {
-    limit: PROJECTS_LIST_PER_PAGE,
-    offset: (pageNum - 1) * PROJECTS_LIST_PER_PAGE,
-    fields: includeFields.join(',')
+    perPage: PROJECTS_LIST_PER_PAGE,
+    page: (pageNum - 1) * PROJECTS_LIST_PER_PAGE,
+    ..._.pick(criteria, ['id', 'status', 'type', 'memberOnly', 'keyword', 'name', 'code', 'customer', 'manager', 'members', 'actualPrice', 'estimatedPrice', 'createdAt', 'updatedAt', 'createdBy', 'updatedBy', 'details', 'lastActivityAt', 'lastActivityUserId', 'version', 'templateId', 'description', 'sort' ])
```
maxceem: Before in this place, we requested ALL the fields from `includeFields`. In the new logic you are requesting only the fields which are set in `criteria`. What is the reason for this change?

src/api/projects.js
```diff
@@ -4,36 +4,31 @@ import { TC_API_URL, PROJECTS_API_URL, PROJECTS_LIST_PER_PAGE } from '../config/
...
     if (filter.segment && filter.segment.indexOf(',') > -1) {
       filter.segment = `in(${filter.segment})`
     }
-    // convert filter object to string
-    const filterStr = _.map(filter, (v, k) => `${k}=${encodeURIComponent(v)}`)
-    params.filter = filterStr.join('&')
   }
-  // sort fields
-  const sort = _.get(criteria, 'sort', null)
-  if (sort) params.sort = sort
```
maxceem: Why the `sort` is removed?

src/api/projects.js
```diff
@@ -4,36 +4,31 @@ import { TC_API_URL, PROJECTS_API_URL, PROJECTS_LIST_PER_PAGE } from '../config/
...
 
-  return axios.get(`${PROJECTS_API_URL}/v4/projects/`, { params })
+  return axios.get(`${PROJECTS_API_URL}/v5/projects/`, { ...params })
     .then( resp => {
       return {
-        totalCount: _.get(resp.data, 'result.metadata.totalCount', 0),
-        projects: _.get(resp.data, 'result.content', [])
+        totalCount: _.get(resp.header, 'x-total', 0),
```
maxceem: I think it should be `resp.headerS` as per [axios docs](https://github.com/axios/axios#response-schema). Also, the value in headers would be a string, so we have to convert it to a string.

maxceem:
@suppermancool when I try to filter projects using:
- search filed
- project status
- column filter
- `My project`

Not filters are added to the URL request

image

Also, note, that we have some default query params which are added to this request even without any filters:
```
https://api.topcoder-dev.com/v4/projects/?limit=20&offset=0&fields=id,name,description,members,status,type,actualPrice,estimatedPrice,createdAt,updatedAt,createdBy,updatedBy,details,lastActivityAt,lastActivityUserId,version,templateId&sort=updatedAt+desc
```

src/api/projects.js
```diff
@@ -4,36 +4,31 @@ import { TC_API_URL, PROJECTS_API_URL, PROJECTS_LIST_PER_PAGE } from '../config/
...
-    // convert filter object to string
-    const filterStr = _.map(filter, (v, k) => `${k}=${encodeURIComponent(v)}`)
-    params.filter = filterStr.join('&')
   }
-  // sort fields
-  const sort = _.get(criteria, 'sort', null)
-  if (sort) params.sort = sort
 
-  return axios.get(`${PROJECTS_API_URL}/v4/projects/`, { params })
+  return axios.get(`${PROJECTS_API_URL}/v5/projects/`, { ...params })
```
maxceem: This issue https://github.com/appirio-tech/connect-app/pull/3360#issuecomment-541510314 is most likely caused by this change. Here we have a `get` request. So `params` here is a part of `axios` lib and represent `query` params, it's not included in the `body`. So we have to keep it, otherwise, query params are not added.

src/api/projects.js
suppermancool: @maxceem i already added sort in line `..._.pick(criteria,`

suppermancool:
@maxceem for this https://github.com/appirio-tech/connect-app/pull/3360#issuecomment-541510314
i think limit, offset are changed to perPage, page.

All done in the latest commit

maxceem:
FYI @vikasrohit we have now draft v5 version of Connect App, which points to the currently deployed version of https://circleci.com/gh/topcoder-platform/projects-api.

Not everything yet works there due to an outdated version of Project Service or some API inconsistencies. But already can be checked out.

So once you confirm a separate DEV env for this, we may start deploying it.

Issue #3359, opened 2019-10-11
Local Timezone


ajefts:
**Objective:** During registration and on the profile page of Connect capture user's local timezone, which will aid in level-setting availability and responsiveness for all users.

<p></p>

**Requirements:**

- For Connect registration, include a required question and answer for user to indicate their local timezone. Below describes the question and answer format:

  - Question: Local Timezone (*marked as required on the form)

  - Answer: User can type an entry, which should filter options for selection from a dropdown list of world timezones.

- For the Connect 'My Profile' page, display 'Local Timezone' as an editable field. Users should be able to update their local timezone by saving a new timezone from the dropdown list and saving changes. Further, users should be able to update and save their local timezone in their profile with out having to update any other fields.

<p></p>

Note: Inside a Connect's project space, this field of data will be included as an enhancement to the 'team' view.

ajefts:
Link to feature: https://tcprod.productboard.com/feature-board/planning/features/2794134

RishiRajSahu:
@vigneshTheDev would you like to pick this up ?

RishiRajSahu:
@lakshmiathreya could you please verify this in dev.

lakshmiathreya:
Verified in Dev. TZ field updates seem to happen successfully. However, possibly due to member service issue, the Profile fields are not visible - hence not able to completely verify this feature. -- --- Other fields including TZ appear blank post updates in User profile.
-- TZ not visible in Team view.
-- TZ not visible in Member profile view.

<img width="1436" alt="Screenshot 2019-10-31 at 2 09 58 PM" src="https://user-images.githubusercontent.com/51733042/67931579-33b51280-fbe8-11e9-93af-6e6cc943a11a.png">

<img width="1417" alt="Screenshot 2019-10-31 at 2 14 43 PM" src="https://user-images.githubusercontent.com/51733042/67931899-d79ebe00-fbe8-11e9-9c47-fc94fd35c6f7.png">
<img width="1480" alt="Screenshot 2019-10-31 at 2 13 57 PM" src="https://user-images.githubusercontent.com/51733042/67931907-df5e6280-fbe8-11e9-99bd-10cf26616d88.png">

RishiRajSahu:
@lakshmiathreya okay will check for old users, please verify it by creating new user. Also community profile page is out of scope for this ticket.

lakshmiathreya:
-- TZ capture and display in Profile is now working for Old and New users. 
[...]

Issue #3277, opened 2019-08-22
Registration & Profile Page Updates - Part II


RishiRajSahu:
1. Registration Page: **Update Subtitle on Screen 1**
- [x] Update the subtitle on Screen 1 of the registration workflow to read "We're excited to meet you!" instead of "First we need to know you a bit better"

2. Registration Page: **Update Text on Screen 2 (PIN Entry)**
- [x] Update message displayed on this page to read, "We've sent a 1-time PIN to your email. Check your inbox and enter the PIN provided. If its been a few minutes and you have not received your PIN, check your spam folder first and then request a new PIN be sent if not found."
- [x] Add sub-text under PIN Entry field to indicate that entry is case-sensitive: "PIN is case-sensitive"

3. Registration Page: **Update Language & Workflow on Screen 3 (Registration complete) of registration workflow.**
- [x] Update title subtext to read: "The World's Largest Crowdsourcing Platform" instead of "The World's Biggest Crowdsourcing Platform"
- [x] Update page content to read: "Hello [username], Your Topcoder account was successfully created." (consolidate the messaging here)
- [x] Add a button underneath page content, reading "Review Catalog" which should bring the user into the Connect Catalog
- [x] Update the existing button to read, "Go to Dashboard" which should bring the user into the 'All Projects' dashboard of Connect 

4. **Merge the "Notifications Settings" and "Account & Security" tabs** in the left sidebar into the "My Profile" Tab, as both are related to managing your personal profile and preferences.
- [x] The layout of information within the "My Profile" tab should be - 
Profile Information
Notifications Settings
Account & Security

5. Profile Page: **Lift "Required" Requirement for Saving Changes**
- [ ] On the "My Profile" page, a user should be able to edit any of the following fields, ~without having to update/complete any of the other incomplete fields:~
1. First Name
2. Last Name
3. Business phone
4. Business email
5. Company~
Add edit privileges as below per roles as mentioned below - 

**Role -** Customers
First Name
Last Name
Business phone
Business email

**Role -** Admin&Managers
First Name
Last Name
Business phone
Business email
~Client company list~

**Role -** Copilots 
First Name
Last Name
Email

lakshmiathreya:
@RishiRajSahu 'go to Dashboard' doesnt have link to All Projects Connect Dashboard.

vigneshTheDev:
[...]

PR #3365, opened 2019-10-23
add timezone input in profile settings page

vigneshTheDev:
Fix issue: https://github.com/appirio-tech/connect-app/issues/3359

### DIFF
--- a/src/routes/settings/routes/profile/components/ProfileSettingsForm.jsx
+++ b/src/routes/settings/routes/profile/components/ProfileSettingsForm.jsx
@@ -210,7 +210,7 @@ class ProfileSettingsForm extends Component {
             }
           </div>
         </div>
-        {this.getField('Company name', 'companyName', false, (isCustomer || isCopilot) && !isManager)}
+        {this.getField('Company name', 'companyName', true, (isCustomer || isCopilot) && !isManager)}
         <div className="field">
           <div className="label">
             <span styleName="fieldLabelText">Country</span>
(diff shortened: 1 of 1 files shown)


maxceem:
@RishiRajSahu I've noticed that in this PR apart from adding the timezone field some other change has been done. I'm not sure if they were required by some task or no:

- **Business phone** field is not required anymore, though it was required before.

RishiRajSahu:
@vigneshTheDev could you address @maxceem comments here ?

vigneshTheDev:
Hi @maxceem, sorry, it looks like a mistake. 
We had a scenario where some users 
* don't have values for required fields like company name and business phone. 
* and, the field is disabled due to the role based restrictions (https://github.com/appirio-tech/connect-app/issues/3277). 

In such case, the user will not be able to submit the profile settings form (because, the field is mandatory but it's empty and disabled. Related: https://github.com/appirio-tech/connect-app/issues/3277#issuecomment-534878815)

To handle this, I removed mandatory requirement from those fields. But later, we decided to make them mandatory again as the above mentioned issue can be addressed through customer support request. So, I reverted the company name field in this commit (https://github.com/appirio-tech/connect-app/pull/3365/commits/56b29cb4158d2c6fdb11c344d688379ee00f4536 - which is part of this pull request). I missed to revert mandatory requirement for business phone. Shall I fix it and raise a PR now?

maxceem:
Thanks for the details @vigneshTheDev.

Yes, please, revert the unnecessary changes.

maxceem:
Hey @vigneshTheDev, please, let me know when there is PR with reverted changes that are not necessary.
Thanks.

vigneshTheDev:
Hi @maxceem, I've raised the PR here: https://github.com/appirio-tech/connect-app/pull/3416

Issue #3456, opened 2019-12-20
Markdown block is being removed on Edit


vikasrohit:
### Expected behavior
Should not remove the user content.

### Actual behavior
It is removing the user content written in the markdown block in a bulleted list.

### Steps to reproduce the problem
- Go to https://connect.topcoder-dev.com/projects/8353/messages/34540
-  Try to edit the post
- Observe that the content `Some text which goes away mysteriously` is gone in the edit window (Please don't hit save, otherwise you will loose the content)

### Screenshot/screencast
<img width="803" alt="Screen Shot 2019-12-20 at 3 59 58 PM" src="https://user-images.githubusercontent.com/2417632/71248813-d2850200-2341-11ea-92b4-0a7a6a01f2c1.png">
<img width="802" alt="Screen Shot 2019-12-20 at 4 00 05 PM" src="https://user-images.githubusercontent.com/2417632/71248825-da44a680-2341-11ea-8750-9968dcc56631.png">

--

#### Environment
- OS: Any
- Browser (w/version): Any
- User role (client, copilot or manager): Any
- Account used: N/A

vikasrohit:
fyi @lakshmiathreya

maxceem:
@vikasrohit I've found one more case in which the first line disappears instead of the second line. And also how to easily reproduce this issue by creating new posts, see demo video https://monosnap.com/file/6OIfYorGVB1Ofm7tzmRWnH7cwTTB6X

maxceem:
Looks like our Markdown renderer works slightly different from GitHub. On GitHub, to make the second line to be shown as a `code` we have to add an extra line after the first line, see examples bellow:

Test 1:

- First line starts with dash and after press Enter
       Second line starts with 6 spaces

Test 2:
- First line starts with dash and after press Enter
  Second line starts with 2 spaces

Test 3:
- First line starts with dash and after press Enter **TWICE**

      Second line starts with 6 spaces

maxceem:
Another way to reproduce the issues is to create a `blockquote` and add a `code` block inside:
```
>     This text starts from 5 spaces
```

It would be "viewed" as:

![image](https://user-images.githubusercontent.com/146016/71432343-f1a1dc00-2713-11ea-8ecd-df9a4a0b122a.png)

but during editing the inner "code" block would disappear:

![image](https://user-images.githubusercontent.com/146016/71432357-054d4280-2714-11ea-97ae-4e2719336fac.png)

## General issue

[...]

PR #3480, opened 2020-01-05
Fix: Always treat nested code block as inline with code block prefix …

sanju-singh:
…spaces


sanju-singh:
Thanks @maxceem for mentioning that case, earlier I looked only for code.
Now, I fixed it for paragraph as well.
Few observation:
1. Code block has prefix of 4 sapces
2. Newline inside list item requires 2 space prefix (i.e, 6 space for code).
3. We can have code block in first line of list item as well with 1 space for list item prefix and 4 space for code.

all these scenarios will work fine now.

maxceem:
@vikasrohit we've already pushed to PROD another solution for this issue and it passed the QA. So I think we should be fine with the current solution and close this PR with and alternative solution, see https://github.com/appirio-tech/connect-app/issues/3456#issuecomment-568944094 for details.

Issue #3494, opened 2020-01-20
[$75] [Account App] sync businessPhone validation with member api


maxceem:
The current issue is used to track a task in another repository during the Bug Bash challenge.

See task details https://github.com/appirio-tech/accounts-app/issues/333

All the comments to assign, unassign and ready for review should be done here in the current issue in Connect App.

mhikeo:
@bug-bash-helper assign

maxceem:
As no PR has been provided for 12 hours this issue has been unassigned and now is open for pickup.

yoution:
@bug-bash-helper assign

yoution:
@bug-bash-helper 
https://github.com/appirio-tech/react-components/pull/346
https://github.com/appirio-tech/connect-app/pull/3507
is ready for review

yoution:
@bug-bash-helper https://github.com/appirio-tech/react-components/pull/346  ready for review

yoution:
@maxceem   please review these two repos, https://github.com/appirio-tech/connect-app/pull/3507 
https://github.com/appirio-tech/react-components/pull/346

maxceem:
@yoution works mostly good. Though implementation doesn't reuse `formsy` features, that should make the code simpler, as `PhoneInput` component is already implemented as formsy component.
Please, have a look at the reviews in Connect App https://github.com/appirio-tech/connect-app/pull/3507#pullrequestreview-347112336 and in React Components repo https://github.com/appirio-tech/react-components/pull/346#pullrequestreview-347122456

yoution:
@maxceem   https://github.com/appirio-tech/react-components/pull/347
https://github.com/appirio-tech/react-components/pull/347

i pushed again for country code validate, but the issue in registerscreen page, the country code can not show, it maybe not my problem, the code I changed may not cause the problem, please check again

maxceem:
Thanks @yoution I'll check again tomorrow. I've already marked it as `Ready or Review`.

> but the issue in registerscreen page, the country code can not show, it maybe not my problem, the code I changed may not cause the problem, please check again

I saw on your video it works good, so it's not in issue, the reason was just because example doesn't have country list.

maxceem:
@yoution I've checked it out, please, have a look at review https://github.com/appirio-tech/connect-app/pull/3507#issuecomment-577980713.

nithyaasworld:
[...]

PR #3507, opened 2020-01-22
#3494 sync businessPhone validation with member api

yoution:
https://github.com/appirio-tech/connect-app/issues/3494

### DIFF
--- a/src/routes/settings/routes/profile/components/ProfileSettingsForm.jsx
+++ b/src/routes/settings/routes/profile/components/ProfileSettingsForm.jsx
@@ -68,7 +68,9 @@ class ProfileSettingsForm extends Component {
           businessPhoneValid: true
         })
       }
-    }else if(this.state.businessPhoneValid){
+    }
+    
+    if (this.state.businessPhoneValid && !isValid){
       this.setState({
         businessPhoneValid: false
       })
--- a/src/routes/settings/routes/profile/components/ProfileSettingsForm.jsx
+++ b/src/routes/settings/routes/profile/components/ProfileSettingsForm.jsx
@@ -63,17 +63,17 @@ class ProfileSettingsForm extends Component {
           countrySelected: country.name,
         })
       }
-    }
-
-    if (isValid) {
-      this.setState({
-        businessPhoneValid: true
-      })
-    } else {
+      if(!this.state.businessPhoneValid && isValid) {
+        this.setState({
+          businessPhoneValid: true
+        })
+      }
+    }else if(this.state.businessPhoneValid){
       this.setState({
         businessPhoneValid: false
       })
     }
+
     // external change means, the user didn't change the phone number field.
     // But it was automatically changed due to country selection change. In such case, we should show
     // the alert under country selection only.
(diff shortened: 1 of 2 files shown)


src/routes/settings/routes/profile/components/ProfileSettingsForm.jsx
```diff
@@ -63,18 +63,17 @@ class ProfileSettingsForm extends Component {
           countrySelected: country.name,
         })
       }
+    }
 
-      if (!businessPhoneValid) {
```
maxceem: The logic before makes sure, that we update `businessPhoneValid` only if the value was the opposite. This is to reduce re-renders. Now `setState` would be called every time, even if `businessPhoneValid` is not changed.

yoution: 1,I use `checkphone` in component, because this will write the regexp only one place, if use validation api, will write the regexp out side of the component, will wirte two place
2,for  `businessPhoneValid`, I don't add `oldstate == newstate`, because if the state is same, I thought it  will not render, now I find i am wrong, thanks
3. I will check it

maxceem: 1. Right, if we add `validations` property, it would be added outside the component which is not good, as we would have to add it two times. Though we also can add validation **inside** the component using a `validate` method, please, have look at this code example https://github.com/christianalfoni/formsy-react/blob/master/API.md#validate

yoution: your new suggestion may not work, it only work in `mixin`, we use `react.component` and use `hoc` to transfer the functions, so it will not access the `validate` function, so what  to do next?
image

maxceem: Ah, got it. I guess using `setValidations` we can set validation rule inside `constructor` or in `componentWillMount`  like `matchRegexp: /^(\+)?(?:[0-9] ?){6,14}[0-9]$/`. 

Do you think that would work?

yoution: I have tried , it won't work, every render, it will call the `setValidations`, if you do not set the props 'validate' from outside, so it will call ` setValidations({})`, so it will overwrite the previous value

maxceem: Hmm, strange that formsy doesn't let us define validation rules from within the component. But ok, let's keep how it's done. 

So only check why country detecting doesn't work on the registration page after the fix, the rest works good.

yoution: could you tell me how to deploy the registration page locally, it's not in `connect-app` project, 
I mock the country list, it works https://monosnap.com/file/VuxzQGjCKnqLJ5ep4KVNtVyB9qPXDb, so I can not reproduct the problem in the `WizardExamples `example

maxceem: Ah, got it. Your demo video with the mocked country is enough, don't have deploy fully.

yoution:
@maxceem  no, it not work good, I lost the country code validate in this repo, so I push again

maxceem:
Thanks, @yoution.

On registration page works perfect now: validaties both code and phone with regexp.

On the profile settings page, it does validate code, but doesn't validate phone with regexp: if enter phone with valid country code, but after I add `-` the number is still valid, though `-` is not allowed in regexp:

image

yoution:
@maxceem  have pushed, please review . https://github.com/appirio-tech/connect-app/pull/3507

Issue #3637, opened 2020-01-30
Name overlapping in Last Updated column


vikasrohit:
<img width="222" alt="Screen Shot 2020-01-30 at 11 18 59 AM" src="https://user-images.githubusercontent.com/2417632/73423541-6fc85280-4352-11ea-8dcb-240403440cea.png">

We should ellipsis the name.

maxceem:
What about the dates with year, should we ellipsis them too?

![image](https://user-images.githubusercontent.com/146016/73423758-fcc9d680-4367-11ea-8aa7-f1bdc4b02a71.png)

vikasrohit:
Nopes, we should keep the column width optimal to allow full date. I guess we can reduce some space from name column.

maxceem:
Got it. I guess we can reduce the size of `REF` column a little, and if that is not enough, then also reduce a little bit from `name`.
We should make sure that date is not wrapped to the next line and doesn't go to the next column or cut for any screen size.

yoution:
@maxceem  please review https://github.com/appirio-tech/connect-app/pull/3695

nithyaasworld:
Perfect! Verified in dev.

![image](https://user-images.githubusercontent.com/5114190/73836684-de674d80-4835-11ea-9c48-b8f2c87b1fac.png)

drasticdpk:
QA Verified in Production. 

![image](https://user-images.githubusercontent.com/25297859/75536883-1b41f100-5a3e-11ea-8f14-ef3093edbbc2.png)

PR #3695, opened 2020-02-05
#3637 Name overlapping in Last Updated column

yoution:


yoution:
@maxceem sorry, I pushed again, please review again

Issue #3513, opened 2020-01-24
P1: Update Internal Catalog Display


ajefts:
**Objective:** Improve the presentation of the internal Connect catalog for work options.

<p></p>

**Requirements:**

Designs - https://marvelapp.com/cc9e4j1/screen/62801529

- All 3 screens of the Connect catalog will require updates to UI and content, including:

  - Screen 1: Solutions vs. On Demand options

  - Screen 2: Full List of Solution Options

  - Screen 3: Full List of On Demand options (TaaS/Capacity)

- Screens 2 & 3 will include substantial UI design updates, with new content displayed for:

  - Solution Description

  - Solution Deliverables

  - A link to a case study on the website

  - A 'select' button

ajefts:
Link to feature: https://tcprod.productboard.com/feature-board/planning/features/2851875

maxceem:
Here we have to do the next final fixes:
1. Increase icons size to take the same space as on new design https://marvelapp.com/cc9e4j1/screen/65344038/handoff and https://marvelapp.com/cc9e4j1/screen/65344039. So icons should have size up to 79px x 79px. We want to achieve the same sizing of the cards as in design.
2. Add font mixins https://fonts.google.com/specimen/Roboto+Condensed to the https://github.com/appirio-tech/connect-app/blob/414027e3fcee24061943ef5e5062d03828426929/src/styles/_fonts.scss with all 6 variations (light, regular, bold + italic) and select these languages, while adding font files to "src/fonts/roboto-condensed". To create fonts in all required formats eot/svg/ttf/woff you may use service https://www.fontsquirrel.com/tools/webfont-generator.
3. After font mixings are there update styles of all the headers that use Roboto-Condensed to match the design:
    - https://monosnap.com/file/LZxAP1rEJksnFZNX9y58jSGkJtrdP2
    - https://monosnap.com/file/y0vkVnldB82LavGFhVIUf1jhTGxzO0
    - as we don't have Roboto-Condensed Medium font-weight, use Regular or Bold depend on which one looks closer to design.

maxceem:
@vikasrohit after increasing icon size to 79px as per design it looks like this:

![image](https://user-images.githubusercontent.com/146016/73323236-bb5efb80-4281-11ea-8a74-06e26d69abd2.png)

![image](https://user-images.githubusercontent.com/146016/73323248-c74abd80-4281-11ea-8143-9f15901cd1e2.png)

[...]

PR #3696, opened 2020-02-05
#3513  Update Internal Catalog Display

yoution:


yoution:
@maxceem  I try another method, I add `transform: translateZ(0);`. if I remove `scale`, the issue still exist. please review

maxceem:
@yoution I can still reproduce the issue even with `transform: translateZ(0)`:

image

yoution:
so just remove `scale` ?

yoution:
@maxceem  please review again, remove `scale` and add 'float:right' may perfect

maxceem:
> so just remove scale ?

Looks like removing scale doesn't help.

> @maxceem please review again, remove scale and add 'float:right' may perfect

thanks, I will check it out tomorrow

PR #3712, opened 2020-02-20
Reporting initial work

vikasrohit:
fyi @maxceem


src/projects/detail/containers/ProjectSummaryReportContainer.jsx
```diff
@@ -9,34 +9,66 @@ import {
...
+    }
+    const thisRef = this
+    let timeoutDuration = 60*1000
+    if (REPORT_SESSION_LENGTH > 2*60) {
+      timeoutDuration = REPORT_SESSION_LENGTH*1000 - 2*60*1000
+    }
+    // set timeout for raising alert to refresh the token 2 minutes before the session expire
+    timer = setTimeout(() => {
+      console.log('Calling refresh looker session action')
+      thisRef.props.refreshLookerSession()
```
maxceem: hint: `this` inside arrow functions `() => {}` always point to the outer `this`, so we don't have to use `thisRef` and can use `this` instead, it would work.

@vikasrohit fyi

src/projects/detail/containers/ProjectSummaryReportContainer.jsx
```diff
@@ -9,34 +9,66 @@ import {
...
+
+const LookerEmbedReport = (props) => {
+  return (<iframe width="100%" src={props.projectSummaryEmbedUrl} onLoad={props.onLoad} />)
+}
+
+const EnhancedLookerEmbedReport = spinnerWhileLoading(props => {
+  return !props.isLoading
+})(LookerEmbedReport)
+
+let timer
```
maxceem: Generally, it's safer to keep `timer` inside class like `this.timer`. Otherwise, if we have 2 components on one page, they would use the same `timer` and break each other.

fyi @vikasrohit

vikasrohit: Agreed. I placed it intentionally outside because I do want to use the same time for all reporting components on the same page.
However, on second thought, I think we can move it to inside the class for better reading because I don't expect multiple copies of this component on the same page.

src/projects/detail/containers/ProjectSummaryReportContainer.jsx
vikasrohit: I always found the `this` in side the native javascript methods (like setTimeout), ambiguous. But as it is working for this case as you suggested and I double checked, I am removing the need of this Ref.

vikasrohit:
Implemented both suggestions. Thanks @maxceem

Issue #3630, opened 2020-01-29
[$30] Welcome Back! screen close button UI issue


codejamtc:
![1](https://user-images.githubusercontent.com/42398485/73381953-447b3000-428c-11ea-9700-5f187c61b2b3.png)

1. Open the Application and Login as valid user (Manager)
2. Click + New Project > On Demand > Buy Capacity > Select
3. Fill all the fields > Next 
4. Click ECS
5. Click + New Project

**Actual:** Welcome Back! screen close button UI issue

**Expected:** Should display the top right corner above the Welcome back! title

**Environment:** iPhone 5s | iOS 12.2

**Browser:** Safari 12

codejamtc:
@bug-hunt-helper add label: BugHunt_JanRelease, UI, P3

SathyaJayabal:
@codejamtc , accepted.

lakshmiathreya:
Issue is observed currently ...

vikasrohit:
@maxceem CF candidate

maxceem:
Sum up:
- While fixing this issue on small screens, the fix shouldn't break it on the big screens.

tcchhabra:
@bug-bash-helper assign

tcchhabra:
@bug-bash-helper https://github.com/appirio-tech/connect-app/pull/3756 is ready for review

maxceem:
@tcchhabra there is some feedback for the fix, please have a look https://github.com/appirio-tech/connect-app/pull/3756#pullrequestreview-372516548

tcchhabra:
@maxceem the requirement clearly said `Environment: iPhone 5s | iOS 12.2` so I didn't test other resolutions. Just made sure it is not broken for browser, I guess the video shared is broken before my fix too.

maxceem:
@tcchhabra Ah, sorry for the confusion. The `Environment: iPhone 5s | iOS 12.2` is where the issue has been found to make sure we can reproduce it.

Though we have to fix the issue in general. It can be reproduced on various mobile resolutions, not only on iPhone 5s.

tcchhabra:
@bug-bash-helper #3756 is ready for review

maxceem:
@tcchhabra fix works good now for this page, but it breaks close button on other pages, see https://github.com/appirio-tech/connect-app/pull/3756#pullrequestreview-373259998

tcchhabra:
@bug-bash-helper #3756 is ready for review

[...]

Issue #3673, opened 2020-01-30
[$10] [Common] Section title is not visible in mobile


codejamtc:
![6](https://user-images.githubusercontent.com/42398485/73473056-b6697d00-4351-11ea-8bb1-0b0ce10b70ea.png)

1. Open the Application and Login as valid user (Manager)
2. Click + New Project > Solutions > Data Visualization

**Actual:** Section title is not visible in mobile

**Expected:** Should display the section headings. In this case "Data Visualization"

**Environment:** iPhone 5S | iOS 12.2
**Browser:** Safari 12

codejamtc:
@bug-hunt-helper add label: BugHunt_JanRelease, UI, P4

SathyaJayabal:
@codejamtc , accepted

tcchhabra:
@bug-bash-helper assign

tcchhabra:
@bug-bash-helper https://github.com/appirio-tech/connect-app/pull/3758 is ready for review

drasticdpk:
Verified on Dev Env
![image](https://user-images.githubusercontent.com/25297859/76514968-e2177100-6480-11ea-980c-147bee8f3094.png)

satkanhub:
Working as expected. Video for reference.

[3673.zip](https://github.com/appirio-tech/connect-app/files/4326639/3673.zip)

codejamtc:
![image](https://user-images.githubusercontent.com/42398485/78236257-58ffc280-749f-11ea-9371-4f2b8e12e51d.png)

PR #3756, opened 2020-03-10
fix for issue #3630

tcchhabra:


tcchhabra:
@maxceem Welcome back alignment is not possible to keep same, because of the way the wizard is implemented. For that, we need to re-write impl class, which don't justify the prize here. We will need to bump prize to ~$50 for the same. It is not one button issue, it has whole wizard implementation issue, I guess the ticket was prize considering it is only height change for close button.

maxceem:
Ok, @tcchhabra. We can bump the prize to $50 for this.

The expected result: close button goes up on this screen 
image

But everything else stays as it is. So, please, try not to make too big update if possible.

Also, please, get the latest changes from `cf21` branch as we fixed some issue related to these pages https://github.com/appirio-tech/connect-app/issues/3673.

tcchhabra:
@maxceem taking some time but will try to wrap asap. Trying some tweaks with css and components to see if it could address all scenarios.

tcchhabra:
@maxceem agree bumping the prize to $50 didn't made sense here, but to research and invest time to fix with minimal changes, see if we can maintain it at $30 to compensate for time.

maxceem:
Ok, makes sense, will update to $30. Thanks.

PR #3760, opened 2020-03-10
Fix: remove character styling inside code block on save

sanju-singh:


sanju-singh:
There is a method "encodeContent" inside stateToMarkdown.js, I am not able to understand its purpose, could you help me with that?

maxceem:
Honestly, I don't know the reason why it was added. Do you think it would be safe if we stop calling this method, but only for `code` block type? So we keep for other block types as we don't have any issues with them.

sanju-singh:
@maxceem , I have fixed the issue occurring when writing unstyled code block.

PR #3762, opened 2020-03-10
issue 3634 fix

rashmi73:


maxceem:
Reverted, because it was merged to the wrong branch.

PR #3771, opened 2020-03-11
Fix: grid filter input not auto focused

sanju-singh:


sanju-singh:
@maxceem , I have fixed the highlighted points, please check.
Thanks

PR #3783, opened 2020-03-12
* re-fixed #3750: User should not be able to apply/changes filter in …

### LINKED ISSUE #3750: [$30] User should not be able to apply/changes filter in edit mode
Follow-up from https://github.com/appirio-tech/connect-app/issues/3730

On the Assets Library page we can apply filters to the files and links, see page https://connect.topcoder-dev.com/projects/8600/assets

- When we don't have any filter and start editing, we should not be able to apply filter.
- When we apply filter and start editing, we should not be able to change the filter.

image

- We should not be able to click column filters when are editing a file or link.

image

nqviet:
fixed the overlay z-index issue

### DIFF
--- a/src/components/AssetsLibrary/FilesGridView.jsx
+++ b/src/components/AssetsLibrary/FilesGridView.jsx
@@ -200,7 +200,7 @@ const FilesGridView = ({
               isSharingAttachment={isSharingAttachment}
             />}
           <ul>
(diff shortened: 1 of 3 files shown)


nqviet:
@maxceem It worked: z-index of the overlay have to be > z-index of `.FilterColHeader.filter-drop-down.txt-link` and z-index of `.delete-confirmation-modal` > z-index of the overlay

Issue #3651, opened 2020-01-30
[$20] Engage Talent Icons image are not properly aligned (IE11)


Krishna1507:
### Expected behavior
Engage talent whole page icon  should be aligned 

### Actual behavior
Engage talent whole page icons are not aligned 

### Steps to reproduce the problem
- Open the Application and Login as valid user (Customer).
- Click + New Project > On Demand 
-  Observe the engage talent page icons 
- Icon image overlapped over the text

### Screenshot/screencast
![image](https://user-images.githubusercontent.com/23450398/73445999-f8121c00-4381-11ea-8942-c96ab467906a.png)

#### Environment
   - OS: Windows 10 
   - Browser (w/version): IE11 
   - User role (client, copilot or manager): Customer
   - Account used: topcoderconnect+Customer@gmail.com

Krishna1507:
@bug-hunt-helper add label: BugHunt_JanRelease, UI, P2

SathyaJayabal:
@Krishna1507 , accepted

nithyaasworld:
This is still an issue but in a different way. Screenshot below:

![image](https://user-images.githubusercontent.com/5114190/75333190-b35ba100-58ab-11ea-994c-afc7742e5fc2.png)

vikasrohit:
we haven't done anything to fix the issue.

@maxceem if it is easily possible to fix it for IE11, we can take it in next CF and if it is tricky, we can skip fixing it if it is not an issue with Microsoft Edge.

nqviet:
@bug-bash-helper assign

nqviet:
@bug-bash-helper PR (https://github.com/appirio-tech/connect-app/pull/3757/commits/36db96f8d9cd24166b09714c99d93f373594ac9f) is ready for review

maxceem:
@nqviet please create a separate pull request for this issue. Don't combine fixes for several issues into one pull request. Here is feedback for this issue https://github.com/appirio-tech/connect-app/pull/3757#pullrequestreview-373243786

maxceem:
@nqviet the waves still look broken for me in IE11 https://github.com/appirio-tech/connect-app/pull/3789#pullrequestreview-374731458

maxceem:
Works good after restarting the app:

![image](https://user-images.githubusercontent.com/146016/76684568-e93cab80-6647-11ea-97e7-fcc9ea1695f6.png)

drasticdpk:
Verified on Dev Env
[...]

PR #3789, opened 2020-03-12
- re-fixed issue #3651: Engage Talent Icons image are not properly al…

nqviet:
- fixed IE11: the card icon position centered, card layout issue in mobile view


nqviet:
Can you please try to re-run `npm start`, when the svg files updated, the hot-reload seems not work properly?

PR #3857, opened 2020-04-07
Talent picker question

suppermancool:
Talent picker question


src/projects/detail/components/SpecQuestions.jsx
```diff
@@ -375,6 +376,13 @@ class SpecQuestions extends React.Component {
         onSkillsLoaded: this.handleSkillsLoaded
       })
       break
+    case 'talent-picker':
+      ChildElem = TalentPickerQuestion
+      _.assign(elemProps, {
+        options: q.options,
+        categoriesField: q.categoriesField,
```
maxceem, line 383: It looks like `categoriesField` is not in use in `talent-picker` question type.

src/projects/detail/components/TalentPickerQuestion/TalentPickerQuestion.jsx
```diff
@@ -0,0 +1,229 @@
...
+                        }}
+                        theme="default"
+                      />
+                    </td>
+                    <td styleName="skill-selection">
+                      <SkillsQuestion
+                        frequentSkills={[]}
+                        disabled={false}
+                        isFormDisabled={() => false}
+                        skillsCategories={[skillsCategory ? skillsCategory : '']}
```
maxceem, line 169: Shouldn't we pass an empty array `[]` or `null` if `skillsCategory` is not defined, instead of `['']`?

src/components/SelectDropdown/SelectDropdownBase.jsx
```diff
@@ -0,0 +1,176 @@
...
+    })
+    this.props.setValue(option.value)
+  }
+
+  render() {
+    const { options, theme, disabled } = this.props
+    const { selectedOption, confirmOption } = this.state
+    const selectedValue = selectedOption.title
+    const renderOption = (option, optIdx) => {
+      if (option.hide) {
```
maxceem, line 96: I guess this new property could be added to PropTypes.

src/projects/detail/components/TalentPickerQuestion/TalentPickerQuestion.jsx
suppermancool: The SkillsQuestion component will disable select field if the `selectedCategories` is empty, so i have to pass `skillsCategories` as empty array and use it like `selectedCategories`

https://github.com/appirio-tech/connect-app/blob/feature/taas-question/src/projects/detail/components/SkillsQuestion/SkillsQuestion.jsx#L148

src/projects/detail/components/SpecQuestions.jsx
suppermancool: @maxceem done

src/components/SelectDropdown/SelectDropdownBase.jsx
suppermancool: @maxceem done

src/projects/detail/components/SkillsQuestion/SkillsQuestionBase.jsx
```diff
@@ -0,0 +1,222 @@
...
+
+  getSelectedCategories(props) {
+    const {
+      currentProjectData,
+      categoriesField,
+      skillsCategories,
+    } = props
+    let selectedCategories = currentProjectData ? _.get(currentProjectData, categoriesField, []) : []
+    if (selectedCategories.length === 0 && skillsCategories) {
+      selectedCategories = skillsCategories
```
maxceem, line 133: This method now returns 2 different kinds of data:
- `_.get(currentProjectData, categoriesField, [])` contains possible values from another form field. This is not skills categories, but some values of the fields which has to be mapped to skills categories using `categoriesMapping`
- `skillsCategories` this holds real skills categories

So later if we use `getSelectedCategories` method, we might get confused as it returns a different kind of data.

Even now you pass `selectedCategories` obtained from this method to `getAvailableOptions`, but it accepts to get a different kind of data in `selectedCategories` array.

I think we can avoid creating this method as it's only necessary for the next line:
```js
const questionDisabled = isFormDisabled() || disabled || selectedCategories.length === 0
```

This line could be updated to:
```js
const questionDisabled = isFormDisabled() || disabled || (selectedCategories.length === 0 && !_.isUndefined(skillsCategories))
```

So when we provide any kind of value as `skillsCategories` we don't disable the Skills input.

src/projects/detail/components/SkillsQuestion/SkillsQuestionBase.jsx
```diff
@@ -0,0 +1,222 @@
...
+ * @param {Array}  options            all possible options
+ *
+ * @returns {Array} available options
+ */
+const getAvailableOptions = (categoriesMapping, selectedCategories, skillsCategories, options) => {
+  let mappedCategories
+  if (categoriesMapping) {
+    mappedCategories = _.map(selectedCategories, (category) => categoriesMapping[category] ? categoriesMapping[category].toLowerCase() : null)
+  }
+  if (skillsCategories && skillsCategories.length > 0 && !!skillsCategories[0]) {
```
maxceem, line 26: Checking `!!skillsCategories[0]` looks like quite hacky solution, I would suggest the next logic:
- `skillsCategories === null` - show all options
- is `skillsCategories` is array even empty we filter options

maxceem:
@suppermancool I've added 2 more comments. The main idea is instead of `skillsCategories = ['']` use `skillsCategories === null` to show all options. Then we don't have to introduce a new method `getSelectedCategories`.

Would this work?

src/projects/detail/components/SkillsQuestion/SkillsQuestionBase.jsx
```diff
@@ -0,0 +1,222 @@
...
+
+let cachedOptions
+
+/**
+ * If `categoriesMapping` is defined - filter options using selected categories.
+ * Otherwise returns all `options`.
+ *
+ * @param {Object} categoriesMapping  form data and API model categories mapping
+ * @param {Array}  selectedCategories selected categories in the form
+ * @param {Array}  options            all possible options
```
maxceem, line 17: Missing argument definition for `skillsCategories`.

src/projects/detail/components/SkillsQuestion/SkillsQuestionBase.jsx
suppermancool: @maxceem done , i also fixed the validation for optional talent picker field

src/projects/detail/components/SkillsQuestion/SkillsQuestionBase.jsx
suppermancool: @maxceem done , i also fixed the validation for optional talent picker field

src/projects/detail/components/SkillsQuestion/SkillsQuestionBase.jsx
suppermancool: @maxceem done

src/projects/detail/components/SkillsQuestion/SkillsQuestionBase.jsx
maxceem: Thanks for also fixing `optional` case.

src/projects/detail/components/SkillsQuestion/SkillsQuestionBase.jsx
```diff
@@ -0,0 +1,211 @@
...
+      categoriesMapping,
+      getValue,
+      frequentSkills,
+      skillsCategories,
+      currentProjectData,
+      categoriesField,
+    } = this.props
+    const { options, customOptionValue } = this.state
+
+    const selectedCategories = currentProjectData ? _.get(currentProjectData, categoriesField, []) : []
```
maxceem, line 142: I guess don't have to update this line as `_.get(currentProjectData, categoriesField, [])` already would return the empty array `[]` if `currentProjectData` is not defined. So now it does the same check twice.

suppermancool: @maxceem done

PR #3861, opened 2020-04-09
Talent Picker Question - Functionality - Part 1

maxceem:
Talent Picker Question implemented only functionality part without design as per specification http://www.topcoder.com/challenges/30121973


vikasrohit:
It seems it is a bigger refactoring. Just curious, why we need to refactor the SelectDropdown component? I guess, we have to consolidate the SelectDropdown components between react-components and connect-app anyway because there are places in connect app where we are using this SelectDropdown from the connect-app repo while we should use it from react-components repo. But that is a different talk.

maxceem:
> It seems it is a bigger refactoring. Just curious, why we need to refactor the SelectDropdown component?

There are two changes done to this components, and both of them are quite small and straightforward:
1. The components is splitter in 2 parts: Base and with Formsy HOC, so we can use it as a part of Formsy form and as a part of other components.
2. Added support to show the default option (`0` in our case) without ability to select it after choosing another option. (Actually, such functionality was implemented in `workstreams` branch as I thought it would be merged a long time ago).

> I guess, we have to consolidate the SelectDropdown components between react-components and connect-app anyway because there are places in connect app where we are using this SelectDropdown from the connect-app repo while we should use it from react-components repo. But that is a different talk.

Yes, I'm also reguralry thinking about this.

PR #3893, opened 2020-04-13
Progressive Registration

suppermancool:
Progressive Registration


suppermancool:
@maxceem done in https://github.com/appirio-tech/connect-app/pull/3893/commits/c89607d088db3882cf0a3434f8c023597178cd5f

suppermancool:
@maxceem all done in the latest commit except `7` i still can't reproduce it

Screen Shot 2020-04-17 at 13 07 08
Screen Shot 2020-04-17 at 13 07 42

suppermancool:

#### For src/routes/settings/routes/profile/components/ProfileSettingsForm.jsx:
I think we should create new child component for profile form. This component will have all the field, and it will have props to check which fields should be shown or required.
Then import this new component in both ProfileSettingsForm.jsx and `projects/create/components/UpdateUserInfo.js`

maxceem:
My code skippet which I used for testing (in case I would need to test it in the future):
```js

    .then(data => {
      console.warn({ data })

      const connectTrait = _.find(data, { traitId: 'connect_info' })
      const basicInfo = _.find(data, { traitId: 'basic_info' })

      console.warn({ connectTrait })

      if (connectTrait) {
        connectTrait.traits.data[0].workingHourStart = null
        connectTrait.traits.data[0].workingHourEnd = null
        connectTrait.traits.data[0].businessPhone = null
        connectTrait.traits.data[0].title = null
        connectTrait.traits.data[0].country = null
        connectTrait.traits.data[0].timeZone = null
      }

      if (basicInfo) {
        // basicInfo.traits.data[0].firstName = null
        // basicInfo.traits.data[0].lastName = null
      }

      return data
    })
```

PR #3894, opened 2020-04-14
Skills question performance

maxceem:
We would use this PR to trach code changes for refactoring.

Don't merge it before refactoring is done.


src/projects/detail/components/SkillsQuestion/SkillsQuestionBase.jsx
```diff
@@ -130,33 +160,29 @@ class SkillsQuestion extends React.Component {
...
+    let currentValues = value || getValue() || []
     // remove from currentValues not available options but still keep created custom options without id
     currentValues = currentValues.filter(skill => _.some(availableOptions, skill) || !skill.id)
 
     const questionDisabled = isFormDisabled() || disabled || (selectedCategories.length === 0 && _.isUndefined(skillsCategories))
     const hasError = !isPristine() && !isValid()
     const errorMessage = getErrorMessage() || validationError
 
-    const checkboxGroupOptions = availableOptions.filter(option => frequentSkills.indexOf(option.id) > -1)
+    const checkboxGroupOptions = availableOptions.filter(option => (frequentSkills || []).indexOf(option.id) > -1)
```
maxceem, line 182: Instead of adding `frequentSkills || []` we can define the default value in `SkillsQuestion.defaultProps` below.

src/projects/detail/components/SkillsQuestion/SkillsQuestionBase.jsx
```diff
@@ -130,33 +160,29 @@ class SkillsQuestion extends React.Component {
...
     const checkboxGroupValues = currentValues.filter(val => _.some(checkboxGroupOptions, option => option.id === val.id ))
 
-    const selectGroupOptions = availableOptions.filter(option => frequentSkills.indexOf(option.id) === -1)
+    const selectGroupOptions = availableOptions.filter(option => (frequentSkills || []).indexOf(option.id) === -1)
```
maxceem, line 185: Instead of adding frequentSkills || [] we can define the default value in SkillsQuestion.defaultProps below.

maxceem:
I suggest making the next improvements to this performance fix:

1. Don't use `identifier` anywhere to identify `SkillsQuestion` in an array. Instead of this create a new component to render one row of `TalenPickerQuestion`, like `TalentPickerRow`. I guess this component `TalentPickerRow` should be `pure` and  accept the `index` or `id` of the row and bind all necessary methods inside it to avoid rerenders.

2. Don't add `currentValue` and use existent `getValue` instead as they mainly doing the same thing. We prefer using `getValue` as we use our components inside Formsy, which requires components to have this method.
   I think it should be possible to avoid adding `currentValue` if we have `TalentPickerRow` for each row.

vigneshTheDev:
Hi @maxceem, I've made 3 different commits to fix the 3 issues mentioned above. Please review the PR: https://github.com/appirio-tech/connect-app/pull/3895

PR #3912, opened 2020-04-15
issue 3468

sumitdaga:


sumitdaga:
@maxceem  oh yes, sorry i missed that, pushed a fix. Let me know if its fine now ?

sumitdaga:
@maxceem sorry for the inconvenience, please check again ..it should be fine now!

src/projects/actions/projectMember.js
```diff
@@ -157,20 +158,15 @@ export function acceptOrRefuseInvite(projectId, item, currentUser) {
...
+      payload: updateProjectMemberInvite(projectId, inviteId, item.status).then(response => {
         // we have to add delay before applying the result of accepting/declining invitation
         // as it takes some time for the update to be reindexed in ES so the new state is reflected
         // everywhere
-        delay(ES_REINDEX_DELAY).then(() => response)
-      ),
-      meta: { projectId, currentUser },
+        // if request is accepted, then also refresh project members
+        const inviteAccepted = item.status === PROJECT_MEMBER_INVITE_STATUS_REQUEST_APPROVED
+        return delay(ES_REINDEX_DELAY).then(() => inviteAccepted ? dispatch(loadProjectMembers(projectId)) : response)
```
maxceem, line 167: @sumitdaga let's always return the response in the payload of `ACCEPT_OR_REFUSE_INVITE_SUCCESS` action for consistency as it was done before.

At the moment per new logic payload always would be an updated invite except if we change the status to `PROJECT_MEMBER_INVITE_STATUS_REQUEST_APPROVED`. It could lead to the issues in the future as we would expect that the updated invite is always returned.

sumitdaga: @maxceem you are absolutely right!   
i have fixed it and also used that response to remove the `inviteId` from the list instead of passing to `action.meta`

Issue #3774, opened 2020-03-11
[$10] Toggle button is not working for Copilot and customer users. 


drasticdpk:
### Expected behavior
The toggle button should function accordingly while clicking on an off.

### Actual behavior
There is no difference in the project list even after clicking the My Projects toggle button.   
Suggestion: If this doesn’t imply for Copilot/Customer or other roles then we don’t need to have this feature for them.

### Steps to reproduce the problem
1.	Login as Copilot/customer user. 
2.	Go to the dashboard.
3.	Go to Top Right header menu.
4.	Click on Toggle Button of “My Projects”

### Screenshot/screencast
Attach or link a resource.
![off](https://user-images.githubusercontent.com/25297859/76414033-a4521400-63be-11ea-9856-97b2e8abccd3.png)
![On](https://user-images.githubusercontent.com/25297859/76414035-a61bd780-63be-11ea-85e8-12b83b019a92.png)

--

#### Environment
- OS: Windows-10
- Browser (w/version): Chrome/ Version 80.0.3987.122 (Official Build) (64-bit))
- User role (client, copilot or manager): Copilot
- Account used: topcoderconnect+copilot@gmail.com

vikasrohit:
Good catch @drasticdpk . I will add this to feature planning board so that we can get product manager's feedback.

vikasrohit:
Got confirmation from product manager, we are good with removing the switch for customers and copilots for now. Not urgent, so we can keep it in release 2.9.

maxceem:
Sum up:
- add a new permission `SEE_MY_PROJECTS_FILTER` as per https://github.com/appirio-tech/connect-app/blob/dev/docs/permissions-guide/permissions-guide.md

iversonLv:
@bug-bash-helper assign

iversonLv:
@bug-bash-helper unassign

ijxsid:
@bug-bash-helper assign

rashmi73:
@bug-bash-helper assign

rashmi73:
@maxceem I would like to work on it

simranb86:
@bug-bash-helper assign

ijxsid:
@bug-bash-helper unassign

simranb86:
@bug-bash-helper assign

simranb86:
@maxceem FYI ☝️

rashmi73:
[...]

PR #3917, opened 2020-04-16
Fix for Issue #3774

simranb86:


simranb86:
@maxceem oh ok, will make the change now

PR #3919, opened 2020-04-16
issue 3841

sumitdaga:


sumitdaga:
@maxceem  while adding new feed also, links count was not showing, i have fixed that too!  

thanks

Issue #3773, opened 2020-03-11
[$35] Some of the buttons doesn’t get highlighted on mouse hovering.


drasticdpk:
### Expected behavior
The button should get highlighted when the mouse gets hover on it. .

### Actual behavior
The button is not getting focused while hovering the mouse.  

Note : Please implement this feature across all the buttons. 

### Steps to reproduce the problem
1.	Login as Copilot user. 
2.	Go to the dashboard.
3.	Go to Top Right header “+New Project” . 
4.	Hover the mouse over “+New Project” button. 

### Screenshot/screencast
Attach or link a resource.

--
![mousehover](https://user-images.githubusercontent.com/25297859/76411652-56d3a800-63ba-11ea-8754-9f4b558c7f1d.png)

#### Environment
- OS: Windows-10
- Browser (w/version): Chrome/ Version 80.0.3987.122 (Official Build) (64-bit))
- User role (client, copilot or manager): Copilot
- Account used: topcoderconnect+copilot@gmail.com

maxceem:
Sum up:
- we have a default **hover** style for buttons, for example, see **Save settings** button on https://connect.topcoder-dev.com/settings/profile:

   ![hover](https://user-images.githubusercontent.com/146016/79345848-52128e80-7f3a-11ea-8e56-d58081eaa3cf.gif)

- in other places, where we show the standard button we have to **reuse** our global styles as we do for **Save settings** button. Also, we have to make sure we don't break it.
   
- if a button has some additional styles we have to keep them (but also keep standard hover effect)

- places to fix:
   - **New project** on https://connect.topcoder-dev.com/projects?sort=lastActivityAt%20desc

      ![image](https://user-images.githubusercontent.com/146016/79346386-f7c5fd80-7f3a-11ea-8ae4-8629d755d07f.png)

  - **Add new phase** on the project page https://connect.topcoder-dev.com/projects/8600

     ![image](https://user-images.githubusercontent.com/146016/79346429-0c09fa80-7f3b-11ea-8cd3-fd6b4bd62659.png)

  - **Add new..** on the Assets Library page https://connect.topcoder-dev.com/projects/8600/assets:

      ![image](https://user-images.githubusercontent.com/146016/79346791-6efb9180-7f3b-11ea-8a9e-48116d256b39.png)

      - replace it with standard blue button same like **Save settings** (use the same size class).

maxceem:
[...]

PR #3933, opened 2020-04-18
feedback fix

Vips19:


Vips19:
#3773

src/projects/detail/containers/AssetsInfoContainer.jsx
```diff
@@ -836,7 +836,7 @@ class AssetsInfoContainer extends React.Component {
             </div>
             {(showAddNewButton) && (
               <div styleName="assets-header-button">
-                <button type="button" onClick={newButtonClick} styleName="add-new-button">Add new...</button>
+                <button type="button" onClick={newButtonClick} styleName="add-new-button" className="tc-btn tc-btn-primary tc-btn-sm">Add new...</button>
```
maxceem, line 839: As now we use standard styles for buttons, can we remove `styleName="add-new-button"` and corresponding style from SCSS file, as we don't use it?

Vips19: @maxceem Yes we can remove the above class, I have also updated the same styles from _button.scss as mentioned by you. Thanks for your feedback

Vips19:
@maxceem Thanks

PR #3935, opened 2020-04-18
style talent picker component

vigneshTheDev:


vigneshTheDev:
Hi @maxceem, Issue #1 is fixed.

For #2, are you trying to copy paste non numbers?

maxceem:
@vigneshTheDev I'm trying to copy/paste numbers. 

Note, even if I try to `copy` from the field, it doesn't copy.

I think because I use key `command` and in the PR you are checking for `ctrl` (I think).

vigneshTheDev:
@maxceem, yeah you're correct. Fixed it now. Can you please verify?

maxceem:
@vigneshTheDev from the quick test it works great now. I would recheck everything the final time tomorrow morning.

PR #3940, opened 2020-04-20
Issues-3853: respect the filters when switch tab

nahidshahin:
Issues-3853: respect the filters when switch tab

### DIFF
--- a/src/projects/list/components/Projects/Projects.jsx
+++ b/src/projects/list/components/Projects/Projects.jsx
@@ -173,19 +173,20 @@ class Projects extends Component {
   }
 
   setFilter(name, filter) {
-    let { previousFilters } = this.state
-
     let criteria = _.assign({}, this.props.criteria)
     if(filter && filter !== '') {
       const temp = {}
       temp[`${name}`] = `*${filter}*`
-      previousFilters[`${name}`] = `*${filter}*`
       criteria = _.assign({}, criteria, temp)
+      this.setState((state) => {
+        return {previousFilters: _.assign({}, state.previousFilters, temp)};
+      });
     } else if(_.has(criteria, name)){
-      previousFilters = _.omit(previousFilters, name)
       criteria = _.omit(criteria, name)
+      this.setState((state) => {
+        return {previousFilters: _.omit(state.previousFilters, name)};
+      });
     }
-    this.setState({previousFilters})
 
     this.props.loadProjects(criteria)
   }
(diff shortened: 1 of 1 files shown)


src/projects/list/components/Projects/Projects.jsx
```diff
@@ -155,7 +158,8 @@ class Projects extends Component {
   }
 
   applyFilters(filter) {
-    const criteria = _.assign({}, this.props.criteria, filter)
+    const { previousFilters } = this.state
+    const criteria = _.assign(previousFilters, this.props.criteria, filter)
```
maxceem, line 162: @nahidshahin In this line, the `state` is mutated, because `previousFilters` is mutated. Please, avoid mutating `state` as it's a strong antipattern, which may lead to issues, see https://reactjs.org/docs/react-component.html#state

image

src/projects/list/components/Projects/Projects.jsx
```diff
@@ -169,14 +173,19 @@ class Projects extends Component {
   }
 
   setFilter(name, filter) {
+    let { previousFilters } = this.state
+
     let criteria = _.assign({}, this.props.criteria)
     if(filter && filter !== '') {
       const temp = {}
       temp[`${name}`] = `*${filter}*`
+      previousFilters[`${name}`] = `*${filter}*`
```
maxceem, line 182: In this line, the `state` is mutated too, because `previousFilters` is mutated. Please, avoid mutating `state` as it's a strong antipattern, which may lead to issues, see https://reactjs.org/docs/react-component.html#state

image

src/projects/list/components/Projects/Projects.jsx
nahidshahin: @maxceem changed this line.

src/projects/list/components/Projects/Projects.jsx
nahidshahin: @maxceem refactored

nahidshahin:
@maxceem 

> So we have to make it consistent and apply one of two possiblee fixes:
> 
> 1. When we change tab we have to clear table fitlers, because we don't respect them
> 2. OR we don't clear filters in table by Customers but we have to respect it, and still filter projects by Customers even when we switch tab
> 

I implemented the 2nd solution here.

src/projects/list/components/Projects/Projects.jsx
maxceem: @nahidshahin I see you've updated the code, but the mutation is still there, see the screenshot:

image

PR #3952, opened 2020-04-22
filter fixed

abdullatku:


src/projects/list/components/Projects/ProjectListFilterColHeader.jsx
```diff
@@ -12,7 +12,7 @@ class ProjectListFilterColHeader extends React.Component {
     this.setFilter = this.setFilter.bind(this)
     this.onChange = this.onChange.bind(this)
     // debounced setFilter function to prevent polluting server with requests
-    this.debouncedSetFilterCheck = _.debounce(this.setFilter, 500)
+    this.debouncedSetFilterCheck = _.debounce(this.setFilter, 1500)
```
maxceem, line 15: @abdullatku please, don't increase this delay, as we want the project list to be dynamically updated when we type something into the filter.

src/projects/list/components/Projects/ProjectListFilterColHeader.jsx
```diff
@@ -30,6 +30,11 @@ class ProjectListFilterColHeader extends React.Component {
       setTimeout(() => this.textInputFilter.focus())
     }
   }
+  componentWillReceiveProps() {
+    this.setState({
+      value: this.props.value || ''
+    })
+  }
```
maxceem, line 37: @abdullatku as you've made this component controllable we don't need the internal state anymore then, and can use `this.props.value` directly, and remove `value` from state.

src/projects/list/components/Projects/Projects.jsx
```diff
@@ -155,7 +157,10 @@ class Projects extends Component {
   }
 
   applyFilters(filter) {
-    const criteria = _.assign({}, this.props.criteria, filter)
+    const sort = {"sort": this.props.criteria.sort}
+    const criteria = _.assign({},sort, filter)
+    this.setState({criteria: criteria})
```
maxceem, line 162: @abdullatku here the `criteria` is set to the state. It seems to me that we never `criteria` from the state, do we need this?

maxceem:
After discussion with Vikas we decided not to apply this solution, but to refactor the code responsible for projects searching, filtering and loading first, to make code clear and straightforward first.

PR #3954, opened 2020-04-22
fix #3951 - talent picker not updating value text in read optimized mode

### LINKED ISSUE #3951: Talent Picker Component - Issue: value is not updated in read-optimized mode
### Expected behavior
The value should be updated in read-optimized mode when we change values inside the component.

### Actual behavior
The value is not updated.

- Open this project creation form https://connect.topcoder-dev.com//new-project/talent-as-a-service-specialists
- Fill the fields of Talent Picker Component and click **Next**

   image

- After clicking Next, see how the values are displayed in "read-optimized" mode:

   image

- Now open the question again by clicking the tick on the right and change the quantity of some people: expected behavior, the quantity is also updated in read optimized mode, when we change it in the Talent Picker Component, but it's not:

   image

### Screenshot/screencast

Demo video https://monosnap.com/file/CF0pwjMYVmHv90U6jrCR3OMcQgQ7sT

vigneshTheDev:


vigneshTheDev:
@maxceem, the issue is fixed. I didn't notice the empty default props

Issue #3950, opened 2020-04-22
Talent Picker Component


maxceem:
Implement a new component for project creation form to choose talents:

![image](https://user-images.githubusercontent.com/146016/79947553-d012f000-847a-11ea-8d6e-ed9cfef963b5.png)

### Functionality

- The user needs to enter # People, Duration, and Skills for a minimum of one row for the question to be considered “answered.”
- The user does not need to complete all rows for the question to be considered “answered.”
- Clicking the + button at the end of a row should generate another row for the same role. 
Example: Clicking + at the end of the Designer row should create a second Designer row where the user can enter different numbers of people required/duration required/skills required.
- Clicking the X button at the end of a row deletes a row created by the user. They cannot delete the default rows, but they can leave these default rows as blank (as long as one row is completed, to fulfill the required answer functionality).

### Design

https://marvelapp.com/89h2ch3/screen/68335433/handoff with the following adjustments:

- don't show "months" inside, just a number. Add "months" to column header: "Duration (months)"
- If we enter too many skills, let the height of the input of the skills become multiline and grow down
- Reuse the existent style of inputs for **peoples**, **duration** and **skills**, don't implement a new design for inputs
- Don't make the component wide as on design, it should be the same width as other components in the project creation form
- As there is not enough space for "skills" component on one line, put it to the second line, and make the duration and people inputs take the whole space

### Read-optimized mode

- When we answer questions and click Next button, we render them in `read-optimized` mode inside an Accordion, see https://monosnap.com/file/5tHApYK9YEXh5IWXXXRdu2PTohEhJQ. We can open accordion and edit the values using normal UI https://monosnap.com/file/bkSCib1FWEovlfOGf60NutP7x0ZYPp.
- For this component it should display the quantity of people per role: e.g. "Designer: 9, Developer: 5".
- When we open the accordion, it should show the normal UI to edit question as we implemented above. Similar to other question types.

### QA guidelines

This new component should support some general form feature, and we should verify that they work:
- When we select some skills and click **Back** and after **Next** - the selected values should persist.
- When we select some skills and click **Next** and after **Back** - the selected values should persist.
- If we started filling the form, select some specialists, closed the browser tab and after follow project creation link again https://connect.topcoder-dev.com/new-project/talent-as-a-service-specialists we restore the form on the last place where the user left it. Selected specialists should be properly restored.
[...]

PR #3967, opened 2020-04-28
Feature/improved talent question validation

vigneshTheDev:


vigneshTheDev:
Fixes this: https://github.com/appirio-tech/connect-app/issues/3950#issuecomment-619919353

vigneshTheDev:
@maxceem, we're setting the value as a string because the number input's value is actually of type string. 

While comparing, it's possible to compare with a string or a number. When we use a number 0, JavaScript will automatically convert the other value (E.g. value.duration) into a number and compares.

String comparison will be a comparison of ASCII values. Number comparison will be the comparison of actual number values. Our intention here is to do an actual number comparison.

Also, when the conversion happens, empty string and "0" as a string both produce number 0. This is in line with our assumption as we're considering empty value in a field as 0.

With our assumption, string comparison might fail this case: `"" >= "0" // false` where number comparison passes all combination of comparisons. Though we don't have the >= scenario in our logic, to express the intention, I've used number comparison.

maxceem:
@vigneshTheDev thanks for explanation.

I would prefer to avoid implicit type conversion, as we do when we use `===` instead of `==`. Though, we can try to keep it this way, and see if it creates any drawbacks in the future or no.

PR #3981, opened 2020-05-04
enforce max values for people and duration in talent picker

vigneshTheDev:


src/components/PositiveNumberInput/PositiveNumberInput.jsx
```diff
@@ -44,8 +46,34 @@ class PositiveNumberInput extends React.PureComponent {
...
+   */
+  enforceInputBelowMax(evt) {
+    const { onChange } = this.props
+    const previousValue = evt.target.value
+    if (this.isBelowMaxLimit(previousValue)) {
+      // persists the synthetic event. So that we can use it inside setTimeout
+      evt.persist()
+
+      // setTimeout to let the input element set the actual value after the keydown/paste
+      setTimeout(() => {
```
maxceem, line 61: @vigneshTheDev To force users to input numbers above we used `evt.preventDefault()`, and in this case you use `setTimeout` with overriding event objects. I just wonder if this way is better for this case, or we cannot use `evt.preventDefault()` here?

vigneshTheDev: @maxceem, There are a couple of issues with using preventDefault in this case. 

- During the key down event, "event.target.value" holds the previous value. We will not be able to compute the resultant value of the input as it varies based on the cursor position and selection if any.

- For example, If I have 10000 in the input field, and if the cursor is at the end of the input field. I press 1 on the keyboard. The resultant value is going to be 100001 which is above the max value 10000. But if there is a selection like the below one, the resultant value is going to be 100 which is a valid value.
firefox_38cxD24ngl

- Unfortunately, number input does not allow detecting the cursor position/selection like other input fields

> According to https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement,
selectionStart and selectionEnd properties applies only to elements with type text/password/search/tel/url/week/month

We have two alternative options. Neither of them is perfect. But mentioning it here:
1. If we can use text input, we'll be able to calculate the resultant value using the cursor positions. But we lose the features provided by the number input like increase/decrease buttons and number keypad on mobile.

2. We have an option to use input type "tel" which gives a keypad with numbers only with slight modifications (more like a dial pad instead of a number pad). But we lose the increment decrement buttons on desktop.

vigneshTheDev: Here is a comparison of number key pad and dial pad
Screenshot_20200506-121401
Screenshot_20200506-121407

vigneshTheDev:
> 
> 
> @vigneshTheDev works good in most cases except if we increase number using arrows on the keyboard or in UI, see [demo video](https://monosnap.com/file/pip2IdoOjTApeOCq1W05Unvx5CFwbd). Is it possible to enforce this case too?

Got it. I'll fix this issue.

src/components/PositiveNumberInput/PositiveNumberInput.jsx
maxceem: Got it. Thanks for details and suggesting alternatives. The current approach looks the best choice for us, as it is the most simple one.

vigneshTheDev:
Hi @maxceem, fixed the issue. Please review

PR #4155, opened 2020-10-19
feat: don't reload phases after adding

maxceem:
This fix to avoid the issue with a delay of ES indexing after moving from RabbitMQ to Kafka in Project Service, see https://github.com/topcoder-platform/tc-project-service/issues/584#issuecomment-709286420.

### DIFF
--- a/src/projects/detail/containers/ProjectAddPhaseContainer.jsx
+++ b/src/projects/detail/containers/ProjectAddPhaseContainer.jsx
@@ -14,8 +14,6 @@ import { renderComponent, branch, compose, withProps } from 'recompose'
 import spinnerWhileLoading from '../../../components/LoadingSpinner'
 import { createProduct } from '../../actions/project'
 import { getProductTemplateByKey } from '../../../helpers/templates'
-import { loadProjectDashboard } from '../../../projects/actions/projectDashboard'
-
 
 import CoderBot from '../../../components/CoderBot/CoderBot'
 import Wizard from '../../../components/Wizard'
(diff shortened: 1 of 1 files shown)


vikasrohit:
@maxceem how does it fixes the delay in indexing? It is still reloading the page which might get old data. Should we put explicitly delay of 5 seconds like member acceptance?

maxceem:
@vikasrohit page should not be reloaded after this fix. Is it still reloading for you?

vikasrohit:
Nopes, I didn't try, I just tried to interpret the code.

maxceem:
I've removed the reference to the action `loadProjectDashboard` as it's not called to keep things clear.

vikasrohit:
ok, so how the problem of delay is solved with this change?

maxceem:
yes, as per my local testing.

vikasrohit:
Okay, however, my concern is that what is solution that we implemented in this PR to resolve the delay issue? I am not seeing the 5 second delay added.

maxceem:
Workflow before was:
- `POST /phases` - create a new phase on the server
- `GET /phases` - load all the phases from the server as we expected cascading updates
- redraw all the phases which we loaded from the server

Workflow now:
- `POST /phases` - create a new phase on the server
- just add a newly added phase client-side without reloaded all the phases

So we don't need a 5 seconds delay, because we don't reload the phases. And `POST /phases` endpoint returns newly created phase immediately.

vikasrohit:
Gotcha!!

Issue #4130, opened 2020-09-22
Update TaaS Intake Form: Backend Calls - Goal Due Date: 12/5


acshields:
- Required Skills and Preferred Skills reference data needs updating to use EMSI standard.
- Once TaaS intake form is submitted, the "Bookings API" (name TBD) should be called and a record for each requested role, along with the associated skills, should be created.

### Tasks here (this list is managed by @maxceem)
- [x] create jobs on TaaS project creation
- [x] update skill dropdowns to list skills from Skills API
- [x] skills dropdowns should load all the pages from Skills API (and don't' make duplicate calls for each skills dropdown instance)
- [x] show TaaS Project link after project creation

acshields:
@RishiRajSahu - This is the issue I gave you a heads up on a couple weeks ago, regarding some future updates that will be needed for the TaaS intake form. As you prepare for this feature, I would recommend chatting with Will Price and Nagendra (he will be owning TaaS App from the engineering standpoint) to learn more about the "Bookings API" they are currently planning, as well as any other questions you might have.

wdprice:
The APIs are still in development, but I can at least outline the top level items.

- Skills selection list must populate with values pulled from api.topcoder.com/v5/skills. A skill provider id will need to specified - we will provide this at the time of implementation

- Create project as you do today on v5/projects, with a Type= TaaS. (I'm sure the actual field is different, we can hammer that out.

- POST the job, job description, number of openings, list of required skills, and other relevant values to to v5/jobs ([swagger here](https://github.com/topcoder-platform/taas-apis/blob/dev/docs/swagger.yaml))

- Update the the link on the button the appears on conclusion of intake to direct the user to TaaS-App (URL structure TBD)

wdprice:
@RishiRajSahu @acshields - FYI, our TAAS APIs should be deployed to topcoder-dev tomorrow. Once deployed this ticket should be able to be worked on.

cc: @nkumar-topcoder

nkumar-topcoder:
@RishiRajSahu  @acshields  Dev deployment of **taasapp** completed. 
The Dev integration For taasapp is completed. You can verify for JOB creation. Few pointers.

1. Endpoint : http://api.topcoder-dev.com/v5/taasapi/jobs
2. Swagger & Postman collection : https://github.com/topcoder-platform/taas-apis/tree/dev/docs
[...]

PR #4211, opened 2020-12-12
load and cache all skills for skills question

vigneshTheDev:
Fix for https://github.com/appirio-tech/connect-app/issues/4130#issuecomment-740608131


maxceem (changes requested):
Thanks, @vigneshTheDev it works great.

Only in the edge case when API request fails it cases uncaught error in the code:

image

We can set empty skills list in if error happens so the code doesn't get broken.

vigneshTheDev:
Hi @maxceem, I've fixed the issue. Please review the latest

maxceem (approved):
Thanks, @vigneshTheDev works perfectly now.

Issue #4236, opened 2020-12-29
[$45] If we change status on the project listing then customer topcoder users disappear


drasticdpk:
**Reference** 
![image](https://user-images.githubusercontent.com/25297859/103284847-7ea50280-4a04-11eb-91b0-c9aef9257a49.png)

maxceem:
Contest https://www.topcoder.com/challenges/30161152 has been created for this ticket.<br/><br/>```This is an automated message for maxceem via Topcoder X```

yoution:
@maxceem please assign to me

maxceem:
Contest https://www.topcoder.com/challenges/30161152 has been updated - it has been assigned to yoution.<br/><br/>```This is an automated message for maxceem via Topcoder X```

yoution:
@maxceem please review

maxceem:
This ticket was not processed for payment. If you would like to process it for payment, please reopen it, add the ```tcx_FixAccepted``` label, and then close it again<br/><br/>```This is an automated message for maxceem via Topcoder X```

drasticdpk:
Verified on Dev.
Status from **In- review** to **Active** 
![image](https://user-images.githubusercontent.com/25297859/103734637-81ab7e80-5014-11eb-81ae-564370bfb874.png)

drasticdpk:
Verified on Prod
![image](https://user-images.githubusercontent.com/25297859/104001872-63867f80-51c8-11eb-83d5-878108a6738a.png)

maxceem:
Challenge https://www.topcoder.com/challenges/f5f3e9bd-838e-4646-9eac-cb8b09ef2f07 has been created for this ticket.<br/><br/>```This is an automated message for maxceem via Topcoder X```

maxceem:
Challenge https://www.topcoder.com/challenges/f5f3e9bd-838e-4646-9eac-cb8b09ef2f07 has been assigned to yoution.<br/><br/>```This is an automated message for maxceem via Topcoder X```

maxceem:
Payment task has been updated: https://www.topcoder.com/challenges/f5f3e9bd-838e-4646-9eac-cb8b09ef2f07
*Payments Complete*
Winner: yoution
Copilot: maxceem
Challenge `f5f3e9bd-838e-4646-9eac-cb8b09ef2f07` has been paid and closed.<br/><br/>```This is an automated message for maxceem via Topcoder X```

PR #4238, opened 2021-01-04
fix: #4236

yoution:
@maxceem please review for #4236

### DIFF
--- a/src/projects/list/components/Projects/Projects.jsx
+++ b/src/projects/list/components/Projects/Projects.jsx
@@ -331,11 +331,7 @@ class Projects extends Component {
   }
 }
 
-const mapStateToProps = ({ projectSearch, members, loadUser, projectState, templates, notifications }) => {
-  if (projectState.project && projectState.project.id && projectSearch.projects) {
-    const index = _.findIndex(projectSearch.projects, {id: projectState.project.id})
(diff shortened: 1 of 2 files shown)


maxceem (changes requested):
@yoution great fix.

There is only one thing to improve. When I update project status using admin user (maybe other too), additional data which we load for some members got lost. See demo video https://monosnap.com/file/f8Z2U9EVbZasx8oTcY5wcBuZlX5i8p.

I think it happens because after loading project list, we additionally load some user details and put it inside project data in Redux store.

So we have to 2 way so go:
- after updating project reload user data for it.
- or when we update project we can get `members` from the previous project data, and apply it to the updated project data, so we don't have to reload members.

Maybe you have another idea.

yoution:
@maxceem please review again

maxceem:
Thanks for the update, @yoution. Would it be hard to update the project object as it was done before and just keep members?

I like the previous way as it was more universal, if tomorrow we update something else from this page. Also, last updated time is updating during status update and the user who did the last update. So it would be nice if all these data got updated.

yoution:
@maxceem I know ,I will update soon

yoution:
@maxceem please review again again

maxceem (approved):
Thanks, @yoution, works perfectly.

Issue #4234, opened 2020-12-29
[$35] User can update same Project Status


drasticdpk:
User can update same Project Status ( eg Active to Active )

**Reference** 
https://monosnap.com/file/8FJS03n4OEFbVNZdl54gslmi5njyfe

 At Project listing page too.
![image](https://user-images.githubusercontent.com/25297859/103283318-5a472700-4a00-11eb-9845-987f67784014.png)

acshields:
@drasticdpk - good catch. we will get this resolved.

maxceem:
Contest https://www.topcoder.com/challenges/30161153 has been created for this ticket.<br/><br/>```This is an automated message for maxceem via Topcoder X```

yoution:
@maxceem please assign to me

yoution:
![image](https://user-images.githubusercontent.com/3863914/103551398-3fcfea80-4ee5-11eb-9b78-eb5e162507ff.png)
should also add the tooltip? what 's the  content in the tooltip?

maxceem:
Contest https://www.topcoder.com/challenges/30161153 has been updated - it has been assigned to yoution.<br/><br/>```This is an automated message for maxceem via Topcoder X```

maxceem:
> should also add the tooltip? what 's the content in the tooltip?

@yoution no, the option should look the same as now. Just make the cursor as default instead of a hand. And when we click it, nothing should happen.

yoution:
@maxceem please review

maxceem:
This ticket was not processed for payment. If you would like to process it for payment, please reopen it, add the ```tcx_FixAccepted``` label, and then close it again<br/><br/>```This is an automated message for maxceem via Topcoder X```

drasticdpk:
Verified on Dev env.
From list view >  **OK**.
![image](https://user-images.githubusercontent.com/25297859/103733682-650e4700-5012-11eb-9a31-5fdce51a6038.png)

But still user can able to update same status from project details page , please fix it.
cc @yoution @maxceem 

![image](https://user-images.githubusercontent.com/25297859/103733828-b1f21d80-5012-11eb-9b46-1cca66b787b5.png)

maxceem:
@yoution there is also another place on the challenge details page, could you please also take care of that?

yoution:
@maxceem ok

maxceem:
@drasticdpk great catch.

This is fixed and ready for QA.

drasticdpk:
[...]

PR #4242, opened 2021-01-05
fix: #4234

yoution:
@maxceem please review #4234


maxceem (approved):
@yoution thanks, works good.

Issue #4206, opened 2020-12-09
[$170] Summarized TaaS Intake Form Changes


acshields:
@RishiRajSahu - the TaaS team and will have requested these updates be made to the TaaS intake form. Please review - if you have questions, clarify with Will directly :) 

https://docs.google.com/presentation/d/1jQFX-eAMjGn5D6uCp18OKmivW9FfYXd8vjTTWOxEUg4/edit#slide=id.p

acshields:
@RishiRajSahu - do you think these changes can be accomplished before 1/15? If not, I will deprioritize from January for now (re: our new "project proposal" process that is kicking off in Prod/Dev at the start of the new year).

maxceem:
@RishiRajSahu if I understand it correctly, from the code point of view we have to create a new component for entering Jobs instead of Talents as per the description of these 2 screens and the rest can be done by editing Project Template in Metadata:

     ![image](https://user-images.githubusercontent.com/146016/103518549-a371fd80-4e7c-11eb-9962-2e9245e3f35c.png)

     ![image](https://user-images.githubusercontent.com/146016/103518568-acfb6580-4e7c-11eb-8389-4284fc2cd1a6.png)

maxceem:
@RishiRajSahu one question I have: should we remove the existing big field for **Job Description**, as we are adding another **Job Description** field on the first line?

RishiRajSahu:
@maxceem yes let's remove it and also remove optional skills field too.

@acshields we are removing the redundant job-description field from the component.

maxceem:
@RishiRajSahu do we have any ProjectTemplate which you would like to use on DEV for this?

I think we would name a new component for the jobs as `job-picker` (`JobPickerQuestion`) instead of `talent-picker`/`talent-picker-2` (`TalentPickerQuestion`/`TalentPickerQuestionV2`).

RishiRajSahu:
@maxceem feel free to duplicate the one we are using in dev for taas.

maxceem:
Sum up:
- At the moment we have a form for selecting talents https://connect.topcoder-dev.com/new-project/talent-as-a-service which looks like this when we are selecting talents:

   ![image](https://user-images.githubusercontent.com/146016/103640449-b5c16980-4f58-11eb-8b74-c39f8c5137f4.png)
   
[...]

PR #4245, opened 2021-01-06
fix: #4206

yoution:
@maxceem please reivew #4206

### DIFF
--- a/src/projects/detail/components/SpecQuestions.jsx
+++ b/src/projects/detail/components/SpecQuestions.jsx
@@ -12,6 +12,7 @@ import SpecQuestionIcons from './SpecQuestionList/SpecQuestionIcons'
 import SkillsQuestion from './SkillsQuestion/SkillsQuestion'
 import TalentPickerQuestion from './TalentPickerQuestion/TalentPickerQuestion'
 import TalentPickerQuestionV2 from './TalentPickerQuestion/TalentPickerQuestionV2'
+import JobsPickerQuestion from './JobsPickerQuestion/JobsPickerQuestion'
 import SpecFeatureQuestion from './SpecFeatureQuestion'
 import ColorSelector from './../../../components/ColorSelector/ColorSelector'
 import SelectDropdown from './../../../components/SelectDropdown/SelectDropdown'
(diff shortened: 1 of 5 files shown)


src/projects/detail/components/SpecQuestions.jsx
```diff
@@ -389,6 +390,12 @@ class SpecQuestions extends React.Component {
         options: q.options,
       })
       break
+    case 'jobs-picker':
+      ChildElem = JobsPickerQuestion 
+      _.assign(elemProps, {
+        options: q.options,
+      })
```
maxceem, line 397: @yoution we don't use options for this component, so this can be removed.

maxceem (changes requested):
@yoution thanks for fast fix. I've noticed a few issues.

1. Styles has changed for the text fields. We have to use the same styles as before:

    image
    
2. Skills list is not shown, it should be shown. Probably have to pass some other value for `skillsCategories` maybe `undefined` or `null`. Our aim is to show all skills, not only one category.

    image
    
3. If I add 3 items, and after remove the second item, then in dropdowns I would see the data from the 3rd, items instead of the 2nd item. See demo video https://monosnap.com/file/eWPhOjBU3CNRX091n5ehcYuVXJoqZN

    This issue usually happens because of using `key={jobIndex}`. Instead of this when creating default values for new row, add some new field, like `key = _.uuid('job_key_')` which would be unique, and use it for `key` instead of `key={jobIndex}`.
 
 I could continue reviewing after these 2 issues are fixed.

yoution:
@maxceem  
image
1. the api between  dev and my local , maybe not the same api, 
2. 
image
there is no entry for the job page

maxceem:
image

@yoution this is correct in `tass-jobs-2` branch we use a new endpoint for getting skills without categories.

image

Yeah, we hide it. But you can still use a direct link https://connect.topcoder-dev.com/new-project/taas-jobs

maxceem:
@yoution regarding showing skills. It was not your issue, I've fixed loading skills in the existent code. Please, pull the latest changes from `appirio-tech:feature/taas-jobs-2` I've used the fix there.

yoution:
@maxceem please review again

yoution:
sorry,I forget to change the fix message, I just use the previous message, next time I will pay attention to

maxceem (approved):
Great @yoution. Everything works good. Merging for testing by QA team.

PR #4246, opened 2021-01-06
Fix #4209: Empty message breaks showing all the messages

### LINKED ISSUE #4209: [$200] Empty message breaks showing all the messages
This error happened on PROD for the client project.

Create a post with the next body:
```js
{
            "id": 111111,
            "date": "2020-12-08T21:58:35.334Z",
            "updatedDate": "2020-12-08T21:59:04.087Z",
            "userId": 111111,
            "read": false,
            "body": " ",
            "rawContent": " ",
            "attachments": [],
            "type": "post"
          }
```

Load the Messages page with such a message. You would see no message and error in the console:

image

For local testing you may add the next code inside the file https://github.com/appirio-tech/connect-app/blob/574231d25b0f9a66c9b7e4bc254dd97c23880da5/src%2Fapi%2Fmessages.js and hardcode it as a response for `getTopics` API method.

```js
const rep = {
  "id": "8a237331-14a8-4561-943c-c7b4d7ef25f7",
  "version": "v5",
  "result": {
    "success": true,
    "status": 200,
    "content": [
      {
        "id": 1111111,
        "reference": "project",
        "referenceId": "25233",
        "tag": "PRIMARY",
        "date": "2020-10-27T12:18:51.950Z",
        "updatedDate": "2020-12-08T21:58:55.454Z",
        "lastActivityAt": "2020-12-08T21:58:55.447Z",
        "title": "test",
        "userId": 1111111,
        "totalPosts": 1,
        "retrievedPosts": 1,
        "postIds": [
          1111111
        ],
        "posts": [
          {
            "id": 1111111,
            "date": "2020-12-08T21:58:35.334Z",
            "updatedDate": "2020-12-08T21:59:04.087Z",
            "userId": 1111111,
            "read": false,
            "body": " ",
            "rawContent": " ",
            "attachments": [],
            "type": "post"
          }
        ],
        "read": false
      },
    ],
    "metadata": {
      "totalCount": 1
    }
  }
}
```

##  TODO:

1. This is not the first time when some particular message breaks the rendering of all messages. So we have to catch errors during rendering posts and don't let one bad message to break all the posts. When we fail to render one message, instead of the message we should show:
   - row code of the message, so users can still know what is inside
   - small error message, that we could not render the message properly, please contact Topcoder Support
2. Fix rendering method, so if we have an empty message (with spaces only) it doesn't get broken and just shows an empty message.
3. Don't allow posting empty messages (with space only).

sanju-singh:


maxceem (changes requested):
Works great @sanju-singh.

The only thing, we have to show the error message next to the message, like this:

image

You can use the color `$tc-orange-70`.

PR #4247, opened 2021-01-06
fix: issue #4234 for detail page

yoution:
@maxceem please review for #4234


maxceem (approved):
thanks, @yoution work good.

PR #4250, opened 2021-01-07
feat: delete _key for job talent form

yoution:
@maxceem please review for #4206 

I found the talent form also have the issue, when you add three design, and delete the middle one , the select will show the wrong value


maxceem:
@yoution this feels like not the best approach, as we can use component `JobsPickerQuestion` not only in the template with type `talent-as-a-service` but any other template.

Do you know any better solution so we don't have to bind this logic with project template type?

PR #4253, opened 2021-01-08
Empty msg break

sanju-singh:
Add error message when message could not be rendered.

### DIFF
--- a/src/components/Feed/FeedComments.jsx
+++ b/src/components/Feed/FeedComments.jsx
@@ -176,6 +176,21 @@ class FeedComments extends React.Component {
     }
   }
 
+  convertMdToHtml(markdown) {
+    try {
+      return (<div dangerouslySetInnerHTML={{__html: markdownToHTML(markdown)}} />)
+    } catch (e) {
+      return  ( 
+        <div>
+          <p>{markdown}</p>
+          <p styleName="comment-render-error">
+            This message could not be rendered properly, please contact Topcoder Support.
+          </p>
+        </div>
+      );
+    }
+  }
+
   render() {
     const {
       currentUser, onLoadMoreComments, isLoadingComments, hasMoreComments, onAddNewComment,
@@ -325,7 +340,7 @@ class FeedComments extends React.Component {
           canDelete={comments && (idx !== comments.length - 1)} // cannot delete the first post which is now shown as a last one
           commentAnchorPrefix={commentAnchorPrefix}
         >
-          <div dangerouslySetInnerHTML={{__html: markdownToHTML(itemContent)}} />
+          { this.convertMdToHtml(itemContent) }
         </Comment>
       )
     })
--- a/src/components/Feed/FeedComments.jsx
+++ b/src/components/Feed/FeedComments.jsx
@@ -187,7 +187,7 @@ class FeedComments extends React.Component {
             This message could not be rendered properly, please contact Topcoder Support.
           </p>
         </div>
-      );
+      )
     }
   }
 
--- a/src/components/Feed/FeedComments.scss
+++ b/src/components/Feed/FeedComments.scss
@@ -32,6 +32,14 @@
   padding-bottom: 2 * $base-unit;
 }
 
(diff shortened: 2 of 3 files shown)


maxceem:
Thanks, @sanju-singh. I didn't test it yet, but looking at the source code I have one concern. It seems that error would be catched inside `markdownToHtml` method, so the error message would be never shown:

image

sanju-singh:
@maxceem ,
1. Currently, to render messages on opening, markdownToState method is used to get the links in the messages to show link icon at the right side of message. It is not used to render message.
    - Earlier, due to empty comment, markdownToState method was failing at 'convertFromRaw' method.
    - As we are not sure of possible failures, I have added exception handling logic for convertFromRaw.

2. Secondly, raw comment in markdown is converted to html to render using markdownToHtml. This is the second place where unexpected break could happen. And this is the place that actually renders the method, if anything wrong goes here, we show the comment in raw text and show the above mentioned error. 

I have tested it by throwing the error from markdownToHtml method. And it is working as expected.

maxceem:
@sanju-singh thanks for the detailed explanation. Though as per my testing the error message is not shown when it suppose to be shown.

Imagine, that we don't fix the error with rendering empty messages, I comment the next lines:
image

So now, when empty message is rendered I'm expecting to see:
- empty message rendered as a raw text
- error message shown that message wasn't rendered properly
- though only empty message is shown without error message:

   image

Another way for testing it. Imagine that method `convertFromRaw` fails and thorws some error, I can simulate error by adding `throw` like this:
image

Now if, open a message I would see the message rendered as a raw text, which is correct. But when we render raw message instead of formatted message, we have to show an error message:
image

sanju-singh:
@maxceem ,
As I mentioned earlier, markdownToState method is not used to render the message, it's used to extract the links in the comment to show attachment icon like below:
Screenshot 2021-01-09 at 6 01 26 PM

1. Earlier convertFromRaw in  markdownToState method was breaking while extracting links. So, if any comment contains link, in case of break attachment icon will not show up except that everything will work as expected.
2. Secondly, markdownToState is used while switching to edit mode. Here also, after exception handling in markdownToState, in the edit mode raw text will show up that user can edit.

------
For above mentioned case.
1. Commenting empty message fix:
            - When actually rendering empty message using markdownToHTML, it converts it properly and renders it as expected.
            - What that fix is doing is preventing the exception thrown by convertFromRaw while extracting the links.
            - After commenting, exception handling at the bottom is doing its job and preventing from any break, and comment render logic is anyway working fine.

2. Throwing exception in markdownToHTML to mimic exception in convertFromRaw.
        - It doesn't affect anything, as this method is used just to extract links.
        - Even we throw error, catch is handling it and we will end up not extracting any links even if comment had it.
        - Rendering logic will not be affect by this, I checked the rendering of blockquote earlier to my fix and it renders like that only. Please use other markdown options to see the change like: bold, list etc.
        
        
Conclusion: Solution is working as expected with both the above suggested approach. Blockquote renders like raw text only . Please check with other markdown formatting like: bold, list etc.

maxceem:
@sanju-singh Thank you, got it now.

The task is fixed correctly as per the initial description. 

Though it reveled one more issue. The method `markdownToState` is used not only to extract links but also to format message during editing. If some error happens in `markdownToState` during editing we would see a raw message inside the editor, which might be confusing.

image

It would be good in such a case to show the same error message during editing like this:

image

If you like you may fix this case as a separate issue https://github.com/appirio-tech/connect-app/issues/4257 - price $40.

And the current issue is fixed and accepted.

sanju-singh:
@maxceem ,
As its a markdown editor, showing in raw format (which is in markdown only) under exceptional scenario is valid. And cannot be categories as issue.

Showing error here will be somewhat difficult, as markdownToState method return signature doesn't allow passing error and we are handling the error there itself.

let me know if you still think that we should show error there?

maxceem:
@sanju-singh Thank you for sharing your thoughts. Let me think about it a little bit more.

Issue #4254, opened 2021-01-08
[$20] Update the "Submitted Screen" for TaaS Projects


maxceem:
Follow up from https://github.com/appirio-tech/connect-app/issues/4194#issuecomment-756802537

If we created a project `type=PROJECT_TYPE_TALENT_AS_A_SERVICE` we should show different text on the project creation page. See for example, how we already use different links in source code https://github.com/appirio-tech/connect-app/blob/feature/taas-jobs-2/src/projects/create/components/ProjectSubmitted.jsx#L11-L25.

For the TaaS Projects the texts should like these:
Header: "Congratulations"
Header Sub-Text: "Your talent request has been created"

Main Body Text: "Topcoder will be contacting you soon to discuss your talent needs."

Button 1:
Button Header Text: "View Talent Request"
Button Sub-Text: "Modify your request and track fulfillment"

Button 2:
Button Header Text: "All Projects"
Button Sub-Text: "View all of your projects"

This has to be done in branch `feature/taas-jobs-2`.

maxceem:
Contest https://www.topcoder.com/challenges/30161685 has been created for this ticket.<br/><br/>```This is an automated message for maxceem via Topcoder X```

maxceem:
@yoution this is open for pick up.

yoution:
@maxceem please assign to me

yoution:
![image](https://user-images.githubusercontent.com/3863914/104081817-38556c00-526c-11eb-8e36-ec26f1619f0c.png)

how about this text? delete it?

maxceem:
@yoution yes, remove this text.

maxceem:
Contest https://www.topcoder.com/challenges/30161685 has been updated - it has been assigned to yoution.<br/><br/>```This is an automated message for maxceem via Topcoder X```

maxceem:
@acshields one small clarification, should "All Projects" button lead to the list of all projects in Connect App, or to the list of all Teams in TaaS App?

![image](https://user-images.githubusercontent.com/146016/104091130-0ea83f00-5284-11eb-9bf6-f425a46d0424.png)

acshields:
@maxceem - Have it lead to All Projects in Projects App (Connect).

drasticdpk:
verified.
![image](https://user-images.githubusercontent.com/25297859/104301099-5979bf00-54ef-11eb-9385-a36b5a958ac8.png)

maxceem:
[...]

PR #4255, opened 2021-01-09
fix:issue #4254

yoution:
@maxceem  please for review #4254


maxceem (approved):
Thanks for a good structure @yoution. Works well.

PR #4261, opened 2021-01-12
fix: #4206 add trim()

yoution:
@maxceem  please review for #4206


maxceem (approved):
> User should not able to put space char in text box which is mandatory fields.

@yoution this part is done.

PR #4264, opened 2021-01-13
fix: #4206 disblae next button when add empty job

yoution:
@maxceem please review for #4206


maxceem (approved):
@yoution works great.

Issue #4273, opened 2021-01-14
"Project Defaults": Terms Display & Management


acshields:
### Update A
Display the project's terms options as follows on the "Project Defaults" screen:
- Field Label: "Enforce Topcoder NDA"
- Field Data: [Yes/No Radio Buttons]
-Enforce Topcoder NDA [Hyperlinked to Topcoder NDA: https://www.topcoder.com/challenges/terms/detail/c41e90e5-4d0e-4811-bd09-38ff72674490]

### Update B
The user should be able to change selected defaults.

vikasrohit:
@acshields I don't think we need the Update 2 in this story, I don't see any relation between Terms and BA.

acshields:
@vikasrohit - correct. I was copy/pasting content and failed to delete that part. I have updated the spec.

maxceem:
We would have to define NDA Terms ID in the config for DEV and PROD.

For DEV, we can use this one `7245bb7d-d7c9-45a0-9603-d5ff05af0977` (Appirio NDA v2.0) or you may suggest another one.

For PROD we need this to be provided.

maxceem:
@vikasrohit we have some API issue with terms.

Projects have a property `terms: []` which accepts the array numbers, as legacy terms had number ids.
New V5 Terms have ids as uuids like "7245bb7d-d7c9-45a0-9603-d5ff05af0977".

So if I try to set V5 terms ids, Projects API would return error:
```js
{
    "message": "validation error: \"0\" must be a number"
}
```

There are several things to consider:
1. We can update the validation rule and accept strings for `terms` but should we do this? Because as a result the `terms` property would contain old terms ids for old projects and new terms ids for new projects. Wouldn't it break third party services that might consume this field value?

2. I'm not sure if ES index would allow storing string in this array as we already storing numbers in this array. I guess yes, but we have to check it.

vikasrohit:
> For PROD we need this to be provided.

We can copy it from https://github.com/topcoder-platform/work-manager/blob/develop/config/constants/production.js#L27

For `terms` field in Projects API, I don't think we are using that anywhere, it is a left over of initial projects api and I don't think we would have any project with any real data in that field. So, we should be safe to refactor that field now. We can do complete reindexing to change the data types.

maxceem:
Sum up:
- if user enables NDA, then we should add NDA term to the terms array of the project `terms`. If use disables NDA, then we should remove NDA terms from the `terms` array if present.
[...]

PR #4286, opened 2021-01-28
project defaults page

CDharmateja:
#4273


maxceem (commented):
Thank you, @CDharmateja this PR works good.

Though after reviewing it I can see that we have to create a new component for the NDA radio-boxes. I'm merging this PR as it works good and we are short on time. But I would provide you details how we should improve its code by creating a separate component.

Issue #4270, opened 2021-01-14
"Project Defaults": Groups Display & Management


acshields:
### Update A
Display the project's group question as follows on the "Project Defaults" screen:
- Field Label: "Intended Work Groups"
- Field Data: [Responsive Text Field]

### Update B
When the user types into the responsive text field, the returned options should only be the groups that they have access too.

vikasrohit:
@acshields we don't get Salesforce BA id when we lookup from our platform's BA API. So, we would need changes in SFDC to lookup for the BA in SFDC based on provided platform BA id.

acshields:
@vikasrohit - was this comment supposed to be in regards to the issue(s) for billing account, and not for groups? If it was for groups, I am not understanding how a BA plays a role for that.

vikasrohit:
@acshields sorry for the confusion, yes your are right my comment was in context of billing account https://github.com/appirio-tech/connect-app/issues/4269. I would comment in that issue for my intended concern.

However, I guess Update 2 is not necessary in this issue as well because groups does not have link to BAs as well.

acshields:
@vikasrohit - updated to remove original "Update 2." Now reqs are labeled Update A and B and they should be accurate.

maxceem:
@vikasrohit at the moment Project object doesn't have a designated place to store assigned groups.

We have 2 ways to go:
1. Create a new field `groups` in the `project` model to store the list of Groupd ids as array of string. See how we store `terms` https://github.com/appirio-tech/connect-app/issues/4273#issuecomment-765292459.
2. Store groups in the `details` field which is JSON and we can add any fields there.

I guess groups are kind of fundamental thing of projects and would hardly be dropped, so creating a separate field might have sense.

vikasrohit:
Yes, agreed. we should make it first class citizen of the Project object because we might need to query projects based on groups.

maxceem:
@vikasrohit I guess I found where we previously stored groups `challengeEligibility.groups` https://github.com/topcoder-platform/tc-project-service/blob/develop/src/events/projects/index.js#L114.

Though this field also only accepted `numbers`.

Should we update the type of it and reuse it? Or add a new one, what do you think?

vikasrohit:
[...]

PR #4294, opened 2021-02-04
Project Defaults, part 2

CDharmateja:
- Refactor NDAField implementation #4273 
- Add Groups Field https://github.com/appirio-tech/connect-app/issues/4270

### DIFF
--- a/src/actions/loadUser.js
+++ b/src/actions/loadUser.js
@@ -17,7 +17,7 @@ import {
 } from '../config/constants'
 import { getFreshToken, configureConnector, decodeToken } from 'tc-auth-lib'
 import { getUserProfile } from '../api/users'
-import { getUserGroups } from '../api/groups'
+import { fetchGroups } from '../api/groups'
 import { getOrgConfig } from '../api/orgConfig'
 import { EventTypes } from 'redux-segment'
 
@@ -147,7 +147,10 @@ export function loadUserFailure(dispatch) {
  */
 function loadGroups(dispatch, userId) {
   if (userId) {
-    getUserGroups(userId, 'user').then((groups) => {
+    fetchGroups({
+      memberId: userId,
+      membershipType: 'user'
+    }).then((groups) => {
       const groupIds = _.compact(_.uniq([
         // get old and new ids as organizations may refer to any of them
         ..._.map(groups, 'oldId'),
--- a/src/api/groups.js
+++ b/src/api/groups.js
@@ -4,13 +4,13 @@ import { TC_API_URL } from '../config/constants'
 /**
  * Get a user groups based on it's member id and membership type
  *
- * @param {String} memberId user member id
+ * @param {Object} query fetch group query
  *
- * @param {String} membershipType user membership type
+ * @param {String} params fetch group url params
  *
  * @returns {Promise<Object>} user groups data
  */
-export function getUserGroups(memberId, membershipType) {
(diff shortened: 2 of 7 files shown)


maxceem (changes requested):
Thank you for the timely update @CDharmateja. There are some improvements I'd like to make.

1. Once we save the form, the "Save" button never gets enabled again, see demo video https://monosnap.com/file/fPtLTnJcbqCt0kzGLpz8FGe9NToRGS

2. Can we, please, add prop definitions for the NDAField, like this https://github.com/appirio-tech/connect-app/blob/9cb0ed8e9dbf09aa4fab6cb25e11b6525ff6156f/src/projects/detail/components/AddonOptions/AddonOptions.jsx#L117-L125

3. Looks like `this.handleChange` is called twice
     image

4. Sorry, that I was not clear there. But can we keep 
   ```<SpecQuestionList>
              <Accordion
                title="Enforce Topcoder NDA"
                type="radio-group"
                options={opts}
              >
                <SpecQuestionList.Item>
     ```
     outside `NDAField` so we can use `NDAField` wrapper in the accordion, or if we don't need it, we can use it without an accordion.

CDharmateja:
@maxceem  For 4th one I am having issues completing. Accordion is showing value 'Yes' or 'No' from its child formsy field. If NDAField is child component instead of RadioGroup, then the value passed to NDAField is used by Accordion instead of value of RadioGroup.

maxceem:
Thanks, @CDharmateja the form works great now.

> For 4th one I am having issues completing. Accordion is showing value 'Yes' or 'No' from its child formsy field. If NDAField is child component instead of RadioGroup, then the value passed to NDAField is used by Accordion instead of value of RadioGroup.

Got it. Let's then remove Accordion completely for this form, as it makes everything harder. Please, remove `SpecQuestionList`, `Accordion` and `SpecQuestionList.Item` from `NDAField` and from `EditProjectDefaultForm`

image

and just add "label" for `RadioGroup` with value `Enforce Topcoder NDA`.

It would look like this, which is fine:

image


Also, I've noticed that there is a lint error. Please, fix them before I can merge it (don't disable lint rules).

image

maxceem (changes requested):
Thank you @CDharmateja. Works almost perfectly. A few notes:

1. When I add or remove a group in the group select, there is API request is sent for each individual group, even though we already know the group name from the suggestions. Can we avoid it? As I understand we only need to request individual groups on page load to show their names. But when we edit them, we already know group names from the suggestions and don't have to make additional requests. See video https://monosnap.com/file/jozUeiuqy7WIdcjDBRe9OXFnHdNpy0

2. The label for the Groups Field should be "Intended Work Groups" instead of "Groups Field"

3. When showing label for NDAField you are using `<fieldset>` with `<legend>`, while for showing label for Group you just use `<div className={styles.container}>` with `<div className={styles.fieldName}>`. Is there any need in making it different ways? Or we can follow the same way in both cases?

CDharmateja:
For 3, I used `fieldset` and `legend` for radio group since they have semantic meaning make more sense from accessibility point of view.

maxceem (approved):
> For 3, I used fieldset and legend for radio group since they have semantic meaning make more sense from accessibility point of view.

Makes sense.

Thank you @CDharmateja for taking care of all the feedback. Works perfectly now.

Issue #4269, opened 2021-01-14
"Project Defaults": Billing Account Display & Management


acshields:
### Update 1
Display the project's billing account as follows on the "Project Defaults" screen:
- Field Label: "Billing Account"
- Field Data: [Billing Account ID and Name populated in responsive dropdown menu]
- Field Sub-Text Label: [Hyperlinked: "Manage the billing account in Salesforce"]

### Update 2
Only these users should see the Field Sub-Text Label: [Hyperlinked: "Manage the billing account in Salesforce"]:
- Admin
- Connect Manager
- Copilot Manager
- Project Manager
- Program Manager
- Account Manager
- BDR
- Account Executive
- Presales
- Solution Architect

### Update 3
Ensure that the hyperlink can do the following:
- Hyperlink to the project's lead in Salesforce, if it has not been converted to an opportunity
- Hyperlink to the project's opportunity in Salesforce

### Update 4
From the Salesforce API, retrieve:
All billing account ID and names that the user has access to.

vikasrohit:
@acshields 

I am not sure why we need Update 3 because we are already showing a link to manage the BA along with the dropdown. Why would we need another link for linking to Lead or Opty now? In case, we need it where to show it? We already have that in left panel right now.

vikasrohit:
@acshields we have to decide on how to pull the available BA for showing options because it would increase the scope if we have to alter the BA api or implement some changes in SFDC to get this work done.

acshields:
@vikasrohit
**Feedback on Update 3**
What I am saying here is that the "Manage billing account in Salesforce" hyperlink should be smart enough to link to the project's lead, or opportunity, depending on what state it is in in Salesforce.
**Feedback on BA Workflow**
To modify the current BA API so that it can retrieve all active BAs associated with a project in SFDC, you said that changes would be needed in SFDC. To accomplish this retrieval goal to display all project BA options in a dropdown menu, what is the total estimate?

vikasrohit:
> What I am saying here is that the "Manage billing account in Salesforce" hyperlink should be smart enough to link to the project's lead, or opportunity, depending on what state it is in in Salesforce.

[...]

PR #4299, opened 2021-02-14
Billing account field in project defaults

CDharmateja:
#4269


src/api/billingAccounts.js
```diff
@@ -0,0 +1,14 @@
+import { axiosInstance as axios } from './requestInterceptor'
+import { TC_API_URL } from '../config/constants'
+
+/**
+ * Get a user groups based on it's member id and membership type
```
maxceem, line 5: Description from another method.

src/projects/detail/components/BillingAccountField/styles.module.scss
```diff
@@ -0,0 +1,5 @@
+.container {
+  display: flex;
+  flex-direction: column;
+  gap: 10px;
```
maxceem, line 4: `gap` is not supported by Safari on macOS:

image

Can we, please, avoid using it.

src/projects/detail/components/EditProjectDefaultsForm/EditProjectDefaultsForm.scss
```diff
@@ -6,6 +6,9 @@
     .container {
       padding: 50px 0 70px 0;
       margin: 0 50px 20px 50px;
+      display: flex;
+      flex-direction: column;
+      gap: 20px;
```
maxceem, line 11: `gap` is not supported by Safari on macOS:

image

Can we, please, avoid using it.

src/projects/detail/components/GroupsField/GroupsField.module.scss
```diff
@@ -1,6 +1,5 @@
 .container {
-  margin: 10px 0;
-}
-.fieldName {
-  margin-bottom: 10px;
+  display: flex;
+  flex-direction: column;
+  gap: 10px;
```
maxceem, line 4: `gap` is not supported by Safari on macOS:

image

Can we, please, avoid using it.

src/projects/detail/components/BillingAccountField/index.js
```diff
@@ -0,0 +1,98 @@
...
+    return (
+      <div className={styles.container}>
+        <div>Choose a Billing Account</div>
+        <Select
+          placeholder={
+            this.state.billingAccounts.length > 0 
+              ? 'Select billing account'
+              : 'No Billing Account Available'
+          }
+          heightAuto
```
maxceem, line 79: I guess we only need `heightAuto` if we have a select with multiple values that can go to the several lines. Can we remove it here?

src/projects/detail/components/BillingAccountField/index.js
```diff
@@ -0,0 +1,98 @@
...
+    fetchBillingAccounts(this.props.projectId)
+      .then(({data: billingAccounts}) => this.setState({billingAccounts}))
+      .then(() => {
+        if (this.props.billingAccountId) {
+          this.setState(state => {
+            const filteredArray = state.billingAccounts.filter(
+              ({tcBillingAccountId}) =>
+                tcBillingAccountId === this.props.billingAccountId
+            )
+            if (filteredArray.length === 0) {
```
maxceem, line 33: Technically this works correctly, but logically the filter method suppose to filter array and return array. Though our intention here is to find out if element exists in array or no. So it would be preferable to use here a `lodash` method `_.find`, something like this:

```js
const existentBillingAccount = _.find(state.billingAccounts, { tcBillingAccountId: this.props.billingAccountId })

if (!existentBillingAccount) {
```

maxceem (changes requested):
Thank you, @CDharmateja it works great.

There are just minor improvements I'd like to make to the code, as per comments below.

Also, can we please style the link "Manage the billing account in Salesforce" the same like the support email (small and blue):

image

maxceem (approved):
Thank you @CDharmateja. All is perfect now.

Issue #4298, opened 2021-02-11
TaaS app : New req changes for description fields


nkumar-topcoder:
- For `TaaS Intake form` : convert (job) Description free format field to marksdown editor.
the editor should be  similar to **workmanager** `public specification`

![image](https://user-images.githubusercontent.com/33625707/107643075-b1951400-6c9b-11eb-89e4-9a760850d4ad.png)

cc : @RishiRajSahu @maxceem @wdprice @acshields

maxceem:
On the one hand we already have a RichTextEditor which we use inside Connect App to edit messages:
![image](https://user-images.githubusercontent.com/146016/107644362-01fe7880-6c80-11eb-971f-b9a590ce22e8.png)

Though this editor has one big drawback - it doesn't work on mobile devices. So on mobile devices, we are falling back to a plain textarea field and I think this is not acceptable for the project intake form which is a crucial thing to retrieve requirements from the customers and should work flawlessly.

So I guess for the Intake Forma we have to use another editor, for example, the one we use in Work Manager, we have to make sure it works well on mobile devices too.

We could also consider using Markdown editor which we use on the new Topcoder Forums it also looks good (maybe it's the same as in WM):

![image](https://user-images.githubusercontent.com/146016/107644671-66213c80-6c80-11eb-95b6-af729335df4e.png)

maxceem:
Sum up:

TaaS Intake form available by link:
- DEV https://connect.topcoder-dev.com/new-project/taas-jobs
- PROD https://connect.topcoder-dev.com/new-project/talent-as-a-service

Should use Markdown editor to edit the `Description` field:

![image](https://user-images.githubusercontent.com/146016/108199579-eaaf0780-7125-11eb-8780-5df9df5ed333.png)

- Update component [JobPickerRow](https://github.com/appirio-tech/connect-app/blob/dev/src/projects/detail/components/JobPickerRow/JobPickerRow.jsx#L19) so it shows Markdown editor for the Duration field.
- Create a reusable component for the Markdown text
- Description filed should have a bigger height like here https://challenges.topcoder-dev.com/projects/16839/challenges/9b129be0-6d68-4bd8-8211-65e5f994e108/edit

    ![image](https://user-images.githubusercontent.com/146016/108199832-437ea000-7126-11eb-9e12-00dc8e5d6543.png)

- For markdown editor use https://www.npmjs.com/package/easymde or suggest a better alternative.

maxceem:
[...]

PR #4302, opened 2021-02-18
fix: add easymde #4298

yoution:


yoution:
@maxceem please review

src/components/MarkdownText/MarkdownText.jsx
```diff
@@ -0,0 +1,75 @@
...
+        }
+      })
+    } else {
+      this.ref.current.innerHTML = value ? marked(value) : ''
+    }
+  }
+
+  render () {
+    const { isPrivate, readOnly, placeholder, className } = this.props
+    return (<div className={cn(className, styles.editor, { [styles.isPrivate]: isPrivate })}>
```
maxceem, line 50: I'm not sure what does`isPrivate` do?

src/components/MarkdownText/MarkdownText.jsx
```diff
@@ -0,0 +1,75 @@
...
+import _ from 'lodash'
+
+class MarkdownText extends Component {
+  constructor (props) {
+    super(props)
+    this.state = {
+      isChanged: false
+    }
+    this.blurTheField = this.blurTheField.bind(this)
+    this.updateDescriptionThrottled = _.throttle(this.updateDescription.bind(this), 10000) // 10s
```
maxceem, line 16: Because of this `onChange` would be called only once per 10 seconds. This seems like not a good default behavior. When we reuse this component other logic might want to know the fresh value immediately.

maxceem (changes requested):
@yoution generally looks good, but it has a few issues for me:

1. If we have an empty field we show a validation error that is correct. When type one character I expect that the validation error is gone, but it doesn't. And after removing it, the validation error is gone, but it shouldn't. For example, check how title field works and how the description with editor works. It should work the same like title, see demo video https://monosnap.com/file/j1aSlQNabfysgrebw1W42qwL7rwzRy

2. If I type a lot inside the text editor and scrollbar appears, then I can also scroll editor buttons and bottom summary information. But I should only scroll what is inside editor. See demo video https://monosnap.com/file/NqVLbDBXgFY875RB3pHt02BR6TkUsn. And check how it's done in Work Manager for example 

     image

3. We already use a package to convert Makrdown to HTML https://www.npmjs.com/package/remarkable. Can we reuse it instead of introducing a new package `marked`.  We even have a method which we maybe can reuse [markdownToHTML](https://github.com/appirio-tech/connect-app/blob/dev/src/helpers/markdownToState.js#L215)

4. If we click "Preview Mode" then text would almost have no styles:

    image
    
    Can we have some basic styles for the features that we use: headings, bold/italic, ordered/unordered lists, blockquote, code? Similar to what we have in Work Manager:
    
    image

Example code:
```
# h1 Heading
## h2 Heading
### h3 Heading
#### h4 Heading
##### h5 Heading
###### h6 Heading

## Emphasis

**This is bold text**

__This is bold text__

*This is italic text*

_This is italic text_

~~Strikethrough~~


## Blockquotes


> Blockquotes can also be nested...
>> ...by using additional greater-than signs right next to each other...
> > > ...or with spaces between arrows.


## Lists

Unordered

+ Create a list by starting a line with `+`, `-`, or `*`
+ Sub-lists are made by indenting 2 spaces:
  - Marker character change forces new list start:
    * Ac tristique libero volutpat at
    + Facilisis in pretium nisl aliquet
    - Nulla volutpat aliquam velit
+ Very easy!

Ordered

1. Lorem ipsum dolor sit amet
2. Consectetur adipiscing elit
3. Integer molestie lorem at massa


1. You can use sequential numbers...
1. ...or keep all the numbers as `1.`

Start numbering with offset:

57. foo
1. bar


## Code

Inline `code`

Indented code

    // Some comments
    line 1 of code
    line 2 of code
    line 3 of code
```


Also, there are few comments for the code below.

src/components/MarkdownText/MarkdownText.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+class MarkdownText extends Component {
+  constructor (props) {
+    super(props)
+    this.state = {
+      isChanged: false
+    }
+    this.blurTheField = this.blurTheField.bind(this)
+    this.updateDescriptionThrottled = this.updateDescription.bind(this)
+    // this.updateDescriptionThrottled = _.throttle(this.updateDescription.bind(this), 10000) // 10s
```
maxceem, line 15: Can we remove this now and rename `updateDescriptionThrottled` to `updateDescription`?

src/components/MarkdownText/MarkdownText.jsx
```diff
@@ -0,0 +1,72 @@
...
+      this.easymde.codemirror.on('change', () => {
+        this.setState({ isChanged: true })
+        this.updateDescriptionThrottled(this.easymde.value())
+      })
+      this.easymde.codemirror.on('blur', () => {
+        if (this.state.isChanged) {
+          this.setState({ isChanged: false })
+          this.blurTheField()
+        }
+      })
```
maxceem, line 41: Do we still need this logic after we don't use `throttle`? Or can it be removed together with `this.setState({ isChanged: true })`?

src/components/MarkdownText/MarkdownText.jsx
```diff
@@ -0,0 +1,72 @@
...
+    } else {
+      this.ref.current.innerHTML = value ? markdownToHTML(value) : ''
```
maxceem, line 43: This doesn't work. I guess it should be `this.refs.textarea.innerHTML = value ? markdownToHTML(value) : ''`

maxceem (changes requested):
1. You've implemented the `readOnly` mode which is nice to have. But if I enable it (set `readOnly = true`) then there are is an error in the console and the component doesn't work. It can be fixed by `this.refs.textarea.innerHTML = value ? markdownToHTML(value) : ''` and then readonly mode works, but styles looks broken:
   - height should be automatic or with a scrollbar
   - content should not go out
   - content should have the style as when preview

   image
 
 
@yoution I wonder why did you add `throttle` initially? Does typing work slowly for you without `throttle`? Or what was the reason? maybe I'm missing something?

src/components/MarkdownText/MarkdownText.jsx
yoution: I don't know why I add the `throttled`, I copy some logic from the * workmanager* project

maxceem (approved):
All is perfect now. Thank you @yoution for the quick turn arounds.

Issue #4297, opened 2021-02-11
Hide Project Default Menu/Screen for Customer Role


drasticdpk:
**Reference** 
![image](https://user-images.githubusercontent.com/25297859/107638072-33367300-6c97-11eb-9670-0eb218f74687.png)

![image](https://user-images.githubusercontent.com/25297859/107638562-d8e9e200-6c97-11eb-8d09-1528a37b3525.png)

cc @acshields

acshields:
@drasticdpk - agreed.

vikasrohit:
@drasticdpk are you suggesting that https://github.com/appirio-tech/connect-app/issues/4268 is failing because the tab is visible for customers?

If yes, @maxceem I guess we need to fix it.

maxceem:
Fixed via https://github.com/appirio-tech/connect-app/commit/2d04a30779397a22ddf4d586a1f64aa6e509214c and ready for QA in DEV:

![image](https://user-images.githubusercontent.com/146016/107846698-50ca2080-6dee-11eb-8f31-9f0c8a110a03.png)

drasticdpk:
@maxceem , Let's  disable with direct link too for customer role

**Reference** 
![image](https://user-images.githubusercontent.com/25297859/108045617-c4c52c80-706b-11eb-8d58-f8682ac86b7c.png)

maxceem:
Task:
- Users who cannot see **Project Default** menu item, should not be able to access the page Project Defaults by the direct link https://connect.topcoder-dev.com/projects/17076/settings.

- If user doesn't have permissions, they should see the page with error message like we show for the metadata pages https://connect.topcoder-dev.com/metadata/projectTemplates (only admins can access this page):
   
   ![image](https://user-images.githubusercontent.com/146016/108054831-d7356b00-7057-11eb-8b0a-9f0a6d43ed2c.png)

- Implement it in some general way. Create some [HOC](https://reactjs.org/docs/higher-order-components.html#gatsby-focus-wrapper) which we could apply to other pages inside Project if we want.

- use method `hasPermission` to check permissions the same way as we do when checking permission for the menu item.

maxceem:
@drasticdpk This is done by https://github.com/appirio-tech/connect-app/pull/4305 and ready for QA on DEV.

drasticdpk:
Verified on dev
![image](https://user-images.githubusercontent.com/25297859/108684995-b4093080-751b-11eb-8e74-41f67b745284.png)

[...]

PR #4305, opened 2021-02-19
hide project defaults page for customer role

CDharmateja:
#4297


maxceem (approved):
Thanks, @CDharmateja, works good.

Issue #4249, opened 2021-01-07
[$40] Hide Change password/email functionality for Social Logged in user


sachin-maheshwari:
Please find detail here - https://github.com/topcoder-platform/tc-auth-lib/issues/29

maxceem:
Sum up:
- we can handle it the same way as it's done in Community App on the page https://www.topcoder-dev.com/settings/account

   ![image](https://user-images.githubusercontent.com/146016/109490024-52ebda80-7a90-11eb-9af4-01b4ced24cf5.png)
   
- if a user is logged-in using SSO Service, then 
   - disable email field so we cannot change it
   - show red warning message similar to Community App
   - hide section for changing the password

   ![image](https://user-images.githubusercontent.com/146016/109490188-8fb7d180-7a90-11eb-9e3f-12509d39abbb.png)
   
- on how to detect SSO login see community app source code https://github.com/topcoder-platform/community-app
- you can create a new DEV account for testing using this link https://accounts-auth0.topcoder-dev.com/?retUrl=https://www.topcoder-dev.com/start&mode=signUp (NOTE: you have to sign up using Google or GitHub)

maxceem:
Challenge https://www.topcoder.com/challenges/27d1ab70-77c7-4ff5-bfdd-6cbcc7007f5c has been created for this ticket.<br/><br/>```This is an automated message for maxceem via Topcoder X```

yoution:
@bug-bash-helper assign me

maxceem:
Challenge https://www.topcoder.com/challenges/27d1ab70-77c7-4ff5-bfdd-6cbcc7007f5c has been assigned to yoution.<br/><br/>```This is an automated message for maxceem via Topcoder X```

yoution:
@bug-bash-helper    https://github.com/appirio-tech/connect-app/pull/4322 is ready for review

drasticdpk:
**Verified with Github login**
![image](https://user-images.githubusercontent.com/25297859/110063173-baf13800-7d92-11eb-87ff-afff7f569a60.png)

maxceem:
@drasticdpk thanks for the report. I cannot reproduce it myself. Could you please check one server request?

In the dev tools there should one request which ends `?fields=credentials` and in response there should be value `hasPasssowrd`, what value do you have there?

![image](https://user-images.githubusercontent.com/146016/110465039-98d21f80-80dc-11eb-8c71-fbc1ad7763ba.png)

[...]

PR #4322, opened 2021-03-01
fix: issue #4249

yoution:


src/routes/settings/routes/system/containers/SystemSettingsContainer.jsx
```diff
@@ -16,7 +17,13 @@ const FormEnhanced = enhance(SystemSettingsForm)
...
 class SystemSettingsContainer extends Component {
   componentDidMount() {
-    this.props.getSystemSettings()
+    const {
+      getSystemSettings,
+      getUserCredential,
+      user
+    } = this.props
+    getSystemSettings()
+    getUserCredential(user.userId)
```
maxceem, line 26: nice update to get variables from `this.props` instead of adding `this.props` every time 👍

src/routes/settings/routes/system/components/SystemSettingsForm.jsx
```diff
@@ -40,20 +40,28 @@ class SystemSettingsForm extends Component {
...
 
-        <div styleName="section-heading">
+        {!usingSsoService ?
+          <div styleName="section-heading">
           Retrieve or change your password
-        </div>
+          </div>: null}
 
-        <div className="form">
+        {!usingSsoService ? <div className="form">
```
maxceem, line 58: this could be simplified as `{!usingSsoService && <div className="form">`. No need for ternary operator `.. ? .. : null`.

src/routes/settings/routes/system/components/SystemSettingsForm.jsx
```diff
@@ -40,20 +40,28 @@ class SystemSettingsForm extends Component {
...
+
+          {usingSsoService ? <div styleName="error-message">
+            Since you joined Topcoder using your &lt;SSO Service&gt; account,
+            any email updates will need to be handled by logging in to
+            your &lt;SSO Service&gt; account.
+          </div>: null}
         </div>
 
-        <div styleName="section-heading">
+        {!usingSsoService ?
```
maxceem, line 53: this could be simplified as `{!usingSsoService &&`. No need for ternary operator `.. ? .. : null`.

src/routes/settings/routes/system/components/SystemSettingsForm.jsx
```diff
@@ -40,20 +40,28 @@ class SystemSettingsForm extends Component {
             checkEmailAvailability={checkEmailAvailability}
             onSubmit={(email) => changeEmail(email)}
             {...systemSettings}
+            disabled={usingSsoService? true: systemSettings.disabled === true}
           />
+
+          {usingSsoService ? <div styleName="error-message">
```
maxceem, line 46: this could be simplified as `{usingSsoService && <div styleName="error-message">`. No need for ternary operator `.. ? .. : null`.

src/routes/settings/routes/system/components/SystemSettingsForm.jsx
```diff
@@ -40,20 +40,28 @@ class SystemSettingsForm extends Component {
             checkEmailAvailability={checkEmailAvailability}
             onSubmit={(email) => changeEmail(email)}
             {...systemSettings}
+            disabled={usingSsoService? true: systemSettings.disabled === true}
```
maxceem, line 43: To keep things clear, I suggest passing `usingSsoService` to the `ChangeEmailForm` component and use it inside. Because we have various reasons when we disable form. Also, this condition is a bit hard to ready.

src/routes/settings/routes/system/components/ChangeEmailForm.jsx
```diff
@@ -140,7 +140,7 @@ class ChangeEmailForm extends React.Component {
               validationErrors={{
                 isEmail: 'Provide a correct email'
               }}
-              disabled={isEmailChanging || !hasPermission(PERMISSIONS.UPDATE_USER_EMAIL)}
+              disabled={disabled ||isEmailChanging || !hasPermission(PERMISSIONS.UPDATE_USER_EMAIL)}
```
maxceem, line 143: As I suggest in another comment, let's not pass `disabled` here, but pass `usingSsoService` directly here, so we can understand more clearly why we disable this field.

src/api/users.js
```diff
@@ -17,6 +17,23 @@ export function getUserProfile(handle) {
     })
 }
 
+
+/**
+ * Gets credential for the specified user id.
+ *
+ * NOTE: Only admins are authorized to use the underlying endpoint.
```
maxceem, line 24: This is not completely true. Also, any user can get this data but only for the themselves.

maxceem (changes requested):
@yoution it works good, but there are some small improvements I'd like to make.

1. We have to wait until the user credentials are loaded, to make sure that when we show the form we already know is user uses SSO login or no. See how we check if we still loading setting or no https://github.com/appirio-tech/connect-app/blob/dev/src/routes/settings/routes/system/containers/SystemSettingsContainer.jsx#L15.

2. There are some small code improvements to make. Please, kindly check comments below.

maxceem (approved):
@yoution all is perfect now.

Issue #4301, opened 2021-02-18
[$50] Invitations: Typed Handle/Email Deletes if User Doesn't hit "Enter" Immediately


acshields:
**Current Behavior**
I type in a user's email address (or handle), but before I hit "enter" I click on the white space of the invitation pop-up window. The typed information disappears and I have to retype the information. 

**Desired Behavior**
The typed information should save if the user:
- Hits "enter"
- Hits "space"
- Clicks on the pop-up window

maxceem:
Sum up:
- I'm not sure if there is a solution for this task. The payment would be processed only if there is a solution provided.

maxceem:
Challenge https://www.topcoder.com/challenges/3b86da61-87ed-4c88-8fcd-2f763e45f750 has been created for this ticket.<br/><br/>```This is an automated message for maxceem via Topcoder X```

yoution:
@bug-bash-helper assign me

maxceem:
Challenge https://www.topcoder.com/challenges/3b86da61-87ed-4c88-8fcd-2f763e45f750 has been assigned to yoution.<br/><br/>```This is an automated message for maxceem via Topcoder X```

yoution:
@maxceem which page can reproduce the issue?

yoution:
@maxceem https://monosnap.com/file/0Vo9vJRgefMiibAp7PzfA7vvKQULoU 
how about this approach

yoution:
@bug-bash-helper   https://github.com/appirio-tech/connect-app/pull/4323 is ready for review

coderhacker:
@drasticdpk @lakshmiathreya this is passed on dev

https://user-images.githubusercontent.com/62358241/110065286-d2312500-7d94-11eb-8c83-97bad13de3f9.mp4

drasticdpk:
@lakshmiathreya  we need your help  here to verify this issue. Thank you!

lakshmiathreya:
Invitations don't get deleted if cursor moves away from the field without the 'Enter':
<img width="1680" alt="Screenshot 2021-03-17 at 8 34 12 PM" src="https://user-images.githubusercontent.com/51733042/111490510-180cc680-8761-11eb-9d6b-f5481ee31bc6.png">
<img width="1680" alt="Screenshot 2021-03-17 at 8 33 52 PM" src="https://user-images.githubusercontent.com/51733042/111490491-13481280-8761-11eb-90f1-9cc9c11c3098.png">

PR #4323, opened 2021-03-01
fix: issue-4301

yoution:


yoution:
for issue #4301

maxceem (changes requested):
@yoution Approach looks great. But I guess how Ariel wants it to work is, when we click outside it should work the same as if we press `Enter` or `Space`. I. e. the value should be created not only keep text:

image

yoution:
@maxceem how about this approach 
https://monosnap.com/file/P0kV6LXfKtdlzh4PI9JE6AHwvbZVjb

maxceem (changes requested):
@yoution this is very close to what we need.

The only one thing, when we click outside can we add an option which we enter into the field, not what is chose in the list?

image

The logic is, if we want to select an option, we have to click the option. But if we click outside, we want to select what we enter into the field, not the option.

maxceem (approved):
Works great now @yoution!

Issue #4300, opened 2021-02-17
[$10] Billing Accounts: Displayed Data Enhancements


acshields:
For all returned billing account options in a project, display the following details:
- Billing Account ID
- Billing Account Name
- Billing Account Expiration Date

maxceem:
Sum up:
- update format of options for billing account:

  ![image](https://user-images.githubusercontent.com/146016/109493383-166ead80-7a95-11eb-9cee-14509cfe682b.png)
  
  - additionally, show the `endDate` from the server response in format `23 Feb 2021`. So in total options should looks like:
     `Billing Account Name (2342545345) - 23 Feb 2021`

maxceem:
Challenge https://www.topcoder.com/challenges/c3f58fe3-1608-441c-a44d-3fe3bb106ffb has been created for this ticket.<br/><br/>```This is an automated message for maxceem via Topcoder X```

Gauravseta:
@bug-bash-helper assign me

maxceem:
Challenge https://www.topcoder.com/challenges/c3f58fe3-1608-441c-a44d-3fe3bb106ffb has been assigned to gauravseta.<br/><br/>```This is an automated message for maxceem via Topcoder X```

maxceem:
@Gauravseta 12 hours have passed, but a Pull Request was not provided, so I have to unassign you from this task.

maxceem:
Challenge https://www.topcoder.com/challenges/c3f58fe3-1608-441c-a44d-3fe3bb106ffb gauravseta has been unassigned.<br/><br/>```This is an automated message for maxceem via Topcoder X```

maxceem:
@yoution this is open for pickup now.

yoution:
@bug-bash-helper assign me

maxceem:
Challenge https://www.topcoder.com/challenges/c3f58fe3-1608-441c-a44d-3fe3bb106ffb has been assigned to yoution.<br/><br/>```This is an automated message for maxceem via Topcoder X```

yoution:
![image](https://user-images.githubusercontent.com/3863914/109620510-ee567b80-7b74-11eb-8eaa-e4e4eb3278ce.png)

@maxceem I can not reproduce the issue, can you give me one account and projectid?

maxceem:
@yoution looks like my demo account doesn't have them anymore, I reach out for obtaining some demo account.

yoution:
I found the test data @maxceem

[...]

PR #4326, opened 2021-03-02
fix: issue #4300

yoution:


maxceem (changes requested):
@yoution there is one case where we don't know the date.

When a billing account is assigned, but the value of the billing account is not in the list. At the moment in such case is shows today date, but actually we don't know the date:

image

Here is a project for testing when such billing account is assigned http://connect.topcoder-dev.com/projects/16496/settings

And here is a project for testing when we have only good billing accounts: https://connect.topcoder-dev.com/projects/16952/settings

maxceem (approved):
Thank you @yoution, works good now.

PR #4327, opened 2021-03-02
fix: issue 4316

narekcat:


maxceem (approved):
Thank you @narekcat. I've fixed Projects API as per your suggestion and it works great with your Client-side changes.

Issue #4281, opened 2021-01-22
[$10] About us, contact us & Privacy/Terms menu are missing on Web View.


drasticdpk:
![image](https://user-images.githubusercontent.com/25297859/105482493-89d01300-5cd0-11eb-8ef5-9f790d74621f.png)

maxceem:
@drasticdpk it was removed as per https://github.com/appirio-tech/connect-app/issues/4061

@acshields would you like these links in the mobile menu too? Or would you like to show them back?

acshields:
@maxceem - I am unclear on the question. Are you asking whether these options should always show in the mobile menu (for example, while the user is inside their project and not just on all projects)?

maxceem:
@acshields sorry for the confusion. Some time ago we removed these links from the desktop version of the Connect App, see #4061.

As I understand in the current issue @drasticdpk is asking to put them on the Destop version of the Connect App, because we have them in mobile menu.

So I guess there are 2 options:
1. We have to revert what was done in #4061 and put these links back to the Desktop version, and show them at the bottom.
2. Or if we don't need these links because now we have a FAQ section, maybe we have to remove these links from the mobile menu too, instead of adding them to the Desktop version.

acshields:
@maxceem - let's remove them from the Mobile version :)

maxceem:
@yoution could you please comment on this issue, so I could manually assign you.

the final task here is:
- remove links from the mobile menu
- the year should be updated automatically using the current date, not hardcoded

![image](https://user-images.githubusercontent.com/146016/109934354-1b21a480-7cd5-11eb-947f-dcc2c6ecbf3f.png)

yoution:
@maxceem  please assign to me

maxceem:
Challenge https://www.topcoder.com/challenges/d0402357-b4c5-4ab8-9d35-8491754fc2f2 has been created for this ticket.<br/><br/>```This is an automated message for maxceem via Topcoder X```

maxceem:
Challenge https://www.topcoder.com/challenges/d0402357-b4c5-4ab8-9d35-8491754fc2f2 has been assigned to yoution.<br/><br/>```This is an automated message for maxceem via Topcoder X```

drasticdpk:
Verified on Prod
![image](https://user-images.githubusercontent.com/25297859/110745681-bd0c3880-8263-11eb-8aac-1e73465486ee.png)

PR #4333, opened 2021-03-04
fix: issue #4281

yoution:
@maxceem please review


maxceem (approved):
Thanks, @yoution all good.

Issue #4310, opened 2021-02-22
[$20] Increase Body Width of All Intake Forms


acshields:
Increase the body width of all intake forms to match the updated width parameters of the [new TaaS form](https://github.com/appirio-tech/connect-app/issues/4298#issuecomment-783242097).

maxceem:
@acshields at the moment the width is `600px`. How much would you like to increase it? I suggest, not that much, for example until `720px`.

![image](https://user-images.githubusercontent.com/146016/109494973-4ae36900-7a97-11eb-8c84-2e557f7ff726.png)

I've noticed that designers prefer to make the forms not so wide, for example.

In TaaS App it's 640px:

![image](https://user-images.githubusercontent.com/146016/109495204-9c8bf380-7a97-11eb-9d68-fb8a3618127c.png)

New Work Manager design 720px:

![image](https://user-images.githubusercontent.com/146016/109495452-f8ef1300-7a97-11eb-9091-46dfe1007490.png)

And not only Topcoder designers tend to make forms narrow. For example forms in Google Forms:

![image](https://user-images.githubusercontent.com/146016/109495757-5c794080-7a98-11eb-9d85-671e172e5ddd.png)

Should we consult with designers first? Or do you have any preference on what width to set?

acshields:
@maxceem - lets do the 720px

maxceem:
Sum up:

- increase the width of all the forms for creating projects https://connect.topcoder-dev.com/new-project/ to `720px` 
- make sure that nothing got broken inside the forms after increasing the width
- provide a demo video showing at least 3 forms that everything works good inside them - UI is not broken:
  - https://connect.topcoder-dev.com/new-project/app_new_intake
  - https://connect.topcoder-dev.com/new-project/taas-jobs
  - https://connect.topcoder-dev.com/new-project/qa_acceptance_testing

maxceem:
@yoution could you please comment here so I can assign you manually.

yoution:
@maxceem  please assign to me

maxceem:
Challenge https://www.topcoder.com/challenges/c54d0323-b68b-4090-9f6a-01136938f9da has been created for this ticket.<br/><br/>```This is an automated message for maxceem via Topcoder X```

maxceem:
[...]

PR #4334, opened 2021-03-05
fix: issue #4310

yoution:
@maxceem please review
[demo video](https://monosnap.com/file/hfMOilrqpqKBESjAUgWBDk0izqsErW)


maxceem (changes requested):
Thanks, @yoution.

Generally all good, I've just noticed one thing. In the TaaS form there are 2 fields that before take 50% each:

image

but now they have different width, not 50%, and there is some extra space:

image

Issue #4348, opened 2021-04-14
Customer must not see the status of the phase


SathyaJayabal:
Based on this req, the customer team must not see the published status. Since they do not see the draft status too, we could remove the status column for the customers.
https://github.com/appirio-tech/connect-app/issues/4339#issue-826435818

```
No Text Display
Once a phase is published, the Client users do not need to see the "Published" label.
```

vikasrohit:
@SathyaJayabal can you please add screenshot of how it is looking for the customer right now?

SathyaJayabal:
@vikasrohit 
<img width="789" alt="Screenshot 2021-04-15 at 9 54 36 AM" src="https://user-images.githubusercontent.com/58783823/114813887-a3b15b80-9dd0-11eb-968b-5a94406c18d8.png">

SathyaJayabal:
@mirzailhami @vikasrohit , The status label has been removed, but the phase status is still visible for the customers.
example:
https://connect.topcoder-dev.com/projects/17270

Customer user: tester1234
<img width="1440" alt="Screenshot 2021-04-20 at 9 31 38 AM" src="https://user-images.githubusercontent.com/58783823/115335717-66c3db00-a1bb-11eb-906f-2fcda2a52fa4.png">

vikasrohit:
@mirzailhami we need to hide actual phase status as well from customers.

mirzailhami:
> @mirzailhami we need to hide actual phase status as well from customers.

Ok, @diwosuwanto working on it.

SathyaJayabal:
I dont see the fix yet @mirzailhami , is it deployed ?

mirzailhami:
@SathyaJayabal deployed was not successful. Could you check pls @vikasrohit.
![Screen Shot 2021-04-20 at 21 59 04](https://user-images.githubusercontent.com/8240895/115418573-c950c180-a223-11eb-825e-1b1b4055b0e9.png)

vikasrohit:
@mirzailhami there is a lint error https://app.circleci.com/pipelines/github/appirio-tech/connect-app/472/workflows/019b14dc-c006-41ef-a306-bc935dd7c868/jobs/5677

SathyaJayabal:
@mirzailhami @vikasrohit , the phase status must be removed only for the new simplified projects for the customer. Older phases should still display the phase statuses.

Example: This is an older project 
https://connect.topcoder-dev.com/projects/16573/

The customer must see the phase status details
[...]

PR #4359, opened 2021-04-20
Fix: hide actual phase status from customers

diwosuwanto:
Fix this issue: [https://github.com/appirio-tech/connect-app/issues/4348](url)

### DIFF
--- a/src/projects/detail/components/PhaseCard/PhaseCard.jsx
+++ b/src/projects/detail/components/PhaseCard/PhaseCard.jsx
@@ -225,7 +225,7 @@ class PhaseCard extends React.Component {
                   }
 
                   <div styleName="col hide-md">
-                    {status && (isSimplePlan || status !== PHASE_STATUS_ACTIVE) && !this.props.currentUserRoles.includes(PROJECT_ROLE_CUSTOMER) &&
+                    {status && (isSimplePlan || status !== PHASE_STATUS_ACTIVE) && !this.props.currentUserRoles.includes(PROJECT_ROLE_CUSTOMER) && this.props.currentUserRoles.length > 2 &&
                           (<div styleName="status-details">
                             <div styleName={'status ' + (status ? status.toLowerCase() : '')}>
                               {statusDetails.name}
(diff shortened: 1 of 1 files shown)


src/projects/detail/components/PhaseCard/PhaseCard.jsx
```diff
@@ -225,15 +225,15 @@ class PhaseCard extends React.Component {
                   }
 
                   <div styleName="col hide-md">
-                    {status && (isSimplePlan || status !== PHASE_STATUS_ACTIVE) && !this.props.currentUserRoles.includes(PROJECT_ROLE_CUSTOMER) &&
+                    {status && (isSimplePlan || status !== PHASE_STATUS_ACTIVE) && !this.props.currentUserRoles.includes(PROJECT_ROLE_CUSTOMER) && this.props.currentUserRoles.length > 2 &&
```
mirzailhami, line 228: @vikasrohit I dont think this solution is good enough. The `currentUserRoles` sometime has 2 and 3 values. 

@diwosuwanto account
WhatsApp Image 2021-04-16 at 13 37 00

@mirzailhami account
WhatsApp Image 2021-04-16 at 13 43 10

Any suggestion @vikasrohit?

vikasrohit: You are right @mirzailhami You don't need to check currentUserRoles. We have to check project role of the logged in user because a user can not see without joining a project (unless he/she is admin). And best way to achieve this is following our new permission model. Example code here https://github.com/appirio-tech/connect-app/blob/7d640885c00323ebef298f022a079384decdb8af/src/projects/detail/containers/DashboardContainer.jsx#L100 and https://github.com/appirio-tech/connect-app/blob/7d640885c00323ebef298f022a079384decdb8af/src/config/permissions.js#L178

mirzailhami: @diwosuwanto Please look into the code reference & suggestion above. You can commit the update, then. Thanks.

diwosuwanto: Thank you for the feedback sir @mirzailhami i @vikasrohit, i already commit the update. Please check my commit, thank you.

vikasrohit: We need to create new permission instead of reusing `EXPAND_ACTIVE_PHASES_BY_DEFAULT`. Lets call it `SHOW_PHASE_STATUS`

Issue #4360, opened 2021-04-20
[$20] Member item must be vertically middle aligned


mirzailhami:
### Expected behavior
Member row item should be vertically middle aligned.

### Actual behavior
Top aligned instead of middle aligned.

### Steps to reproduce the problem
- Login into [connect](https://connect.topcoder-dev.com/projects)
- Scroll down and look into left sidebar

### Screenshot/screencast
![Screen Shot 2021-04-17 at 14 27 45](https://user-images.githubusercontent.com/8240895/115360616-fbddc880-a1e9-11eb-92e6-f86abb1b7846.png)

vikasrohit:
@mirzailhami thanks for logging it. We would take care of it in next cool down period.

drasticdpk:
Reference : 
![image](https://user-images.githubusercontent.com/25297859/118396964-a45c3c80-b671-11eb-8c56-c2aeef4d3738.png)

vikasrohit:
@drasticdpk just deployed the feature branch to dev env. Please retest now.

drasticdpk:
Verified
![image](https://user-images.githubusercontent.com/25297859/118443206-5a279980-b70b-11eb-97f8-b9968e95c279.png)

drasticdpk:
Verified on Prod Mode. 
![image](https://user-images.githubusercontent.com/25297859/118625833-13f53780-b7ea-11eb-9fa4-1bc75dde66d0.png)

PR #4375, opened 2021-05-06
Fix issue #4360

diwosuwanto:
Fix member row item should be vertically middle aligned #4360


mirzailhami:
> Fix member row item should be vertically middle aligned #4360
> 
> 

Look goods to me, any suggestion @vikasrohit?

vikasrohit:
Looks good to me

PR #4454, opened 2021-08-11
fix: change table for customer view

yoution:
@elkhawajah please review


elkhawajah:
Hiding columns should not break the style. So you also need to fix style to work perfectly with customer view or PM/Copilot view.

image

yoution:
@elkhawajah please review

PR #4530, opened 2021-11-28
Fix failing tests

nursoltan-s:


maxceem (changes requested):
@nursoltan-s 

The failing case with the project name now looks fixed.

1. The unhandled promise rejections are still there https://monosnap.com/file/snf3ib2b8xJF4sbeHpwrFhg4fZbg5G. We still have to find the places where they come from and fix them.

2. Now some 2 other test cases are failing:

    image

PR #4594, opened 2023-02-14
Universal Nav implementation

jmgasper:
Universal Navigation implementation:

https://github.com/appirio-tech/connect-app/milestone/120


kkartunov (approved):

config/constants/dev.js
```diff
@@ -1,3 +1,5 @@
+const DOMAIN = 'topcoder-dev.com'
```
jpeg22, line 1: Can you move the value of DOMAIN into config?

PR #4599, opened 2023-08-01
Remove navigation for profile and account details from Connect

jmgasper:
https://topcoder.atlassian.net/browse/MP-294


eisbilir (approved):

kkartunov (changes requested):
Can we also do code cleaning of containers & components that are not used anywhere anymore?
Like: profile/settings/container and etc.

jmgasper:
> 

@kkartunov - We *can*, but since we aren't actively developing Connect and it's going to be deprecated, I went with the lowest "touch" of the codebase.  This is something that could open up a can of worms that I'd like to avoid.

kkartunov (approved):

